PR #108, opened 2024-06-07
test: expect event to be emitted (#88)

### LINKED ISSUE #88: [Feature]: Abstract away asserting events get emitted in e2e tests

There's a bunch of code needed to assert an event gets emitted. Take this snippet:

```rust
    let block = alice.signer.get_block_number().await?;
    let filter = Filter::new().address(contract_addr).from_block(block);
    let logs = alice.signer.get_logs(&filter).await?;
    let transfer: Erc721::Transfer =
        logs[logs.len() - 1].log_decode()?.inner.data;
```

This should be abstracted away to something like `receipt.emits(Erc721::Transfer {from, to, tokenId})`.

Note that we can get the block number from the receipt.

### Contribution Guidelines

bidzyyys:
Resolves #88


lib/e2e/src/assert.rs
alexfertel, line 1: Should we maybe rename it to `assertions.rs`? I'm not sure, I feel like `assert.rs` is not a good name for this module.

lib/e2e/src/assert.rs
```diff
@@ -26,3 +30,28 @@ impl<R: SolError> Assert<R> for alloy::contract::Error {
         }
     }
 }
+
+pub trait Emits<E> {
+    /// Asserts that transaction emitted an `expected` event.
```
alexfertel, line 35:
```suggestion
    /// Asserts the transaction emitted the `expected` event.
```

lib/e2e/src/assert.rs
```diff
@@ -26,3 +30,28 @@ impl<R: SolError> Assert<R> for alloy::contract::Error {
...
+{
+    fn emits(&self, expected: E) {
+        // Extract all events the are the expected type;
+        let emitted = self
+            .inner
+            .logs()
+            .iter()
+            .filter_map(|log| log.log_decode().ok())
+            .map(|log| log.inner.data)
+            .any(|event| expected == event);
```
alexfertel, line 53: niceeee

lib/e2e/src/assert.rs
```diff
@@ -26,3 +30,28 @@ impl<R: SolError> Assert<R> for alloy::contract::Error {
...
+
+        assert_eq!(emitted, true, "Event {:?} not emitted", expected);
```
alexfertel, line 55: Since we are adding an assertion message ("Event ....") maybe it makes sense to just `assert!` here instead of `assert_eq!`? wdyt?

examples/erc721/tests/erc721.rs
```diff
@@ -40,7 +40,13 @@ macro_rules! send {
 
 macro_rules! watch {
     ($e:expr) => {
-        $e.send().await?.watch().await
+        send!($e)?.watch().await
+    };
+}
+
+macro_rules! wait_for_receipt {
```
alexfertel, line 47: This is not a strong opinion, but what do you think of calling it just `receipt!`?

When people see this macro, they'll immediately jump to the definition, which is a straightforward one-liner. Then, having to type `wait_for_receipt` can become burdensome.

alexfertel (commented):
Looking super nice, I like this much better!

lib/e2e/src/assert.rs
bidzyyys: Ok.

lib/e2e-proc/src/lib.rs
```diff
@@ -9,7 +9,6 @@ mod test;
 ///
 /// ```rust,ignore
 /// #[e2e::test]
-/// async fn foo(alice: User, bob: User) -> eyre::Result<()> {
```
alexfertel: 👀

lib/e2e/src/lib.rs
```diff
@@ -1,12 +1,88 @@
...
 mod user;
 
-pub use assert::Assert;
+pub use assertions::{Assert, Emits};
 pub use deploy::deploy;
 pub use e2e_proc::test;
 pub use system::{provider, Provider, Signer};
 pub use user::User;
+
+/// This macro provides an easy way for broadcasting the transaction
```
alexfertel, line 14: What do you think of changing the wording like this?

```suggestion
/// This macro provides a shorthand for broadcasting the transaction
```

alexfertel (approved):
LGTM! Just a comment about wording. 💯

lib/e2e-proc/src/lib.rs
bidzyyys: I know, already reverted this change 😅

alexfertel: Yeah, I was reviewing when you updated and the comment stayed. That's why I resolved the conversation.

PR #109, opened 2024-06-07
ref(e2e): remove the need to set tx's from by bumping alloy

alexfertel:
Since https://github.com/alloy-rs/alloy/issues/830 has been resolved, we don't need to set `from` explicitly anymore.


bidzyyys (approved):
LGTM! 🚀

PR #110, opened 2024-06-07
fix(e2e): properly check errors match

alexfertel:
Follow-up work to #80 

We were previously just checking that an error's signature was contained in the JSON-RPC response, but now we check the revert data fully.

Also polished the API.


lib/e2e/src/assert.rs
```diff
@@ -1,28 +1,23 @@
-use alloy::{sol_types::SolError, transports::RpcError};
+use alloy::sol_types::SolError;
 
-pub trait Assert<E> {
+pub trait ErrorExt<E> {
     /// Asserts that current error result corresponds to the typed abi-encoded
```
bidzyyys, line 4: No `assert` here; may be misleading.

bidzyyys (approved):
I like it 🚀 
Left a small comment.

lib/e2e/src/assert.rs
alexfertel: Yeah, I'll rework this a bit, and will probably move this to a separate file.

bidzyyys: sure, ping me when I should re-review it.

bidzyyys (approved):
I like the refactor 🚀

PR #114, opened 2024-06-11
fix(ci): include e2e tests in coverage

### LINKED ISSUE #92: [Feature]: Include e2e tests in coverage

We don't currently include e2e tests in coverage because some setup is required to run them. Specifically, you need `solc` and `nitro-testnode` to be available.

We should either create a new workflow that uploads this to codecov or update the current coverage workflow to include e2e tests.

### Contribution Guidelines

alexfertel:
Resolves #92

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -1,11 +1,15 @@
-name: e2e
 # This workflow runs our end-to-end tests suite and collects coverage.
 #
 # It roughly follows these steps:
-# - A local `nitro-testnode` gets spun up.
-# - The test suite runs.
+# - Install Rust
+# - Install `cargo-stylus`
+# - Install `solc`
+# - Spin up `nitro-testnode`
+#
+# And then we either run the test suite or run `cargo llvm-cov`.
 #
 # Contract deployments and account funding happen on a per-test basis.
+name: e2e
 permissions:
   contents: read
 on:
@@ -19,7 +23,7 @@ env:
   CARGO_TERM_COLOR: always
 jobs:
   required:
-    name: ubuntu / tests
+    name: tests
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


bidzyyys (approved):
looks good!

lib/e2e/src/project.rs
```diff
@@ -49,13 +51,32 @@ fn get_wasm(name: &str) -> eyre::Result<PathBuf> {
...
-    let executable = env::current_exe()?;
-    let out_dir = executable
-        .parent()
-        .expect("executable path should have a final component")
-        .join("../../");
-    let wasm: PathBuf =
-        out_dir.join(format!("wasm32-unknown-unknown/release/{name}.wasm"));
+    let mut target_dir = env::current_exe()?;
+
+    // Recursively find a `target` directory.
```
qalisander, line 56: I would say this approach is much better for distributing library. But a bit hacky:)

.github/workflows/e2e-tests.yml
```diff
@@ -63,3 +63,19 @@ jobs:
         run: |
           export NIGHTLY_TOOLCHAIN=${{steps.toolchain.outputs.name}}
           ./scripts/e2e-tests.sh
+      # Coverage
+      - name: cargo install cargo-llvm-cov
```
qalisander, line 67: Can you explain how it should work?
Are we checking test coverage of test framework here or integration tests?
And if it is last one how it works?

qalisander (commented):
Great!

lib/e2e/src/project.rs
alexfertel: Yeah, we should probably find a better way in the future, I'm not that happy with it. Nonetheless, it should be fine for most use cases.

.github/workflows/e2e-tests.yml
alexfertel: Ofc! So, this is running the `e2e` tests and collecting coverage data, which is uploaded to codecov.io and merged with the data uploaded from the `test` workflow.

> Are we checking test coverage of test framework here, unit tests code or contracts library?

Not entirely sure what you mean. We have two workflows right now that collect coverage data: `test` and this one (`e2e`).

* `test` runs `cargo test --features std`
* `e2e` runs `cargo test --features std,e2e --test '*'`

So, the first one runs unit tests, and the second one runs integration tests.

Does this answer your questions?

qalisander: I mean it in a way of how code coverage implemented inside. Is it changing binary code to insert counters or smth else? And will it work for wasm binaries deployed to nitro node?

alexfertel: ah, no clue hehe

alexfertel: I expect it to work, yes. for example, events are getting covered, but we need to write more tests and see how the diff changes.

qalisander: oke. Will check out then

qalisander (approved):

qalisander:
Also seems that test coverage adds few minutes to pipeline to pass

alexfertel:
> Also seems that test coverage adds few minutes to pipeline to pass

Yeah, coverage has to be a separate workflow, let me fix that.

PR #115, opened 2024-06-11
fix(ci): cache rust dependencies per pr

qalisander:
E2e pipeline forcing recompilation and deps download when we open a new pr despite caching step.
Cache hit still occurs and stale cache retrieved. Since cache hit happened new (valid) cache won't be saved. So it doesn't help after consecutive pushes. 
Just cache reset forces cargo cache to be persisted and it works faster after.

This pr basically adds caching for `e2e` test per scope of pull request (pr id).
And if it wasn't found (like pr just opened) should try retrieve a cache with the same hash of `Cargo.lock` file and persist new cache per pr id.
Will see how it will work.

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -29,7 +29,7 @@ jobs:
             ~/.cargo/registry/index/
             ~/.cargo/registry/cache/
             ~/.cargo/git/db/
-            target/
+            ./target/
           key: ${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}
           
 
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -29,9 +29,8 @@ jobs:
             ~/.cargo/registry/index/
             ~/.cargo/registry/cache/
             ~/.cargo/git/db/
-            ./target/
-          key: ${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}
-          
+            target/
+          key: ${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}-${{ github.event.number }}
 
       - name: cache cargo-stylus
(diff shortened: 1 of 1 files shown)


qalisander:
Experiment pr with caching policies for github

.github/workflows/e2e-tests.yml
```diff
@@ -29,8 +29,9 @@ jobs:
             ~/.cargo/registry/index/
             ~/.cargo/registry/cache/
             ~/.cargo/git/db/
-            target/
+            ./target/
           key: ${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}
```
alexfertel, line 33: I think we also would have to do something like:

```suggestion
          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
```

But I'm not entirely sure.

qalisander: I'm not sure that in this case it will look at the toplevel 'Cargo.lock' file. It is mostly for a project with many lockfiles but we have a single one.

alexfertel: yeah, you're right 👍🏻

bidzyyys:
What is the plan for this PR?

qalisander:
I've noticed that caching is not helping sometimes when we open a new pr. But cache hit still occurs according to a pipeline that makes me think we retrieve a stale cache. That doesn't help much after we recompile but implementation of caching seems correct according to recommendations of action/cache https://github.com/actions/cache/blob/main/examples.md#rust---cargo . This draft pr is just to play with pipelines.

alexfertel (approved):
🤯 

This shaves the running time for tests by half (🚀), but surely there must be a way to make this work more easily.

GitHub Actions DX hits again...

bidzyyys (approved):
LGTM

PR #116, opened 2024-06-12
feat: custom derive macro for Default

### LINKED ISSUE #51: [Feature]: Add a macro that implements `Default` for `solidity_storage` types

We currently implement `Default` by hand for contract structs like this:

```rust
impl Default for ERC20 {
    fn default() -> Self {
        let root = U256::ZERO;
        ERC20 {
            _balances: unsafe { StorageMap::new(root, 0) },
            _allowances: unsafe {
                StorageMap::new(root + U256::from(32), 0)
            },
            _total_supply: unsafe {
                StorageU256::new(root + U256::from(64), 0)
            },
        }
    }
}
```

This is a repetitive and burdensome process, so we should provide a macro to ease our lives.

### Contribution Guidelines

ggonzalez94:
Resolves #51 

This PR implements a custom derive macro for Default inside `motsu-proc` crate that makes it available inside all unit tests. This follows the rules of solidity storage and only works with  `sol_storage!` structs for now.
- All places where Default was implemented were updated and tests are passing

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/lib/motsu-proc/src/default_storage_layout.rs
+++ b/lib/motsu-proc/src/default_storage_layout.rs
@@ -1,4 +1,4 @@
-//! Defines the `#[derive(motsu::StylusDefault)]` procedural macro.
+//! Defines the `#[derive(motsu::DefaultStorageLayout)]` procedural macro.
 use proc_macro::TokenStream;
 use quote::{quote, ToTokens};
 use syn::{Data, DeriveInput, PathArguments, Type};
@@ -7,7 +7,7 @@ pub fn impl_stylus_default(ast: &DeriveInput) -> TokenStream {
     let name = &ast.ident;
 
     let Data::Struct(ref data_struct) = ast.data else {
-        error!(ast, "StylusDefault can only be derived for structs");
+        error!(ast, "DefaultStorageLayout can only be derived for structs");
     };
 
     let fields = match &data_struct.fields {
--- a/lib/motsu-proc/src/default_storage_layout.rs
+++ b/lib/motsu-proc/src/default_storage_layout.rs
@@ -1,9 +1,9 @@
 //! Defines the `#[derive(motsu::DefaultStorageLayout)]` procedural macro.
 use proc_macro::TokenStream;
-use quote::{quote, ToTokens};
+use quote::quote;
 use syn::{Data, DeriveInput, PathArguments, Type};
 
-pub fn impl_stylus_default(ast: &DeriveInput) -> TokenStream {
+pub fn impl_default_storage_layout(ast: &DeriveInput) -> TokenStream {
     let name = &ast.ident;
 
     let Data::Struct(ref data_struct) = ast.data else {
@@ -55,12 +55,12 @@ pub fn impl_stylus_default(ast: &DeriveInput) -> TokenStream {
 
         let field_init = quote! {
             {
-                let instance = unsafe { #type_ident::new(alloy_primitives::U256::from(next_slot), offset) };
-                offset += #type_ident::SLOT_BYTES as u8;
-                if offset >= 32 {
+                if offset + #type_ident::SLOT_BYTES as u8 > 32 {
                     next_slot += 32;
                     offset = 0;
                 }
+                let instance = unsafe { #type_ident::new(alloy_primitives::U256::from(next_slot), offset) };
+                offset += #type_ident::SLOT_BYTES as u8;
                 instance
             }
         };
--- a/lib/motsu-proc/src/default_storage_layout.rs
+++ b/lib/motsu-proc/src/default_storage_layout.rs
@@ -32,22 +32,20 @@ pub fn impl_stylus_default(ast: &DeriveInput) -> TokenStream {
         // Types when using `sol_storage!` look like this:
         // `stylus_sdk::storage::type<generic arguments>`
         // (e.g. uint256 is stylus_sdk::storage::StorageUint<256,4>).
-        // So we must first get the third argument, which is the main
+        // So we must first get the last argument, which is the main
         // type.
         let segments = &type_path.path.segments;
-        let main_type = if segments.len() >= 3 {
-            &segments[2].ident
+        let last_segment = if segments.len() >= 3 {
+            segments.last().unwrap()
         } else {
             error!(type_path, "unexpected type path");
         };
+        let main_type = &last_segment.ident;
 
         // If the type has generic arguments form the token stream that
         // we latter append to access `new` and `SLOT_BYTES`
-        let generic_args = match &segments[2].arguments {
-            PathArguments::AngleBracketed(args) => {
-                let args_tokens = args.to_token_stream();
-                quote! { ::#args_tokens }
-            }
+        let generic_args = match &last_segment.arguments {
+            PathArguments::AngleBracketed(args) => quote! { ::#args },
             _ => quote! {},
         };
 
@@ -57,9 +55,6 @@ pub fn impl_stylus_default(ast: &DeriveInput) -> TokenStream {
 
         let field_init = quote! {
             {
-                // Usually we would include an import of `alloy_primitives::U256`, but this causes conflicts
-                // if it is already imported in the file that is using this macro. Instead we use the full
-                // here to avoid this issue.
                 let instance = unsafe { #type_ident::new(alloy_primitives::U256::from(next_slot), offset) };
                 offset += #type_ident::SLOT_BYTES as u8;
                 if offset >= 32 {
@@ -75,10 +70,6 @@ pub fn impl_stylus_default(ast: &DeriveInput) -> TokenStream {
         });
     }
 
-    let combined_initializations = quote! {
-        #(#field_initializations),*
-    };
-
     quote! {
         use stylus_sdk::prelude::StorageType;
         impl Default for #name {
@@ -86,7 +77,7 @@ pub fn impl_stylus_default(ast: &DeriveInput) -> TokenStream {
                 let mut next_slot: i32 = 0;
                 let mut offset: u8 = 0;
                 #name {
-                    #combined_initializations
+                    #(#field_initializations),*
                 }
             }
         }
--- a/lib/motsu-proc/src/default_storage_layout.rs
+++ b/lib/motsu-proc/src/default_storage_layout.rs
@@ -17,50 +17,19 @@ pub fn impl_default_storage_layout(ast: &DeriveInput) -> TokenStream {
     };
 
     let mut field_initializations = Vec::new();
-
     for field in fields {
         let field_name = &field.ident;
-        let field_ty = &field.ty;
+        let field_type = &field.ty;
 
-        let Type::Path(type_path) = field_ty else {
-            error!(
-                field_ty,
-                "unsupported field type. Only path types are supported"
-            );
-        };
-
-        // Types when using `sol_storage!` look like this:
-        // `stylus_sdk::storage::type<generic arguments>`
-        // (e.g. uint256 is stylus_sdk::storage::StorageUint<256,4>).
-        // So we must first get the last argument, which is the main
-        // type.
-        let segments = &type_path.path.segments;
-        let last_segment = if segments.len() >= 3 {
-            segments.last().unwrap()
-        } else {
-            error!(type_path, "unexpected type path");
-        };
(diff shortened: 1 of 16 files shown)


lib/motsu-proc/src/lib.rs
```diff
@@ -42,3 +44,48 @@ mod test;
...
+/// For now this macro only works with structs that use the
+/// `sol_storage!` macro that allows you to use the solidity syntax.
+/// If you want to write your contracts using `#[solidity_storage]`
+/// and the Rust syntax, you will have to implement `Default` youself.
+///
+/// ## See Also
+///
+/// - [Layout of State Variables in Storage](https://docs.soliditylang.org/en/latest/internals/layout_in_storage.html)
+
+#[proc_macro_derive(StylusDefault)]
```
bidzyyys, line 87: I am not convinced to the name `StylusDefault`, for me this name can be misleading. What do you think about `MotsuDefault` or `TestDefault` to point out that it should be used only in tests?

bidzyyys (commented):
Great job @ggonzalez94 This feature will be super helpful 🚀

lib/motsu-proc/src/lib.rs
alexfertel: What do you think of `DefaultStorageLayout`? cc: @ggonzalez94

bidzyyys: I think I may like it.

ggonzalez94: Yep, that works for me. I'll change the name.

lib/motsu-proc/src/stylus_default.rs
```diff
@@ -0,0 +1,95 @@
...
+            PathArguments::AngleBracketed(args) => {
+                let args_tokens = args.to_token_stream();
+                quote! { ::#args_tokens }
+            }
+            _ => quote! {},
+        };
+
+        // Reconstruct the type with the correct formatting
+        let type_ident =
+            quote! { stylus_sdk::storage:: #main_type #generic_args };
```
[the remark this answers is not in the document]
ggonzalez94: Couldn't find a better way to do this 😮‍💨 - but I did make it shorter, so I don't think is too bad. Can you check it in the renamed file?

ggonzalez94 (commented):
So after adding tests I realized we were not calculating correctly the slot and offset, because the if statement that updated the values was after we initialized the field. So a contract like this:
```
sol_storage! {
        #[derive(motsu_proc::DefaultStorageLayout)]
        pub struct AddressContract {
            address lender;
            address borrower;
        }
    }
```

would end up expanding to something like this(omitting the unsafe keyword and some other syntax to make it shorter):
```
 pub struct AddressContract {
    lender: StorageAddress::new(0,0);
    borrower: SotrageAddress::new(0,20); // This is wrong - an address shouldn't begin at this offset, but we were updating the values after
}
```
With the changes it seems to be expanding correctly and passing all the new tests + the unit tests for contracts

lib/motsu-proc/tests/default.rs
ggonzalez94, line 1: I had to use modules here because otherwise the import from the macro `use stylus_sdk::prelude::StorageType;` was getting imported twice, and thus have a conflict. I don't know if this is expected or if there's a better way to deal with this.

Also do these tests run by default in our pipeline or should I explicitly add them?

lib/motsu-proc/src/default_storage_layout.rs
```diff
@@ -0,0 +1,86 @@
...
+                let instance = unsafe { #type_ident::new(alloy_primitives::U256::from(next_slot), offset) };
+                offset += #type_ident::SLOT_BYTES as u8;
+                instance
+            }
+        };
+
+        field_initializations.push(quote! {
+            #field_name: #field_init
+        });
+    }
```
alexfertel, line 71: Made is simpler and more general, wdyt?

```rust
    let mut field_initializations = Vec::new();
    for field in fields {
        let field_name = &field.ident;
        let field_ty = &field.ty;

        let ty = quote! { <#field_ty as stylus_sdk::storage::StorageType> };
        let field_init = quote! {
            {
                if offset + #ty::SLOT_BYTES as u8 > 32 {
                    next_slot += 32;
                    offset = 0;
                }
                let instance = unsafe { #ty::new(alloy_primitives::U256::from(next_slot), offset) };
                offset += #ty::SLOT_BYTES as u8;
                instance
            }
        };

        field_initializations.push(quote! {
            #field_name: #field_init
        });
    }
```

alexfertel: The key insight is that it failed to compile because `new` comes from `StorageType`, so we need to explicitly cast the type.

lib/motsu-proc/src/default_storage_layout.rs
```diff
@@ -0,0 +1,86 @@
...
+
+    quote! {
+        use stylus_sdk::prelude::StorageType;
```
alexfertel, line 74: A corollary of the previous change is that we no longer need to import this here, which means we can remove the modules in the tests.

```suggestion
```

Resulting in:

```rust
use alloy_primitives::{address, Address};
use alloy_primitives::{U128, U16, U256, U32, U64, U8};
use stylus_sdk::stylus_proc::sol_storage;

sol_storage! {
    #[derive(motsu_proc::DefaultStorageLayout)]
    pub struct Erc20 {
        mapping(address => uint256) _balances;
        mapping(address => mapping(address => uint256)) _allowances;
        uint256 _total_supply;
    }
}

#[motsu::test]
fn erc20_initiates_correctly(contract: Erc20) {
    assert_eq!(contract._total_supply.get(), U256::from(0));
}

sol_storage! {
    #[derive(motsu_proc::DefaultStorageLayout)]
    pub struct UintContract {
        uint256 a;
        uint128 b;
        uint64 c;
        uint32 d;
        uint16 e;
        uint8 f;
        uint8 g;
    }
}

// ...

#[motsu::test]
fn address_updates_correctly(contract: AddressContract) {
    let lender = address!("a935CEC3c5Ef99D7F1016674DEFd455Ef06776C5");
    let borrower = address!("DeaDbeefdEAdbeefdEadbEEFdeadbeEFdEaDbeeF");
    contract.lender.set(lender);
    contract.borrower.set(borrower);
    assert_eq!(contract.lender.get(), lender);
    assert_eq!(contract.borrower.get(), borrower);
}
```

lib/motsu-proc/tests/default.rs
alexfertel: > I don't know if this is expected or if there's a better way to deal with this.

Addressed in the previous comment.

> Also do these tests run by default in our pipeline or should I explicitly add them?

They run by default, if you run `cargo test --features std` locally, you should see these tests.

The idea being that these are plain integration tests, so they don't need any fancy setup, as opposed to tests marked with `#[cfg(feature = "e2e")]`.

lib/motsu-proc/tests/default.rs
```diff
@@ -0,0 +1,94 @@
...
+        #[derive(motsu_proc::DefaultStorageLayout)]
+        pub struct Erc20 {
+            mapping(address => uint256) _balances;
+            mapping(address => mapping(address => uint256)) _allowances;
+            uint256 _total_supply;
+        }
+    }
+
+    #[motsu::test]
+    fn erc20_initiates_correctly(contract: Erc20) {
```
alexfertel, line 14: nit: I'd prefer if we didn't specify `Erc20` here, but instead just the types: `mapping`, arrays, etc. The rationale is that `motsu` shouldn't concern itself with that + we already test that elsewhere.

lib/motsu-proc/tests/default.rs
```diff
@@ -0,0 +1,94 @@
...
+        contract.g.set(U8::from(7));
+
+        assert_eq!(contract.a.get(), U256::from(1));
+        assert_eq!(contract.b.get(), U128::from(2));
+        assert_eq!(contract.c.get(), U64::from(3));
+        assert_eq!(contract.d.get(), U32::from(4));
+        assert_eq!(contract.e.get(), U16::from(5));
+        assert_eq!(contract.f.get(), U8::from(6));
+        assert_eq!(contract.g.get(), U8::from(7));
+    }
```
alexfertel, line 64: Beautiful! This is what I had in mind!

I know it's a bit tedious, but could we also have a test case, where the types are unsorted, like

```
assert_eq!(contract.a.get(), U256::from(1));
assert_eq!(contract.b.get(), U16::from(5));
assert_eq!(contract.c.get(), U64::from(3));
assert_eq!(contract.d.get(), U8::from(6));
assert_eq!(contract.e.get(), U32::from(4));
assert_eq!(contract.f.get(), U128::from(2));
assert_eq!(contract.g.get(), U8::from(7));
```

lib/motsu-proc/tests/default.rs
```diff
@@ -0,0 +1,94 @@
...
+    sol_storage! {
+        #[derive(motsu_proc::DefaultStorageLayout)]
+        pub struct AddressContract {
+            address lender;
+            address borrower;
+        }
+    }
+
+    #[motsu::test]
+    fn address_initiates_correctly(contract: AddressContract) {
```
alexfertel, line 80: nit: initiates -> initializes

Also, I think `correctly` is redundant? But not a strong opinion, feel free to ignore.

alexfertel (commented):
Looking goood!! Only a couple of comments solving your questions and a few style nits.

lib/motsu-proc/src/stylus_default.rs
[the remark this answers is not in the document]
alexfertel: Addressed in https://github.com/OpenZeppelin/rust-contracts-stylus/pull/116#discussion_r1639687090

lib/motsu-proc/tests/default.rs
bidzyyys: +1

lib/motsu-proc/tests/default.rs
```diff
@@ -0,0 +1,94 @@
...
+        pub struct Erc20 {
+            mapping(address => uint256) _balances;
+            mapping(address => mapping(address => uint256)) _allowances;
+            uint256 _total_supply;
+        }
+    }
+
+    #[motsu::test]
+    fn erc20_initiates_correctly(contract: Erc20) {
+        assert_eq!(contract._total_supply.get(), U256::from(0));
```
bidzyyys, line 15: `U256::ZERO`

lib/motsu-proc/tests/default.rs
```diff
@@ -0,0 +1,94 @@
...
+            uint32 d;
+            uint16 e;
+            uint8 f;
+            uint8 g;
+        }
+    }
+
+    #[motsu::test]
+    fn uint_initiates_correctly(contract: UintContract) {
+        assert_eq!(contract.a.get(), U256::from(0));
```
bidzyyys, line 38: If you want to use just `0`, use `U256::ZERO` and same for the rest of these types.

lib/motsu-proc/tests/default.rs
ggonzalez94: Sure, that makes sense! I tried to align the types to be more compact in storage(and thus stress testing more the algorithm), but I'll add a couple more for a contract where the types don't line up with storage the same way.

lib/motsu-proc/src/default_storage_layout.rs
ggonzalez94: Beautiful 🤩 - much cleaner. Thanks!

lib/motsu-proc/tests/default.rs
ggonzalez94: Agreed. I just updated it to be `mapping_initializes_and_updates`. With this we already have tests for mappings, uints(sorted and unsorted) and addresses. Does it make sense to try to cover all types? I feel pretty confortable with this coverage + the confidence we get from using it in the unit tests. But curious to hear your thoughts @bidzyyys @alexfertel

lib/motsu-proc/tests/default.rs
ggonzalez94: Yeah, I was just initializing the other values with `USomething::from(value)` and forgot about zero lol. Thanks!

alexfertel (approved):
LGTM! 🚀 Thanks for the hard work!

PR #118, opened 2024-06-13
fix(ci): remove useless coverage workflow

alexfertel:
Follow-up to #114 

Removes the end-to-end tests coverage workflow until we have a clearer idea of how to make it work, which may entail delegating to the Stylus codebase.

The rationale is: `llvm-cov` instruments binaries in order to collect data used in computing coverage. The output of the instrumented binaries is written to the filesystem, and then uploaded to codecov.io.

Writing to the filesystem is not an operation allowed in the sandboxed webassembly environment where contracts run, namely, the host (Stylus VM).

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -1,12 +1,10 @@
-# This workflow runs our end-to-end tests suite and collects coverage.
+# This workflow runs our end-to-end tests suite.
 #
 # It roughly follows these steps:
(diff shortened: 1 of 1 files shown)


qalisander (approved):

ggonzalez94 (approved):
LGTM. It would have great to have this, I hope we find a way to add it

bidzyyys (approved):
LGTM

PR #119, opened 2024-06-13
feat: bitmap

### LINKED ISSUE #77: [Feature]: BitMaps

BitMaps library contract ([solidity](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/3e6c86392c97fbc30d3d20a378a6f58beba08eba/contracts/utils/structs/BitMaps.sol))

qalisander:
Resolves #77

- [x] Tests
- [x] Documentation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,5 +9,3 @@ docs/build/
 **/.DS_Store
 
 **/nitro-testnode
-
-**/proptest-regressions
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ docs/build/
 **/.DS_Store
 
 **/nitro-testnode
+
+**/proptest-regressions
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,7 @@ keywords = ["arbitrum", "ethereum", "stylus"]
 repository = "https://github.com/OpenZeppelin/rust-contracts-stylus"
 
 [workspace.dependencies]
-alloy-primitives = { version = "0.3.1", default-features = false, features = ["arbitrary"] }
+alloy-primitives = { version = "0.3.1", default-features = false }
 alloy-sol-types = { version = "0.3.1", default-features = false }
 # TODO: lift stylus-sdk and stylus-proc versions once nitro test node will be compatible
 stylus-sdk = { version = "0.4.3", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,28 +1,28 @@
 [workspace]
 members = [
-  "contracts",
-  "lib/crypto",
-  "lib/motsu",
-  "lib/motsu-proc",
-  "examples/erc20",
-  "examples/erc721",
-  "examples/merkle-proofs",
-  "examples/ownable",
-  "examples/access-control",
-  "lib/e2e",
-  "lib/e2e-proc",
+    "contracts",
+    "lib/crypto",
+    "lib/motsu",
+    "lib/motsu-proc",
+    "examples/erc20",
+    "examples/erc721",
+    "examples/merkle-proofs",
+    "examples/ownable",
+    "examples/access-control",
+    "lib/e2e",
+    "lib/e2e-proc",
 ]
 default-members = [
-  "contracts",
-  "lib/crypto",
-  "lib/motsu",
-  "lib/motsu-proc",
-  "examples/erc20",
-  "examples/erc721",
-  "examples/merkle-proofs",
-  "examples/ownable",
-  "examples/access-control",
-  "lib/e2e-proc",
+    "contracts",
+    "lib/crypto",
+    "lib/motsu",
+    "lib/motsu-proc",
+    "examples/erc20",
+    "examples/erc721",
+    "examples/merkle-proofs",
+    "examples/ownable",
+    "examples/access-control",
+    "lib/e2e-proc",
 ]
 
 # Explicitly set the resolver to version 2, which is the default for packages
@@ -38,7 +38,7 @@ keywords = ["arbitrum", "ethereum", "stylus"]
 repository = "https://github.com/OpenZeppelin/rust-contracts-stylus"
 
 [workspace.dependencies]
-alloy-primitives = { version = "0.3.1", default-features = false }
+alloy-primitives = { version = "0.3.1", default-features = false, features = ["arbitrary"] }
 alloy-sol-types = { version = "0.3.1", default-features = false }
 # TODO: lift stylus-sdk and stylus-proc versions once nitro test node will be compatible
 stylus-sdk = { version = "0.4.3", default-features = false }
@@ -46,15 +46,15 @@ stylus-proc = { version = "0.4.3", default-features = false }
 mini-alloc = "0.4.2"
 
 alloy = { git = "https://github.com/alloy-rs/alloy", rev = "0928b92", features = [
-  "contract",
-  "network",
-  "providers",
-  "provider-http",
-  "rpc-client",
-  "rpc-types-eth",
-  "signers",
-  "signer-wallet",
-  "getrandom",
+    "contract",
+    "network",
+    "providers",
+    "provider-http",
+    "rpc-client",
+    "rpc-types-eth",
+    "signers",
+    "signer-wallet",
+    "getrandom",
 ] }
 
 [profile.release]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,28 +1,28 @@
 [workspace]
 members = [
-    "contracts",
-    "lib/crypto",
-    "lib/motsu",
-    "lib/motsu-proc",
-    "examples/erc20",
-    "examples/erc721",
-    "examples/merkle-proofs",
-    "examples/ownable",
-    "examples/access-control",
-    "lib/e2e",
-    "lib/e2e-proc",
+  "contracts",
+  "lib/crypto",
+  "lib/motsu",
+  "lib/motsu-proc",
+  "examples/erc20",
+  "examples/erc721",
+  "examples/merkle-proofs",
+  "examples/ownable",
+  "examples/access-control",
+  "lib/e2e",
+  "lib/e2e-proc",
 ]
 default-members = [
-    "contracts",
-    "lib/crypto",
-    "lib/motsu",
-    "lib/motsu-proc",
-    "examples/erc20",
-    "examples/erc721",
-    "examples/merkle-proofs",
-    "examples/ownable",
-    "examples/access-control",
-    "lib/e2e-proc",
+  "contracts",
+  "lib/crypto",
+  "lib/motsu",
+  "lib/motsu-proc",
+  "examples/erc20",
+  "examples/erc721",
+  "examples/merkle-proofs",
+  "examples/ownable",
+  "examples/access-control",
+  "lib/e2e-proc",
 ]
 
 # Explicitly set the resolver to version 2, which is the default for packages
@@ -46,15 +46,15 @@ stylus-proc = { version = "0.4.3", default-features = false }
 mini-alloc = "0.4.2"
 
 alloy = { git = "https://github.com/alloy-rs/alloy", rev = "0928b92", features = [
-    "contract",
-    "network",
-    "providers",
-    "provider-http",
-    "rpc-client",
-    "rpc-types-eth",
-    "signers",
-    "signer-wallet",
-    "getrandom",
+  "contract",
+  "network",
+  "providers",
+  "provider-http",
+  "rpc-client",
+  "rpc-types-eth",
+  "signers",
+  "signer-wallet",
+  "getrandom",
 ] }
 
 [profile.release]
--- a/contracts/Cargo.toml
+++ b/contracts/Cargo.toml
@@ -17,6 +17,7 @@ mini-alloc.workspace = true
 cfg-if = "1.0"
 
 [dev-dependencies]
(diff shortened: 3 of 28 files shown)


alexfertel (commented):
Looks good at a first glance, will do a proper review when I get some time.

One question I have is that I think this is a good candidate for implementing `StorageType`, right? What do you think?

contracts/src/utils/structs/bitmap.rs
```diff
@@ -0,0 +1,119 @@
+//! Contract module for managing uint256 to bool mapping in a compact and
```
bidzyyys, line 1: `U256` instead of `uint256`

contracts/src/utils/structs/bitmap.rs
```diff
@@ -0,0 +1,119 @@
...
+//! efficient way, provided the keys are sequential. Largely inspired by Uniswap's <https://github.com/Uniswap/merkle-distributor/blob/master/contracts/MerkleDistributor.sol>[merkle-distributor].
+//!
+//! `BitMap` packs 256 booleans across each bit of a single 256-bit slot of
+//! `uint256` type. Hence, booleans corresponding to 256 _sequential_ indices
```
bidzyyys, line 5: `U256`

contracts/src/utils/structs/bitmap.rs
```diff
@@ -0,0 +1,119 @@
...
+//! would only consume a single slot, unlike the regular `bool` which would
```
bidzyyys, line 6: Here you have `'bool'`, in other places you have just `bool`

contracts/src/utils/structs/bitmap.rs
```diff
@@ -0,0 +1,119 @@
...
+            self.unset(index);
+        }
+    }
+
+    /// Sets the bit at `index`.
+    ///
+    /// # Arguments
+    ///
+    /// * `index` - index of boolean value that should be set `true`.
+    pub fn set(&mut self, index: U256) {
```
bidzyyys, line 57: Most of the code for `set` and `unset` is common, only last line differs. IMHO you can reuse the code.

bidzyyys (commented):
Looks good ✅

qalisander:
> One question I have is that I think this is a good candidate for implementing `StorageType`, right? What do you think?
I think yep. I can reuse this contract to build consecutive on top. We don't need to use stylus inheritance just aggregation

contracts/src/utils/structs/bitmap.rs
qalisander: left just "boolean" word

contracts/src/utils/structs/bitmap.rs
qalisander: Agree. It was exactly like that at solidity
Added methods `get_mask` and `get_bucket` to reuse a bit of code

contracts/src/utils/structs/bitmap.rs
```diff
@@ -0,0 +1,140 @@
...
+        fn default() -> Self {
+            let root = U256::ZERO;
+            BitMap { _data: unsafe { StorageMap::new(root, 0) } }
+        }
+    }
+
+    #[motsu::test]
+    fn set_value() {
+        proptest!(|(value: U256)| {
+            let mut bit_map = BitMap::default();
```
qalisander, line 112: This one seems a bit debatable decision, but now `#[motsu::test]` macro locks storage for a test execution even if there is no contract argument.
Assuming that rust contract struct can be created during test.
And the state of this contract will be single per test.
What do you think @alexfertel ? Or may be use `with_context` function?

Cargo.toml
bidzyyys, line 1: @alexfertel @qalisander we need to have consistent formatting for `toml` files.

contracts/Cargo.toml
```diff
@@ -17,6 +17,7 @@ mini-alloc.workspace = true
 cfg-if = "1.0"
 
 [dev-dependencies]
+alloy-primitives = { version = "0.3.1", features = ["arbitrary"] }
```
bidzyyys, line 20: I believe alloy is configured in main `Cargo.toml`

contracts/src/utils/structs/bitmap.rs
```diff
@@ -0,0 +1,140 @@
...
+            prelude::{Arbitrary, ProptestConfig},
+            proptest,
+        },
+        U256,
+    };
+    use stylus_sdk::{prelude::*, storage::StorageMap};
+
+    use crate::utils::structs::bitmap::BitMap;
+
+    impl Default for BitMap {
```
bidzyyys, line 102: You should use our new `DefaultStorageLayout`, check this: https://github.com/OpenZeppelin/rust-contracts-stylus/commit/9b1c54d96fc865eca6c6934630d636803e1bb49a

contracts/src/utils/structs/bitmap.rs
```diff
@@ -0,0 +1,140 @@
...
+impl BitMap {
+    /// Returns whether the bit at `index` is set.
+    ///
+    /// # Arguments
+    ///
+    /// * `index` - index of boolean value at the bit map.
+    #[must_use]
+    pub fn get(&self, index: U256) -> bool {
+        let bucket = index >> 8;
+        let mask = U256::from(1) << (index & U256::from(0xff));
```
bidzyyys, line 33: imho you can reuse here `self.get_mask(index)`

contracts/src/utils/structs/bitmap.rs
```diff
@@ -0,0 +1,140 @@
...
+
+impl BitMap {
+    /// Returns whether the bit at `index` is set.
+    ///
+    /// # Arguments
+    ///
+    /// * `index` - index of boolean value at the bit map.
+    #[must_use]
+    pub fn get(&self, index: U256) -> bool {
+        let bucket = index >> 8;
```
bidzyyys, line 32: you can reuse here `self.get_bucket(index)`

contracts/src/utils/structs/bitmap.rs
```diff
@@ -0,0 +1,140 @@
...
+        }
+    }
+
+    /// Sets the bit at `index`.
+    ///
+    /// # Arguments
+    ///
+    /// * `index` - index of boolean value that should be set `true`.
+    pub fn set(&mut self, index: U256) {
+        let bucket = Self::get_bucket(index);
```
bidzyyys, line 58:
```rust
let bucket = Self::get_bucket(index);
let mask = Self::get_mask(index);
let mut value = self._data.setter(bucket);
let prev = value.get();
```
Is common code for `set` and `unset`

contracts/src/utils/structs/bitmap.rs
qalisander: thnx!

contracts/src/utils/structs/bitmap.rs
qalisander: nice that it was merged!

contracts/Cargo.toml
qalisander: It was used without arbitrary feature. Since we're using resolver = 2 we can plugin features necessary just for testing purposes

contracts/src/utils/structs/bitmap.rs
qalisander: I would say yep, but I don't think we will win much code lanes refactoring it further. It is still looks like solidity counterpart

qalisander: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/53b5d84212b73388a2a7096ff503693dae692726/contracts/utils/structs/BitMaps.sol#L18

Cargo.toml
qalisander: yep. I'll remove auto format for `toml` files in my ide:)

contracts/src/utils/structs/bitmap.rs
alexfertel: I think it's fine to keep it duplicated, it's only in two places, and we won't be reusing that code anywhere else.

.gitignore
```diff
@@ -9,3 +9,5 @@ docs/build/
 **/.DS_Store
 
 **/nitro-testnode
+
+**/proptest-regressions
```
alexfertel, line 13: We shouldn't ignore these! It makes failure cases load faster the next time tests run 👀

contracts/src/utils/structs/bitmap.rs
```diff
@@ -0,0 +1,134 @@
...
+//!
+//! `BitMap` packs 256 booleans across each bit of a single 256-bit slot of
+//! `U256` type. Hence, booleans corresponding to 256 _sequential_ indices
+//! would only consume a single slot, unlike the regular boolean which would
+//! consume an entire slot for a single value.
+//!
+//! This results in gas savings in two ways:
+//!
+//! - Setting a zero value to non-zero only once every 256 times
+//! - Accessing the same warm slot for every 256 _sequential_ indices
```
alexfertel, line 12:
```suggestion
//! Contract module for managing `U256` to boolean mapping in a compact and
//! efficient way, provided the keys are sequential. Largely inspired by Uniswap's 
//! [merkle-distributor].
//!
//! `BitMap` packs 256 booleans across each bit of a single 256-bit slot of
//! `U256` type. Hence, booleans corresponding to 256 _sequential_ indices
//! would only consume a single slot, unlike the regular boolean which would
//! consume an entire slot for a single value.
//!
//! This results in gas savings in two ways:
//!
//! - Setting a zero value to non-zero only once every 256 times
//! - Accessing the same warm slot for every 256 _sequential_ indices
//!
//! [merkle-distributor]: https://github.com/Uniswap/merkle-distributor/blob/master/contracts/MerkleDistributor.sol
```

contracts/src/utils/structs/bitmap.rs
```diff
@@ -0,0 +1,134 @@
...
+        mapping(uint256 => uint256) _data;
+    }
+}
+
+impl BitMap {
+    /// Returns whether the bit at `index` is set.
+    ///
+    /// # Arguments
+    ///
+    /// * `index` - index of boolean value at the bit map.
```
alexfertel, line 30:
```suggestion
    /// * `index` - index of the boolean value in the bit map.
```

contracts/src/utils/structs/bitmap.rs
```diff
@@ -0,0 +1,134 @@
...
+        let value = self._data.get(bucket);
+        (value & mask) != U256::ZERO
+    }
+
+    /// Sets the bit at `index` to the boolean `value`.
+    ///
+    /// # Arguments
+    ///
+    /// * `index` - index of boolean value at the bit map.
+    /// * `value` - boolean value to set into the bit map.
```
alexfertel, line 44:
```suggestion
    /// * `index` - index of boolean value in the bit map.
    /// * `value` - boolean value to set in the bit map.
```

contracts/src/utils/structs/bitmap.rs
```diff
@@ -0,0 +1,134 @@
...
+        let bucket = Self::get_bucket(index);
+        let mask = Self::get_mask(index);
+        let mut value = self._data.setter(bucket);
+        let prev = value.get();
+        value.set(prev & !mask);
+    }
+
+    /// Get mask of value in the bucket.
+    fn get_mask(index: U256) -> U256 {
+        U256::from(1) << (index & U256::from(0xff))
```
alexfertel, line 81: Maybe extract `U256::from(1)` and `U256::from(0xff)` as constants to make this a bit faster.

```rust
pub const ONE: U256 = U256::from_limbs([1, 0, 0, 0]);
pub const UNSURE_WHAT_NAME: U256 = U256::from_limbs([255, 0, 0, 0]);
```

contracts/src/utils/structs/bitmap.rs
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    /// Get bucket index.
+    fn get_bucket(index: U256) -> U256 {
```
alexfertel, line 85: Should we `#[inline]` these? They are private & small.

contracts/src/utils/structs/bitmap.rs
alexfertel: I think `with_context` is slightly cleaner because we are instantiating the contract twice this way (with `default`), right? 

In any case, I'm fine either way, I'll leave it to your judgment!

contracts/src/lib.rs
```diff
@@ -4,6 +4,7 @@
 #![allow(clippy::pub_underscore_fields, clippy::module_name_repetitions)]
 #![cfg_attr(not(feature = "std"), no_std, no_main)]
 extern crate alloc;
+extern crate core;
```
alexfertel, line 7: Are you sure we need `core`? It's my understanding that this is always imported.

alexfertel (approved):
Looks great! 🚀 

Besides the question about `StorageType` I left a few small changes. In any case, I think it's fine to merge as is, and later update it if we need to.

contracts/src/lib.rs
qalisander: weird how it appeared..

.gitignore
qalisander: make sense

contracts/src/utils/structs/bitmap.rs
qalisander: I used to think that those small and private functions should be unlined by default

Issue #53, opened 2024-05-03
Setup integration tests using `nitro-testnode`


alexfertel:
After [reaching stagnation](https://github.com/OpenZeppelin/rust-contracts-stylus/issues/15#issuecomment-2093873884) in #15, we decided to setup temporary infrastructure for integration tests that uses `nitro-testnode`.

PR #120, opened 2024-06-13
ci: ignore nitro test node directory

qalisander:
Since in previously local nitro-testnode clone was lifted up will add a new path to .gitignore
Follow ups #53

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,4 +8,4 @@ docs/build/
 
 **/.DS_Store
 
-**/nitro-testnode
+/nitro-testnode


.gitignore
```diff
@@ -8,4 +8,4 @@ docs/build/
 
 **/.DS_Store
 
-**/nitro-testnode
```
alexfertel, line 11: I think `**/nitro-testnode` includes `/nitro-testnode`. Are you sure this changes anything?

I just tried creating a `nitro-testnode` directory with some files inside, and it is properly ignored 👀

alexfertel: Screenshot 2024-06-13 at 18 32 37

qalisander: Just seemed that it changed my ide's coloured highlight for ignored folder. But yep. Same difference

PR #121, opened 2024-06-14
test(erc20): add integration tests

### LINKED ISSUE #94: ERC-20 Integration tests

bidzyyys:
Resolves #94

### DIFF
--- a/examples/erc20/tests/erc20.rs
+++ b/examples/erc20/tests/erc20.rs
@@ -38,7 +38,7 @@ async fn deploy(
 }
 
 // ============================================================================
-// Integration Tests: ERC20
+// Integration Tests: ERC-20 Token + Metadata Extension
 // ============================================================================
 
 #[e2e::test]
@@ -49,12 +49,14 @@ async fn constructs(alice: User) -> Result<()> {
     let Erc20::nameReturn { name } = contract.name().call().await?;
     let Erc20::symbolReturn { symbol } = contract.symbol().call().await?;
     let Erc20::capReturn { cap } = contract.cap().call().await?;
+    let Erc20::decimalsReturn { decimals } = contract.decimals().call().await?;
     let Erc20::totalSupplyReturn { totalSupply: total_supply } =
         contract.totalSupply().call().await?;
 
     assert_eq!(name, TOKEN_NAME.to_owned());
     assert_eq!(symbol, TOKEN_SYMBOL.to_owned());
     assert_eq!(cap, U256::from(CAP));
+    assert_eq!(decimals, 10);
     assert_eq!(total_supply, U256::ZERO);
     Ok(())
 }
@@ -597,7 +599,7 @@ async fn transfer_from_rejects_invalid_receiver(
 }
 
 // ============================================================================
-// Integration Tests: ERC20Burnable
+// Integration Tests: ERC-20 Burnable Extension
 // ============================================================================
 
 #[e2e::test]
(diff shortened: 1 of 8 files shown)


lib/e2e/src/error.rs
```diff
@@ -1,23 +1,69 @@
...
+    fn reverts(&self, code: i64) -> bool;
+}
+
+impl RevertExt for alloy::contract::Error {
+    fn reverts(&self, code: i64) -> bool {
+        let Self::TransportError(e) = self else {
+            return false;
+        };
+
+        //  ErrorResp(
```
bidzyyys, line 21: @alexfertel @qalisander any better idea for this kind of error is welcome.

lib/e2e/src/error.rs
```diff
@@ -1,23 +1,69 @@
...
+            .unwrap_or_else(|| panic!("{:#?}", e));
+        // .expect("should extract the error");
         let actual = &raw_value.get().trim_matches('"')[2..];
         let expected = alloy::hex::encode(expected.abi_encode());
-        return expected == actual;
+        expected == actual
+    }
+}
+
+impl<E: SolError> ErrorExt<E> for eyre::ErrReport {
```
bidzyyys, line 50: @alexfertel @qalisander any better idea for this kind of error is welcome.

alexfertel (approved):
Daaamn, these are so cleannnnn 💯 🕶️ 

I'll think about the errors, my first impression is just to special-case them, but I'm not sure. I'll take some time to so what we can do.

qalisander (approved):
I think looks perfect!
Good job!

examples/erc20/tests/erc20.rs
```diff
@@ -16,52 +16,912 @@ sol!("src/constructor.sol");
...
+
+async fn deploy(
+    rpc_url: &str,
+    private_key: &str,
+    cap: Option<U256>,
+) -> eyre::Result<Address> {
+    let cap = match cap {
+        Some(cap) => cap,
+        None => U256::from(CAP),
+    };
```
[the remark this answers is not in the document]
alexfertel: I would go as far as to do something like:

```rust
    let args = Erc20Example::constructorCall {
        name_: TOKEN_NAME.to_owned(),
        symbol_: TOKEN_SYMBOL.to_owned(),
        cap_: cap.unwrap_or(U256::from(CAP)),
    };
```

wdyt?

lib/e2e/src/error.rs
```diff
@@ -1,13 +1,37 @@
-use alloy::{contract::Error, sol_types::SolError};
+use alloy::sol_types::SolError;
 
 pub trait ErrorExt<E> {
     /// Checks that `Self` corresponds to the typed abi-encoded error
     /// `expected`.
-    fn is(&self, expected: E) -> bool;
+    fn is_err(&self, expected: E) -> bool;
```
alexfertel, line 6: I named it `ErrorExt::is()`, because of a few reasons:

- It's defined in an `ErrorExt` trait.
- It is implemented for error-like types, like `alloy::contract::Error` and `eyre::ErrReport`.
- At the call site, it looks like this `err.is(/* Some Error */);`

So it's quite clear we are dealing with errors.

And example API that looks like this is the `Object.is()` from javascript. I think this is better than something like `Object.is_object()`. wdyt?

alexfertel: Well, thinking a bit more about it, we could make it more similar to the contracts by doing something like this:

```rust
/// An error representing a panic.
pub trait Panic {
    /// Checks that `Self` corresponds to a panic with code `code`.
    fn panicked_with(&self, code: i64) -> bool;
}

/// An error representing a revert with some data.
pub trait Revert<E> {
    /// Checks that `Self` corresponds to the typed abi-encoded error
    /// `expected`.
    fn reverted_with(&self, expected: E) -> bool;
}
```

And then we keep a list of possible panic codes, [like this](https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-chai-matchers/src/internal/reverted/panic.ts), which is what the Solidity contracts [use](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/83c7e45092dac350b070c421cd2bf7105616cf1a/test/token/ERC20/ERC20.test.js#L60).

lib/e2e/src/error.rs
alexfertel: Yeah, I think we don't need this one here. Since it's an `ErrReport` from `eyre`, we should be able to traverse the sources until we reach the Transport error.

examples/erc20/tests/erc20.rs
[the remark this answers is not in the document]
bidzyyys: I like it!

lib/e2e/src/error.rs
bidzyyys: Will do!

bidzyyys: Regarding `is` and `is_err` I agree that `is` is a better option, but I had a name collision for `eyre::ErrReport`. Will try traverse the `ErrReport` to `TransportError` and then I can revert `err.is()`.

qalisander: We made a long a way: `err.assert(...)` -> `err.is(...)` -> `err.is_err(...)` -> ... Curious what will be next:)
`err.is(...)` seems good for me

alexfertel: naming is hard 😢

alexfertel (approved):
Great job! Super clean 💯

PR #123, opened 2024-06-16
feat(crypto): implement ECDSA for P256

### LINKED ISSUE #17: ECDSA
```[tasklist]
### Tasks
- [ ] https://github.com/OpenZeppelin/rust-contracts-stylus/issues/124
```

### LINKED ISSUE #124: P256 elliptic curve arithmetic

alexfertel:
Resolves #124  

Adds an implementation of elliptic curve operations that is generic over the curve. It's implemented for `P256` (AKA `secp256r1`), and some of the operations are hard-coded for this curve, so it's not fully generic, but the foundation is there.

For a primer on elliptic curves, [this series of blog posts](https://andrea.corbellini.name/2015/05/17/elliptic-curve-cryptography-a-gentle-introduction/) is the best resource I could find.

## Inspiration

The implementation and code structure are largely inspired by `RustCrypto` with some simplifications and cleanup. The rationale for taking this as the reference implementation was: 

- Readability, which translates into auditability.
- The crates are actively maintained.
- They are widely used. For example, the `ecdsa` crate has 28 million all-time downloads.
- Some of the crates in the ecosystem already use them, like `alloy`.

## Notes

I've decided that private key generation is out of scope since:

- There are other well-established and secure ways to do it.
- It would introduce many threat vectors.
- We would need to include a source of randomness, which on-chain is quite tricky. 
- Our aim is to have something like `ecrecover` and `ecdsa` but for `P256`, so we can skip private key generation.

#### PR Checklist

- [ ] Improve error handling.
- [ ] ECDSA `sign` & `verify`. I'll probably leave this for a separate PR, to manage scope, but unsure yet.
- [ ] Polish code structure and docs.
- [ ] Remove dead code.
- [ ] Tests
- [ ] Documentation


lib/crypto/proptest-regressions/elliptic-curve/field.txt
alexfertel, line 1: Unsure if we should ignore this file, probably yes, I have to research first.

lib/crypto/src/ecdsa.rs
bidzyyys, line 1: Should we associate these function with some trait? Just thinking...

lib/crypto/src/elliptic-curve/test_vectors/field.rs
```diff
@@ -0,0 +1,255 @@
+//! Test vectors for the secp256r1 base field.
```
bidzyyys, line 1: How is it generated? Any tool? Trying to figure out edge cases?

lib/crypto/src/elliptic-curve/keys.rs
```diff
@@ -0,0 +1,43 @@
...
+        Self { point }
+    }
+
+    /// Compute a [`PublicKey`] from a secret scalar value (i.e. a secret key
+    /// represented as a raw scalar value). Returns `None` if the value computed
+    /// is the identity point.
+    pub fn from_secret_scalar(scalar: &C::Uint) -> Option<Self> {
+        (&ProjectivePoint::GENERATOR * scalar)
+            .to_affine()
+            .map(|point| Self { point })
```
bidzyyys, line 34: How about implementing `From` trait here? It is just a wrapper.

lib/crypto/src/signature/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+/// resulting in a signature with a total of 64-bytes.
+#[derive(Clone, Eq, PartialEq)]
+pub struct Signature<C: PrimeCurve> {
+    r: C::Uint,
+    s: C::Uint,
+}
+
+impl<C: PrimeCurve> fmt::Debug for Signature<C> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "ecdsa::Signature<{:?}>(", C::default())?;
```
bidzyyys, line 30: Should we use here `C::default()?`

lib/crypto/src/signature/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+
+        for byte in self.to_bytes() {
```
bidzyyys, line 32: How about using https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.for_each

lib/crypto/src/signature/error.rs
```diff
@@ -0,0 +1,46 @@
+//! Signature error type.
+
+use core::fmt::{self, Debug, Display};
+
+// TODO: Once we update `alloy` to a crates.io version, use the `signature`
```
bidzyyys, line 5: IMHO we should have an issue for that.

lib/crypto/src/signature/error.rs
```diff
@@ -0,0 +1,46 @@
...
+/// things like remote signers (e.g. HSM, KMS) to report I/O or auth errors.
+///
+/// [BB'06]: https://en.wikipedia.org/wiki/Daniel_Bleichenbacher
+#[derive(Default)]
+#[non_exhaustive]
+pub struct Error {}
+
+impl Error {
+    /// Create a new error with no associated source
+    pub fn new() -> Self {
```
bidzyyys, line 29: Should we add some explanation in error what happened wrong during signature check?

bidzyyys (commented):
Looks great 🚀

lib/crypto/src/ecdsa.rs
alexfertel: Yeah, I plan to have something like an `ECDSA` struct generic over the curve.

lib/crypto/src/elliptic-curve/test_vectors/field.rs
alexfertel: So, these are just the repeated doubling of the multiplicative identity. That is, `1, 2, 4, 8, 16, 32`, so basically the powers of two.

lib/crypto/src/elliptic-curve/keys.rs
alexfertel: Sadly, this doesn't have the same signature as either `From` (returns `Self`) or `TryFrom` (returns `Result<Self, ..>`.

The reason is that you can end up with the identity when converting from projective coordinates to affine coordinates, but we don't want to leak that to the user, so I decided to return `Option<Self>`.

This is not a strong opinion, I will probably end up returning a `Result<Self, ..>`, but I haven't figured out the best way to do this. If I go this route, I'll turn it into a `TryFrom` implementation.

lib/crypto/src/signature/mod.rs
alexfertel: Yes! Remember that `C` looks like this:

```rust
#[derive(Copy, Clone, Debug, Default, Eq, PartialEq, PartialOrd, Ord)]
pub struct P256;
```

Which means the output of `fmt` will be something like:

```rust
ecdsa::Signature<P256>(/* signature */)
```

Which I think is quite nice tbh.

lib/crypto/src/signature/mod.rs
alexfertel: I prefer plain for loops, since they are easier to refactor. However, if you have a strong reason, let me know, and I'm happy to change it!

lib/crypto/src/signature/error.rs
alexfertel: These are opaque errors, we don't return any information that may enable a sidechannel attack.

It's explained in line 14 of that file, in the documentation for the struct.

```rust
/// Signature errors.
///
/// This type is deliberately opaque as to avoid sidechannel leakage which
/// could potentially be used recover signing private keys or forge signatures
/// (e.g. [BB'06]).
///
/// [BB'06]: https://en.wikipedia.org/wiki/Daniel_Bleichenbacher
```

Cargo.lock
```diff
@@ -1498,6 +1499,17 @@ dependencies = [
  "zeroize",
 ]
 
+[[package]]
+name = "crypto-bigint"
```
qalisander, line 1503: Also I can see we're using two different versions of `crypto-bigint` library. "0.5.5" and `0.6.0-..`

lib/crypto/src/elliptic-curve/projective.rs
```diff
@@ -0,0 +1,326 @@
+//! Projective curve points.
+use core::{
```
qalisander, line 2: Make sense to `projective` and `afine` into `point` module. Wdt?

qalisander: I think `arithmetic` module also can be put there

qalisander (commented):
Great progress so far!
That kind of things sometimes cumbersome to implement, but as a reward usually easy to test.

lib/crypto/src/elliptic-curve/field.rs
```diff
@@ -0,0 +1,455 @@
...
+    /// Returns true iff this element is zero.
+    fn is_zero(&self) -> bool;
+
+    /// Doubles this element.
+    #[must_use]
+    fn double(&self) -> Self;
+
+    /// Squares this element.
+    #[must_use]
+    fn square(&self) -> Self;
```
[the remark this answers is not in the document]
alexfertel: No reason! This is nicer!

lib/crypto/src/elliptic-curve/field.rs
```diff
@@ -0,0 +1,455 @@
...
+
+    /// Computes the multiplicative inverse of this element,
+    /// failing if the element is zero.
+    #[must_use]
+    fn invert(&self) -> Option<Self>;
+}
+
+/// An element in the subgroup with base point [`P256::GENERATOR`].
+#[derive(Clone, Copy, Debug)]
+pub struct FieldElement(pub U256);
```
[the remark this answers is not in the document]
alexfertel: Yes, but I want to first try to make this generic over the curve as well to simplify it a lot.

lib/crypto/src/elliptic-curve/projective.rs
alexfertel: Makes sense 👍

Cargo.lock
alexfertel: Good catch, I need to figure out why

lib/crypto/src/signature/error.rs
alexfertel: Actually, I realize this comment is wrong, since we don't want to depend on `alloy` here. I'll remove it.

lib/crypto/src/elliptic-curve/field.rs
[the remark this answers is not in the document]
alexfertel: Nvm, that would introduce cyclic traits 😢

lib/crypto/src/elliptic-curve/keys.rs
bidzyyys: Sounds good.

bidzyyys (commented):
Looks good 🚀

PR #126, opened 2024-06-17
feat: use v0.1.1 of alloy from crates.io

### LINKED ISSUE #125: feat: use a published version of alloy
We are currently using `alloy` via a git dependency, which is not ideal. For example, we wouldn't be able to publish crates depending on it.

Now that there's a published version of alloy on crates.io, we should start using it.

alexfertel:
Resolves #125

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,24 +38,26 @@ keywords = ["arbitrum", "ethereum", "stylus"]
 repository = "https://github.com/OpenZeppelin/rust-contracts-stylus"
 
 [workspace.dependencies]
-alloy-primitives = { version = "0.3.1", default-features = false }
-alloy-sol-types = { version = "0.3.1", default-features = false }
 # TODO: lift stylus-sdk and stylus-proc versions once nitro test node will be compatible
 stylus-sdk = { version = "0.4.3", default-features = false }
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -38,24 +38,26 @@ keywords = ["arbitrum", "ethereum", "stylus"]
...
   "provider-http",
   "rpc-client",
   "rpc-types-eth",
-  "signers",
-  "signer-wallet",
+  "signer-local",
   "getrandom",
 ] }
+# Even though `alloy` includes `alloy-primitives` and `alloy-sol-types` we need
+# to keep both versions for compatibility with the SDK. Once they start using
```
ggonzalez94, line 57:
```suggestion
# to keep both versions for compatibility with the Stylus SDK. Once they start using
```

ggonzalez94 (approved):
LGTM! Less than a day since the version was published and we already have a PR 🚀

bidzyyys (approved):
LGTM 🟢 🚀

qalisander (approved):

Cargo.toml
alexfertel: ooops, missed this one. I'll add it in a separate PR.

PR #132, opened 2024-06-18
feat: use v0.1.0 of koba from crates.io

### LINKED ISSUE #131: [Feature] Use a published version of koba
We are currently using `koba` via a git dependency, which is not ideal. For example, we wouldn't be able to publish crates depending on it.

Now that there's a published version of alloy on crates.io, we should start using it.

alexfertel:
Resolves #131


bidzyyys (approved):
LGTM!

qalisander (approved):

PR #143, opened 2024-06-21
chore: enforce uint macro for const evaluation

qalisander:
Curious why we are not using const evaluation for alloy number types wherever is possible?


alexfertel (approved):
Oooh, this is great! I didn't know it existed 💯

PR #148, opened 2024-06-22
feat: add full user journey example

### LINKED ISSUE #140: [Feature]: Create an e2e example of using one of the contracts

We don't currently have any place where we document how to use our contracts end-to-end. Our e2e tests are a showcase of this, but we should also create some form of tutorial or script that showcases everything that's needed and when run deploys to Sepolia.

### Contribution Guidelines

alexfertel:
Resolves #140, #149 

This also fixes a long-standing issue we had where we couldn't compile contracts without some hacky feature flags because of multiple panic handler implementations. It turns out the fix was `target_arch = "wasm32"` instead of `target_arch = "wasm32-unknown-unknown"`.

#### PR Checklist

- ~[ ] Tests~
- [x] Documentation


examples/basic/README.md
```diff
@@ -0,0 +1,47 @@
+# Basic Token Example
+
+This example showcases an end-to-end user journey of how to write, deploy and
+query a smart contract written using our library.
+
+## Running the example
+
+There are two crates in the example: a library crate for our contract
+implementation, which is a simple `ERC-20` token extended with metadata, and
```
bidzyyys, line 9: `Metadata` extension

bidzyyys (approved):
Great job @alexfertel 🚀

bidzyyys (approved):
LGTM! 🚀

qalisander (approved):
Good job!

alexfertel:
> I think we should next target into separate cli tool for contract deployment. To be able to deploy any contract in examples directory.
Similar to cargo stylus but that seeks for constructor.sol file and injects it into wasm.

Yes! Ideally, this would be `cargo-stylus` itself, that's why I asked the team at Offchain Labs if we could convert it to a library, so it's easy to integrate with. Hopefully, it's possible soon.

PR #150, opened 2024-06-22
ref: aggregate deps in root manifest

### LINKED ISSUE #144: [Feature]: Aggregate dependencies in root Cargo.toml

We are currently specifying most dependencies of member crates directly, but we should instead define the dependency in the root Cargo.toml, and then do `DEPENDENCY_NAME.workspace = true` in member crates.

The rationale is that this way, we ensure we're using a single version of a dependency.

This may not make sense for all dependencies, so this should be judge on a case-by-case basis.

### Contribution Guidelines

developeruche:
Moved all dependency importation to the root workspace `Cargo.toml` and referenced it at the crate level.

Resolves #144 

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/lib/crypto/Cargo.toml
+++ b/lib/crypto/Cargo.toml
@@ -13,7 +13,7 @@ mini-alloc.workspace = true
 tiny-keccak.workspace = true
 
 [dev-dependencies]
-hex-literal.workspace = true
+hex-literal = "0.4.1"
 rand.workspace = true
 
 [features]
--- a/lib/crypto/Cargo.toml
+++ b/lib/crypto/Cargo.toml
@@ -10,11 +10,11 @@ version = "0.1.0"
 
 [dependencies]
 mini-alloc.workspace = true
-tiny-keccak = { version = "2.0.2", features = ["keccak"] }
+tiny-keccak.workspace = true
 
 [dev-dependencies]
-hex-literal = "0.4.1"
-rand = "0.8.5"
+hex-literal.workspace = true
+rand.workspace = true
 
 [features]
 std = []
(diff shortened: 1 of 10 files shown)


lib/crypto/Cargo.toml
```diff
@@ -10,11 +10,11 @@ version = "0.1.0"
 
 [dependencies]
 mini-alloc.workspace = true
-tiny-keccak = { version = "2.0.2", features = ["keccak"] }
+tiny-keccak.workspace = true
 
 [dev-dependencies]
-hex-literal = "0.4.1"
```
alexfertel, line 16: Let's keep this one here.

lib/e2e/Cargo.toml
```diff
@@ -10,13 +10,13 @@ repository.workspace = true
...
-regex = "1.10.4"
-once_cell = "1.19.0"
-koba = "0.1.0"
+tokio.workspace = true
+eyre.workspace = true
+regex.workspace = true
+once_cell.workspace = true
+koba.workspace = true
 e2e-proc = { path = "../e2e-proc" }
-toml = "0.8.13"
```
alexfertel, line 19: Let's keep this one here as well.

alexfertel (changes requested):
Looks good! 🚀 

Let's keep `toml` and `hex-literal` in the crates that use them, otherwise this looks ready to merge.

lib/e2e/Cargo.toml
```diff
@@ -10,13 +10,13 @@ repository.workspace = true
 
 [dependencies]
 alloy.workspace = true
-tokio = { version = "1.12.0", features = ["full", "process"] }
```
alexfertel, line 13: Let's keep this one as well, since it uses both `features = ["full", "process"]` and set it to only use `features = ["full"]` in the workspace root.

alexfertel (approved):
LGTM 🚀

bidzyyys (approved):
I like it, good job @developeruche 👏

PR #151, opened 2024-06-23
fix: remove most contract feature flags

alexfertel:
Experiment to check contract binary sizes with and without feature flags.

### DIFF
--- a/contracts/Cargo.toml
+++ b/contracts/Cargo.toml
@@ -24,25 +24,6 @@ rand = "0.8.5"
 [features]
 default = []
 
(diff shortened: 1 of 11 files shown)


ggonzalez94 (approved):
This is much cleaner! I remember we had this discussion a few weeks ago(if they were really necessary). I'm just curious how you were able to test they were not

bidzyyys (approved):
LGTM! Now the usage of our library will be much easier! 🚀

qalisander (approved):
Great! Looks good!

PR #161, opened 2024-06-27
refactor: rename tests with error assertions

### LINKED ISSUE #155: [Refactor]: apply same naming convention for all tests that check errors

We should apply the same naming convention for all tests that check errors:
`fn error_when_<function>_<description>(...)`

### Contribution Guidelines

postmeback:
Changes as per the mentioned issue. It's a work in progress, requesting a quick review. Will continue working on it, if the initial changes are fine.

Resolves #155 

#### PR Checklist

The below items are not applicable as such, because it is more about renaming tests to a particular naming convention.

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/src/token/erc721/mod.rs
+++ b/contracts/src/token/erc721/mod.rs
@@ -1129,7 +1129,7 @@ mod tests {
     }
 
     #[motsu::test]
-    fn error_when_checking_balance_of_invalid_owner(contract: Erc721) {
+    fn error_when_balance_of_invalid_owner(contract: Erc721) {
         let invalid_owner = Address::ZERO;
         let err = contract
             .balance_of(invalid_owner)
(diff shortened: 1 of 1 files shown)


contracts/src/token/erc721/mod.rs
```diff
@@ -1141,21 +1141,18 @@ mod tests {
     }
 
     #[motsu::test]
-    fn balance_of_zero_balance(contract: Erc721) {
+    fn error_when_contract_balance_is_zero_balance(contract: Erc721) {
```
qalisander, line 1144: I can see this test is not checking any error. I think we are not expecting to change its name. cc @bidzyyys

contracts/src/token/erc721/mod.rs
```diff
@@ -1182,17 +1178,17 @@ mod tests {
...
     #[motsu::test]
-    fn error_when_minting_token_id_twice(contract: Erc721) {
+    fn error_when_minting_same_token_id_twice(contract: Erc721) {
         let alice = msg::sender();
-        let token_id = random_token_id();
+        let token_id = random_token_id();   
         contract
             ._mint(alice, token_id)
-            .expect("should mint the token a first time");
+            .expect("should mint the token the first time");
```
qalisander, line 1191: good

contracts/src/token/erc721/mod.rs
```diff
@@ -1165,14 +1162,13 @@ mod tests {
     }
 
     #[motsu::test]
-    fn mints(contract: Erc721) {
+    fn when_minting_token_balance_and_owner_are_correct(contract: Erc721) {
```
qalisander, line 1165: In this test we're also not checking for any error. Seems make sense to leave it as it is. cc @bidzyyys

qalisander (commented):
Hey @postmeback!
Appreciate your contribution!
Here are few comments I have:

contracts/src/token/erc721/mod.rs
bidzyyys: `mints` is fine here. The task is to rename tests that checks errors.

contracts/src/token/erc721/mod.rs
bidzyyys: Yep, we should rename only when we check errors in tests.

alexfertel:
Hey @postmeback! We're really sorry about this, but after some discussion, we found that renaming the tests this way hides away the intent of the test, so we don't think it's worth the trade-off.

As such, I'll be closing the PR, since we won't be proceeding with these changes anymore. Hopefully this doesn't deter you from contributing to our library! Thanks.

PR #162, opened 2024-06-27
feat: add the Trace160 contract to keep checkpoints

### LINKED ISSUE #76: [Feature]: Checkpoints

Checkpoints library contract ([solidity](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/d2ba1f625151c5ee63d15ab3870c31e8bf6e2801/contracts/utils/structs/Checkpoints.sol))

qalisander:
Checkpoint library contract that reflects [solidity counterpart](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/c3f8b760ad9d0431f33e8303658ccbdcc1610f24/contracts/utils/structs/Checkpoints.sol#L16).

Resolves #76

#### PR Checklist

- [x] Tests
- [x] Documentation


qalisander:
Add more tests to lift the code coverage

contracts/src/utils/structs/checkpoints.rs
```diff
@@ -0,0 +1,462 @@
...
+use alloy_primitives::{uint, Uint, U256, U32};
+use alloy_sol_types::sol;
+use stylus_proc::{sol_storage, SolidityError};
+use stylus_sdk::storage::{StorageGuard, StorageGuardMut};
+
+use crate::utils::math::alloy::Math;
+
+// TODO#q: add generics for other pairs (uint32, uint224) and (uint48, uint208).
+// Logic should be same.
+type U96 = Uint<96, 2>;
```
qalisander, line 19: Currently implemented for key: U96, value: U160 types. Other types has the same logic for implementation and make sense not to copy paste code like it is in solidity. But use generic typing that I'll explore later a bit.

contracts/src/utils/structs/checkpoints.rs
```diff
@@ -0,0 +1,462 @@
...
+    }
+
+    /// Returns whether there is a checkpoint in the structure (i.g. it
+    /// is not empty), and if so, the key and value in the most recent
+    /// checkpoint.
+    ///
+    /// # Arguments
+    ///
+    /// * `&self` - read access to the checkpoint's state.
+    pub fn latest_checkpoint(&self) -> (bool, U96, U160) {
```
bidzyyys, line 170: Should it be `(bool, (U96, U160))`?

contracts/src/utils/structs/checkpoints.rs
```diff
@@ -0,0 +1,462 @@
...
+    /// * `value` - Checkpoint value corresponding to insertion `key`.
+    fn _unchecked_push(&mut self, key: U96, value: U160) {
+        let mut new_checkpoint = self._checkpoints.grow();
+        new_checkpoint._key.set(key);
+        new_checkpoint._value.set(value);
+    }
+}
+
+#[cfg(all(test, feature = "std"))]
+mod tests {
```
bidzyyys, line 346: I am not sure if we have enough test coverage here.

bidzyyys (commented):
Looks good overall, left some comments.

contracts/src/utils/structs/checkpoints.rs
```diff
@@ -0,0 +1,462 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `&mut self` - Write access to the checkpoint's state.
+    /// * `key` - Last checkpoint key to insert.
+    /// * `value` - Checkpoint value corresponding to insertion `key`.
+    ///
+    /// # Errors
+    ///
+    /// To maintain sorted order if the `key` is lower than
```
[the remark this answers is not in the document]
qalisander: Thnx! Changed it

contracts/src/utils/structs/checkpoints.rs
qalisander: Yep. Highlighted it in my previous comment. Fill fix it

contracts/src/utils/structs/checkpoints.rs
```diff
@@ -0,0 +1,462 @@
...
+    /// previously inserted error [`Error::CheckpointUnorderedInsertion`] is
+    /// returned.
+    pub fn push(
+        &mut self,
+        key: U96,
+        value: U160,
+    ) -> Result<(U160, U160), Error> {
```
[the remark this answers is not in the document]
qalisander: Currently it is matching to solidity.
Make sense to refactor it to type with fields in this case like:
```rust
struct PushResult{
  previous_value: U160,
  current_value: U160,
}
```
But it is tough to come up with good naming for now.. cc @alexfertel

contracts/src/utils/structs/checkpoints.rs
```diff
@@ -0,0 +1,462 @@
...
+//! new checkpoint for the current transaction block using the
+//! [`Trace160::push`] function.
+use alloy_primitives::{uint, Uint, U256, U32};
+use alloy_sol_types::sol;
+use stylus_proc::{sol_storage, SolidityError};
+use stylus_sdk::storage::{StorageGuard, StorageGuardMut};
+
+use crate::utils::math::alloy::Math;
+
+// TODO#q: add generics for other pairs (uint32, uint224) and (uint48, uint208).
```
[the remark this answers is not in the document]
qalisander: Could be. But it is a bit hard to work with code that is totally behind macro

contracts/src/utils/structs/checkpoints.rs
```diff
@@ -0,0 +1,462 @@
+//! Contract module for checkpointing values as they
+//! change at different points in time.
+//! Lets to look up past values by
+//! block number.
+//! See {Votes} as an example.
```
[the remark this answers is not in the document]
qalisander: thnx!

contracts/src/utils/structs/checkpoints.rs
```diff
@@ -0,0 +1,462 @@
...
+    }
+
+    /// Returns the value in the first (oldest) checkpoint with key
+    /// greater or equal than the search key, or zero if there is none.
+    ///
+    /// # Arguments
+    ///
+    /// * `&self` - read access to the checkpoint's state.
+    /// * `key` - Checkpoint's key to lookup.
+    pub fn lower_lookup(&self, key: U96) -> U160 {
```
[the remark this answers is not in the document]
qalisander: Do you mean `Key96` and `Value160` or some other names you'd like to offer?

contracts/src/utils/structs/checkpoints.rs
qalisander: Currently it is exactly like in solidity and I don't like it. I'll change it to `Option<(U96, U160)>`. That is a rust way

qalisander (commented):
Also make sense to stick out from solidity and rework api of the lookup functions to return not U*::ZERO when checkpoint wasn't found. But `Option::None`

contracts/src/utils/structs/checkpoints.rs
```diff
@@ -0,0 +1,463 @@
...
+        let len = self.length();
+        let pos = self._lower_binary_lookup(key, U256::ZERO, len);
+        if pos == len {
+            U160::ZERO
+        } else {
+            self._access(pos)._value.get()
+        }
```
alexfertel, line 93: Well, all of this could be replaced with [`core::slice::binary_search_by`](https://doc.rust-lang.org/beta/core/primitive.slice.html#method.binary_search_by), right?

I think we can simplify this implementation, but then again, we need to have ways to measure gas. Do you mind creating an issue for exploring this?

contracts/src/utils/structs/checkpoints.rs
```diff
@@ -0,0 +1,463 @@
...
+
+    /// Returns checkpoint at given position.
+    ///
+    /// # Arguments
+    ///
+    /// * `&self` - read access to the checkpoint's state.
+    /// * `pos` - index of the checkpoint.
+    pub fn at(&self, pos: U32) -> Checkpoint160 {
+        unsafe { self._checkpoints.get(pos).unwrap().into_raw() }
+    }
```
alexfertel, line 197: We should document that this can panic and we should document safety guarantees in a comment. If there is no way to guarantee safety, we should not use unsafe.

I think I don't understand the need for this if we have `_access`?

contracts/src/utils/structs/checkpoints.rs
```diff
@@ -0,0 +1,463 @@
...
+
+    /// Returns the number of checkpoints.
+    ///
+    /// # Arguments
+    ///
+    /// * `&self` - read access to the checkpoint's state.
+    pub fn length(&self) -> U256 {
+        U256::from(self._checkpoints.len())
+    }
```
alexfertel, line 187: I think this is a small optimization, but can't measure right now 😭 

```rust
U256::from_limbs([self.len(), 0, 0, 0])
```

Did not test!

alexfertel (changes requested):
Great job! This is looking sharp. I left a few questions and comments.

I think our biggest bottleneck for improving this is having a way to measure gas. I'll try to finish that soon.

contracts/src/utils/structs/checkpoints.rs
```diff
@@ -0,0 +1,463 @@
...
+use crate::utils::math::alloy::Math;
+
+// TODO: add generics for other pairs (uint32, uint224) and (uint48, uint208).
+// Logic should be the same.
+type U96 = Uint<96, 2>;
+type U160 = Uint<160, 3>;
+
+sol! {
+    /// A value was attempted to be inserted on a past checkpoint.
+    #[derive(Debug)]
```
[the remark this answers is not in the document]
qalisander: Good first issue https://github.com/OpenZeppelin/rust-contracts-stylus/issues/165

contracts/src/utils/structs/checkpoints.rs
qalisander: Yep. I also was thinking to remove it
p.s. in solidity it was as a public api. `access` and `at` at solidity have different types of arguments (indexes) u256 and u32. And access is basically `unsafe_access`. It uses assembly and works without bound checks. I can't see at stylus functions with indexing without bound checks are public and it always be checking bounds

contracts/src/utils/structs/checkpoints.rs
qalisander: I'm not sure that stylus sdk StorageVec is slice or has similar api. It seems for me we should have a binary search written by ourselves...

contracts/src/utils/structs/checkpoints.rs
```diff
@@ -0,0 +1,463 @@
...
+
+    /// Immutable access on an element of the checkpoint's array.
+    /// The position is assumed to be within bounds.
+    /// Panic when out of bounds.
+    ///
+    /// # Arguments
+    ///
+    /// * `&self` - Read access to the checkpoint's state.
+    /// * `pos` - index of the checkpoint.
+    fn _access(&self, pos: U256) -> StorageGuard<Checkpoint160> {
```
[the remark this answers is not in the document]
qalisander: for me also index seems better. Thnx!

contracts/src/utils/structs/checkpoints.rs
alexfertel: Ah, you're right, I forget we're not dealing with a `Vec` here.

contracts/src/utils/structs/checkpoints.rs
```diff
@@ -0,0 +1,463 @@
...
+    pub struct Checkpoint160 {
+        /// Key of checkpoint. Used as a sorting key.
+        uint96 _key;
+        /// Value corresponding to the key.
+        uint160 _value;
+    }
+}
+
+impl Trace160 {
+    /// Pushes a (`key`, `value`) pair into a Trace160 so that it is
```
bidzyyys, line 52: `Trace160`

contracts/src/utils/structs/checkpoints.rs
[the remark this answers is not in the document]
bidzyyys: Sth like `KeyU96` but leaving as in Solidity is okay for me.

contracts/src/utils/structs/checkpoints.rs
```diff
@@ -0,0 +1,463 @@
...
+            } else {
+                high = mid;
+            }
+        }
+        high
+    }
+
+    /// Immutable access on an element of the checkpoint's array.
+    /// The position is assumed to be within bounds.
+    /// Panic when out of bounds.
```
bidzyyys, line 306: I think we should have `Panics` section like here:
https://github.com/OpenZeppelin/rust-contracts-stylus/blob/main/contracts/src/token/erc20/mod.rs#L324-L326

contracts/src/utils/structs/checkpoints.rs
alexfertel: I'll leave it to your judgment!

contracts/src/utils/structs/checkpoints.rs
qalisander: what do you think should be the real size of `usize` in stylus sdk? It can be different kinda for different targets..

contracts/src/utils/structs/checkpoints.rs
qalisander: I've left it to return tuple of (key, value) same as for `latest_checkpoint` function

qalisander:
I've decided not to change now convention from solidity: "return U160::ZERO when checkpoint not found" since port of consecutive will benefit from the current way. (p.s using rust Optional type would be more reasonable)
Test code coverage is improved.

contracts/src/utils/structs/checkpoints.rs
```diff
@@ -0,0 +1,463 @@
...
+
+    /// Returns whether there is a checkpoint in the structure (i.g. it
+    /// is not empty), and if so, the key and value in the most recent
+    /// checkpoint.
+    /// Otherwise, [`None`] will be returned.
+    ///
+    /// # Arguments
+    ///
+    /// * `&self` - read access to the checkpoint's state.
+    pub fn latest_checkpoint(&self) -> Option<(U96, U160)> {
```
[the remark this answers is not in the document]
qalisander: Will check out later

contracts/src/utils/structs/checkpoints.rs
```diff
@@ -0,0 +1,517 @@
...
+            Some((checkpoint._key.get(), checkpoint._value.get()))
+        }
+    }
+
+    /// Returns the number of checkpoints.
+    ///
+    /// # Arguments
+    ///
+    /// * `&self` - read access to the checkpoint's state.
+    pub fn length(&self) -> U256 {
```
alexfertel, line 185: nit: Should we name this `len` to be consistent with Rust?

alexfertel (approved):
LGTM! 🚀 

I'll polish the comments a bit in a commit later today or tomorrow morn.

contracts/src/utils/structs/checkpoints.rs
qalisander: thnx! changed it

bidzyyys (approved):
Go go go 🚀

PR #163, opened 2024-06-27
fix: default storage macro with generically typed contracts

qalisander:
If once we will need to use contracts with generic params. DefaultStorageMacro will work with them from now on.


bidzyyys (approved):

alexfertel (approved):
Makes sense 👍

PR #164, opened 2024-06-27
docs(ci): remove minimal from workflow description

ggonzalez94:
This is super minor, but I noticed that on the description of this workflow we say that we have a job called `minimal` but we don't.


qalisander (approved):
Look good!

alexfertel (approved):
Great catch! 🚀

PR #168, opened 2024-07-01
feat: benchmark ERC-20 gas usage

### LINKED ISSUE #167: Gas benches for ERC-20

alexfertel:
Resolves #167 

Initial work towards having a full gas benchmarking suite.

It's tested running against `nitro-testnode`, but should be good to test against Arbitrum Sepolia or any other Stylus-enabled RPC endpoint.

The output looks like this:

Screenshot 2024-07-01 at 14 55 13

You can see that the `L1 Gas` is 0, and this is because of the configuration of `nitro-testnode`, for a testnet like Arbitrum Sepolia this number would be different than zero most of the time.

Note that we need to deserialize the extra fields of the transaction receipt in order to access the gas usage in the L1.

```rust
#[derive(Debug, Deserialize)]
struct ArbOtherFields {
    #[serde(rename = "gasUsedForL1")]
    gas_used_for_l1: U128,
    #[allow(dead_code)]
    #[serde(rename = "l1BlockNumber")]
    l1_block_number: String,
}
```

### DIFF
--- a/benches/src/erc20.rs
+++ b/benches/src/erc20.rs
@@ -73,10 +73,7 @@ pub async fn bench() -> eyre::Result<()> {
     let contract_addr = deploy(&alice).await;
     let contract = Erc20::new(contract_addr, &alice_wallet);
 
-    println!("+------------------------------------------------------+");
-    println!("| Function Name      | L2 Gas | L1 Gas | Effective Gas |");
-    println!("|--------------------|--------|--------|---------------|");
-
+    println!("Running benches...");
     let receipts = vec![
         ("name()", receipt!(contract.name())?),
         ("symbol()", receipt!(contract.symbol())?),
@@ -86,11 +83,11 @@ pub async fn bench() -> eyre::Result<()> {
         ("balanceOf(account)", receipt!(contract.balanceOf(alice_addr))?),
         (
             "mint(account, amount)",
-            receipt!(contract.mint(alice_addr, uint!(100_U256)))?,
+            receipt!(contract.mint(alice_addr, uint!(10_U256)))?,
         ),
         (
             "burn(amount)",
-            receipt!(contract.burn(uint!(100_U256)).from(alice_addr))?,
+            receipt!(contract.burn(uint!(1_U256)).from(alice_addr))?,
         ),
         (
             "transfer(account, amount)",
@@ -99,6 +96,10 @@ pub async fn bench() -> eyre::Result<()> {
                 .from(alice_addr))?,
         ),
     ];
+
+    println!("+------------------------------------------------------+");
+    println!("| Function Name      | L2 Gas | L1 Gas | Effective Gas |");
+    println!("|--------------------|--------|--------|---------------|");
 
     for (func_name, receipt) in receipts {
         let l2_gas = receipt.gas_used;
--- a/benches/src/erc20.rs
+++ b/benches/src/erc20.rs
@@ -1,5 +1,5 @@
 use alloy::{
-    network::AnyNetwork,
+    network::{AnyNetwork, EthereumWallet},
     primitives::Address,
     providers::{fillers::ChainIdFiller, ProviderBuilder},
     rpc::types::TransactionReceipt,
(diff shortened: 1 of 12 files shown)


benches/src/erc20.rs
```diff
@@ -0,0 +1,194 @@
...
+};
+use alloy_primitives::U256;
+use e2e::{receipt, Account};
+use koba::config::Deploy;
+
+use crate::ArbOtherFields;
+
+const RPC_URL: &str = "http://localhost:8547";
+
+sol!(
```
bidzyyys, line 17: Is there any way to reuse `abi.rs` from integration tests here?

benches/src/erc20.rs
```diff
@@ -0,0 +1,194 @@
...
+pub async fn bench() -> eyre::Result<()> {
+    let alice = Account::new().await?;
+    let alice_addr = alice.address();
+    let alice_wallet = ProviderBuilder::new()
+        .network::<AnyNetwork>()
+        .with_recommended_fillers()
+        .wallet(EthereumWallet::from(alice.signer.clone()))
+        .on_http(alice.url().parse()?);
+
+    let bob = Account::new().await?;
```
bidzyyys, line 54: How about using `AccountFactory` from E2E tests?

bidzyyys (approved):
Great job @alexfertel! 🚀 
Left some minor comments.

lib/e2e/src/system.rs
```diff
@@ -56,3 +55,25 @@ pub fn provider() -> Provider {
         .expect("failed to parse RPC_URL string into a URL");
     ProviderBuilder::new().with_recommended_fillers().on_http(rpc_url)
 }
+
+/// Send `amount` eth to `address` in the nitro-tesnode.
+pub fn fund_account(address: Address, amount: &str) -> eyre::Result<()> {
```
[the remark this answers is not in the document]
alexfertel: Where should it live?

benches/src/account.rs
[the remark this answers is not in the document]
alexfertel: Oh, good catch!

lib/e2e/src/system.rs
```diff
@@ -44,10 +47,6 @@ fn env(name: &str) -> eyre::Result<String> {
 }
 
 /// Returns an alloy provider connected to the `RPC_URL` rpc endpoint.
-///
-/// # Panics
```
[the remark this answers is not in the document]
alexfertel: Ah, I made a big refactor that I scratched and forgot to readd this comment, great catch!

benches/src/erc20.rs
alexfertel: This would couple benches with e2e tests, so it's better not to do this.

benches/src/erc20.rs
alexfertel: Hmmm, what do you mean? This is using the `AccountFactory` under the hood.

lib/e2e/src/system.rs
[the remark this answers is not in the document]
bidzyyys: I would put it in [`lib/e2e/src/account.rs`](https://github.com/OpenZeppelin/rust-contracts-stylus/blob/main/lib/e2e/src/account.rs) as it introduces logic connected with accounts. But I can agree on leaving it here as it is.

benches/src/erc20.rs
bidzyyys: Fair enough.

benches/src/erc20.rs
bidzyyys: Oh, you are right, never mind.

bidzyyys (approved):

lib/e2e/src/system.rs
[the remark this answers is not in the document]
alexfertel: I chose to move it to this module, since it is completely coupled to `nitro-testnode` (which is our `system`) and funding an account is kind of like a god-like operation, but this is not a strong opinion.

lib/e2e/src/account.rs
```diff
@@ -78,18 +75,7 @@ impl AccountFactory {
...
-        let node_script = get_node_path()?.join("test-node.bash");
-        let output = std::process::Command::new(node_script)
-            .arg("script")
-            .arg("send-l2")
-            .arg("--to")
-            .arg(format!("address_{addr}"))
-            .arg("--ethamount")
-            .arg("10")
-            .output()?;
+        fund_account(addr, "10")?;
```
qalisander, line 78: cool!

qalisander (approved):
Great job!
Coll that we have gas benchmarks

PR #177, opened 2024-07-02
feat(e2e): enable checking whether an event was not emitted

alexfertel:
Small refactor to enable checking an event does *not* get emitted.

I'm currently working on the e2e tests for `AccessControl` and had the need to check that an event was not emitted. However, with the current API of `receipt.emits`, the assertion occurs inside the `emits` function, which does not let me do something like `assert!(!receipt.emits(...));`.

This PR enables this by delegating the assertion to the caller.

This also makes the API more consistent with the Error e2e API.


qalisander (approved):
Good

bidzyyys (approved):
LGTM! 🚀

PR #179, opened 2024-07-04
feat: bump stylus-sdk-rs version to v0.5.2

alexfertel:


contracts/src/token/erc721/mod.rs
```diff
@@ -1092,7 +1088,7 @@ impl Erc721 {
                 operator,
                 from,
                 token_id,
-                data.to_vec(),
+                data.to_vec().into(),
```
bidzyyys, line 1091: Curious if we can use just `data.into()`.

bidzyyys (approved):
LGTM! 🚀

contracts/src/token/erc721/mod.rs
alexfertel: We can't sadly, there's some weirdness here in the new types it seems :(

alexfertel:
Leaving as a draft until we hear back from the Offchain Labs team.

bidzyyys:
There is a new version, `v0.5.2` maybe worth trying?

alexfertel:
@bidzyyys still broken for the same reason :(

Issue #170, opened 2024-07-01
Refactor benchmarking infra


alexfertel:
Currently, our benchmarking suite is quite rough and has a bunch of code dependent on `e2e`, we should decouple them.

Benches are also coupled to displaying the results, which should not be the case. Once this coupling is gone, it should make adding the results as a comment on a PR much easier, as mentioned in #169

PR #196, opened 2024-07-15
feat: bench merkle proofs verification gas usage

### LINKED ISSUE #174: Gas benches for Merkle Proof Verification
Add benches based on the work in #168

alexfertel:
Resolves #174 

Screenshot 2024-07-15 at 16 45 37

### DIFF
--- a/benches/src/merkle_proofs.rs
+++ b/benches/src/merkle_proofs.rs
@@ -1,7 +1,7 @@
 use alloy::{
     hex,
     network::{AnyNetwork, EthereumWallet},
-    primitives::{address, Address},
+    primitives::Address,
     providers::ProviderBuilder,
     sol,
 };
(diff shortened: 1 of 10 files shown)


benches/src/merkle_proofs.rs
```diff
@@ -0,0 +1,128 @@
...
+    .map(|h| h.into())
+    .to_vec();
+
+    let receipts = vec![(
+        "verify()",
+        receipt!(contract.verify(proof, root.into(), leaf.into()))?,
+    )];
+
+    // Calculate the width of the longest function name.
+    let max_name_width = receipts
```
bidzyyys, line 80: Nit: each benchmark has the same way for calculating `max_name_width`. How about common code?

bidzyyys: And printing the table if applicable.

bidzyyys (approved):
I like it 🚀

benches/src/merkle_proofs.rs
qalisander: Agree. Would be more simple to add another benchmarks. At least considering performance while designing new contract

qalisander (approved):
Good job!

benches/src/merkle_proofs.rs
alexfertel: Yes! I'm leaving that for a separate PR, cause benchmarks are a bigger priority. I opened an issue for it some time ago https://github.com/OpenZeppelin/rust-contracts-stylus/issues/170

Issue #43, opened 2024-04-10
ERC-20 Permit


bidzyyys:
```[tasklist]
### Tasks
- [ ] https://github.com/OpenZeppelin/rust-contracts-stylus/issues/182
- [ ] https://github.com/OpenZeppelin/rust-contracts-stylus/issues/183
- [ ] https://github.com/OpenZeppelin/rust-contracts-stylus/issues/184
- [ ] https://github.com/OpenZeppelin/rust-contracts-stylus/issues/17
```

developeruche:
This entails implementing logic similar to https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v4.9.5/contracts/token/ERC20/extensions/ERC20Permit.sol right? 

but implementing this logic; 
```solidity 
address signer = ECDSA.recover(hash, v, r, s);
```
is not yet clear; 

Wasn't able to see how to access precomplies in stylus, would it be cool to use a lib for this that works fine in a `no_std` env?

alexfertel:
> This entails implementing logic similar to [OpenZeppelin/openzeppelin-contracts@v4.9.5/contracts/token/ERC20/extensions/ERC20Permit.sol](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v4.9.5/contracts/token/ERC20/extensions/ERC20Permit.sol?rgh-link-date=2024-06-22T13%3A38%3A21Z) right?

Yes, that's right! We already have an issue that tracks implementing ECDSA (#17). However, the works there needs a bit of design first.

> Wasn't able to see how to access precomplies in stylus

You can just make calls to the precompile address! We'll have an example of this in the implementation of ECDSA.

---

Just a quick heads-up about tackling tasks. Before you dive into something, it'd be great if you could give us a shout. That way, we can assign the issue to you and avoid any accidental double-ups. It's all about keeping things smooth and making sure we're not stepping on each other's toes. So, if you've got your eye on a particular task, let us know, and we'll get you set up. Thanks for helping us keep things organized!

developeruche:
> > This entails implementing logic similar to [OpenZeppelin/openzeppelin-contracts@v4.9.5/contracts/token/ERC20/extensions/ERC20Permit.sol](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v4.9.5/contracts/token/ERC20/extensions/ERC20Permit.sol?rgh-link-date=2024-06-22T13%3A38%3A21Z) right?
> 
[...]

PR #197, opened 2024-07-16
feat: EIP712 implementation

### LINKED ISSUE #184: EIP712
Implement `EIP712` contract. Solidity reference [here](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/EIP712.sol).

developeruche:
Implement EIP712 contract. Solidity reference [here](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/EIP712.sol).

Resolves #184

#### PR Checklist

- [x] Tests
- [x] Documentation


developeruche:
Please I need help on; 

1. This is giving me this error:
```
the trait bound `alloy_primitives::Bytes: AbiType` is not satisfied
the following other types implement trait `AbiType`:
  ()
  (A, B, C, D, E, F, G, H, I, J, K, L, M, N, O, P, Q, R, S, T, U, V, W, X)
  (B, C, D, E, F, G, H, I, J, K, L, M, N, O, P, Q, R, S, T, U, V, W, X)
  (C, D, E, F, G, H, I, J, K, L, M, N, O, P, Q, R, S, T, U, V, W, X)
  (D, E, F, G, H, I, J, K, L, M, N, O, P, Q, R, S, T, U, V, W, X)
  (E, F, G, H, I, J, K, L, M, N, O, P, Q, R, S, T, U, V, W, X)
  (F, G, H, I, J, K, L, M, N, O, P, Q, R, S, T, U, V, W, X)
  (G, H, I, J, K, L, M, N, O, P, Q, R, S, T, U, V, W, X)
and 37 others
required for `(Bytes, String, String, u64, Address, FixedBytes<32>, Vec<Uint<256, 4>>)` to implement `AbiType`
required for `(Bytes, String, String, u64, Address, FixedBytes<32>, Vec<Uint<256, 4>>)` to implement `EncodableReturnType`
the full name for the type has been written to '/Users/user/Documents/projects/CONTRIBUTION/rust-contracts-stylus/target/debug/deps/openzeppelin_stylus-03b9c0872ea01cad.long-type-9805022037568242759.txt'
consider using `--verbose` to print the full type name to the console
the full name for the type has been written to '/Users/user/Documents/projects/CONTRIBUTION/rust-contracts-stylus/target/debug/deps/openzeppelin_stylus-03b9c0872ea01cad.long-type-9805022037568242759.txt'
consider using `--verbose` to print the full type name to the console
```

code:
```rust 
pub fn eip712_domain(
        &self,
    ) -> (Bytes, String, String, u64, Address, B256, Vec<U256>) {
        (
            FEILDS,
            self.eip712_name(),
            self.eip712_version(),
            block::chainid(),
            contract::address(),
            SALT,
            Vec::new(),
        )
    }
```

developeruche:
```rust 
abi.encode(data)
```

Please how can I do this using `alloy`

developeruche:
@bidzyyys these are my blockers before writing the Unit test

bidzyyys:
Hi @developeruche I think I fixed your issue;
Following [Solidity reference](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/EIP712.sol#L119), `fields` is `bytes1`, and in Rust it should be `FixedBytes<1>`.

developeruche:
> Hi @developeruche I think I fixed your issue; Following [Solidity reference](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/EIP712.sol#L119), `fields` is `bytes1`, and in Rust it should be `FixedBytes<1>`.

Thanks!... 
Got you, would stick to the reference...

bidzyyys:
> ```rust
> abi.encode(data)
> ```
> 
> Please how can I do this using `alloy`

Based on the Solidity interface (from `sol!` macro) you need to use alloy as e.g. here:
https://github.com/OpenZeppelin/rust-contracts-stylus/blob/main/examples/erc721-metadata/tests/erc721.rs#L33

developeruche:
> > ```rust
> > abi.encode(data)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Please how can I do this using `alloy`
> 
> Based on the Solidity interface (from `sol!` macro) you need to use alloy as e.g. here: https://github.com/OpenZeppelin/rust-contracts-stylus/blob/main/examples/erc721-metadata/tests/erc721.rs#L33

We cannot go this Route as this type is not an Alloy type, e.g. `sol_types::SolConstructor` from the `sol!` marco.
The Ideal way to go about this is to use the `.abi_ecode()` fn provided by `sol_types::SolValue`, but that is not avaliable in `alloy_sol_type` version `0.3.1` which is curently been used across the workspace.

Got this info from Alloy telegram chat.

Please do you have an alternative?

bidzyyys:
> > > > > ```rust
> > > > > abi.encode(data)
> > > > > ```
> > > > > 
> > > > > 
> > > > >     
> > > > >       
> > > > >     
> > > > > 
> > > > >       
> > > > >     
> > > > > 
> > > > >     
> > > > >   
> > > > > Please how can I do this using `alloy`
> > > > 
> > > > 
> > > > Based on the Solidity interface (from `sol!` macro) you need to use alloy as e.g. here: https://github.com/OpenZeppelin/rust-contracts-stylus/blob/main/examples/erc721-metadata/tests/erc721.rs#L33
> > > 
> > > 
> > > We cannot go this Route as this type is not an Alloy type, e.g. `sol_types::SolConstructor` from the `sol!` marco. The Ideal way to go about this is to use the `.abi_ecode()` fn provided by `sol_types::SolValue`, but that is not avaliable in `alloy_sol_type` version `0.3.1` which is curently been used across the workspace.
> > > Got this info from Alloy telegram chat.
> > > Please do you have an alternative?
> > 
> > 
> > Please share what do you want to achieve, will be easier for me to figure it out.
> 
> This is what I want to achieve;
> 
> ```solidity
> keccak256(abi.encode(TYPE_HASH, _hashedName, _hashedVersion, block.chainid, address(this)));
> ```

@developeruche please prepare the code that does not compile and push it to the repo (may be commented out), I will have a look in a few hours.

developeruche:
> > > > > > ```rust
> > > > > > abi.encode(data)
> > > > > > ```
> > > > > > 
> > > > > > 
> > > > > >     
> > > > > >       
> > > > > >     
> > > > > > 
> > > > > >       
> > > > > >     
> > > > > > 
> > > > > >     
> > > > > >   
> > > > > > Please how can I do this using `alloy`
> > > > > 
> > > > > 
> > > > > Based on the Solidity interface (from `sol!` macro) you need to use alloy as e.g. here: https://github.com/OpenZeppelin/rust-contracts-stylus/blob/main/examples/erc721-metadata/tests/erc721.rs#L33
> > > > 
> > > > 
> > > > We cannot go this Route as this type is not an Alloy type, e.g. `sol_types::SolConstructor` from the `sol!` marco. The Ideal way to go about this is to use the `.abi_ecode()` fn provided by `sol_types::SolValue`, but that is not avaliable in `alloy_sol_type` version `0.3.1` which is curently been used across the workspace.
> > > > Got this info from Alloy telegram chat.
> > > > Please do you have an alternative?
> > > 
> > > 
> > > Please share what do you want to achieve, will be easier for me to figure it out.
> > 
> > 
> > This is what I want to achieve;
> > ```solidity
> > keccak256(abi.encode(TYPE_HASH, _hashedName, _hashedVersion, block.chainid, address(this)));
> > ```
> 
> @developeruche please prepare the code that does not compile and push it to the repo (may be commented out), I will have a look in a few hours.

thanks... just pushed

bidzyyys:
@developeruche I think I figured it out, but you need to write tests to confirm. 
This source code was helpful for me -> https://docs.rs/alloy-sol-types/0.3.1/src/alloy_sol_types/coder/encoder.rs.html

developeruche:
> @developeruche I think I figured it out, but you need to write tests to confirm. This source code was helpful for me -> https://docs.rs/alloy-sol-types/0.3.1/src/alloy_sol_types/coder/encoder.rs.html

thanks a lot... would get on this now

developeruche:
@bidzyyys please I need your technical advice on how to carry out this test without using a constructor. The contract does not have a function `internal` or `public` that updates the state variables leaving this to be done only by using a `constructor` which is not supported ATM.

bidzyyys:
> @bidzyyys please I need your technical advice on how to carry out this test without using a constructor. The contract does not have a function `internal` or `public` that updates the state variables leaving this to be done only by using a `constructor` which is not supported ATM.

Hi @developeruche!

- If you need to set a state in unit test (`motsu`), you can directly write data to storage like here -> https://github.com/OpenZeppelin/rust-contracts-stylus/blob/main/contracts/src/token/erc20/extensions/capped.rs#L61-L70
- If you need to set a state in E2E tests, you can check the code here -> https://github.com/OpenZeppelin/rust-contracts-stylus/blob/main/examples/erc20/tests/erc20.rs#L26-L30

developeruche:
There is a little problem, any function that has `block::chainid()` therein fails with this error;
```
dyld[7502]: missing symbol called
error: test failed, to rerun pass `-p openzeppelin-stylus --lib`
```

Other tests passed fine. I have gone through the codebase, but didn't see where `block::chainid()` was used so there is no sample test I can play with.

bidzyyys:
> There is a little problem, any function that has `block::chainid()` therein fails with this error;
> 
> ```
> dyld[7502]: missing symbol called
> error: test failed, to rerun pass `-p openzeppelin-stylus --lib`
> ```
> 
> Other tests passed fine. I have gone through the codebase, but didn't see where `block::chainid()` was used so there is no sample test I can play with.
> 
> cc: @bidzyyys

Hi @developeruche,
When you see this error, you should implement `vm-hooks`. In this case you are probably missing `chainid`, see https://github.com/OffchainLabs/stylus-sdk-rs/blob/stylus/stylus-sdk/src/hostio.rs#L127-L131.

Here you should add the hook, and check out what we have and how we deal with this:
https://github.com/OpenZeppelin/rust-contracts-stylus/blob/main/lib/motsu/src/shims.rs

In your case, I would hardcode `chainid` in shims as we did for e.g. `msg_sender` -> https://github.com/OpenZeppelin/rust-contracts-stylus/blob/main/lib/motsu/src/shims.rs#L179-L183

developeruche:
> > There is a little problem, any function that has `block::chainid()` therein fails with this error;
> > ```
> > dyld[7502]: missing symbol called
> > error: test failed, to rerun pass `-p openzeppelin-stylus --lib`
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Other tests passed fine. I have gone through the codebase, but didn't see where `block::chainid()` was used so there is no sample test I can play with.
> > cc: @bidzyyys
> 
> Hi @developeruche, When you see this error, you should implement `vm-hooks`. In this case you are probably missing `chainid`, check this https://github.com/OffchainLabs/stylus-sdk-rs/blob/stylus/stylus-sdk/src/hostio.rs#L127-L131.
> 
> Here you should add the hook, and check out what we have and how we deal with the hooks: https://github.com/OpenZeppelin/rust-contracts-stylus/blob/main/lib/motsu/src/shims.rs
> 
> In your case, I would hardcode `chainid` in shims as we did for e.g. `msg_sender` -> https://github.com/OpenZeppelin/rust-contracts-stylus/blob/main/lib/motsu/src/shims.rs#L179-L183

it is still abit confusing how everything plugs in together; 
I added this piece of code frollowing the example; 


```rust
/// Gets the chain ID of the current chain. The semantics are equivalent to
/// that of the EVM's [`CHAINID`] opcode.
/// 
/// [`CHAINID`]: https://www.evm.codes/#46
/// 
/// # Panics
/// Not Enough Gas
/// Stack Overflow
#[no_mangle]
pub unsafe extern "C" fn chain_id(sender: *mut u8) {
    let addr = const_hex::const_decode_to_array::<6>(CHAIN_ID).unwrap();
    std::ptr::copy(addr.as_ptr(), sender, 6);
}

```

but I am still getting the same error 😢

bidzyyys:
> > > There is a little problem, any function that has `block::chainid()` therein fails with this error;
> > > ```
> > > dyld[7502]: missing symbol called
> > > error: test failed, to rerun pass `-p openzeppelin-stylus --lib`
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Other tests passed fine. I have gone through the codebase, but didn't see where `block::chainid()` was used so there is no sample test I can play with.
> > > cc: @bidzyyys
> > 
> > 
> > Hi @developeruche, When you see this error, you should implement `vm-hooks`. In this case you are probably missing `chainid`, check this https://github.com/OffchainLabs/stylus-sdk-rs/blob/stylus/stylus-sdk/src/hostio.rs#L127-L131.
> > Here you should add the hook, and check out what we have and how we deal with the hooks: https://github.com/OpenZeppelin/rust-contracts-stylus/blob/main/lib/motsu/src/shims.rs
> > In your case, I would hardcode `chainid` in shims as we did for e.g. `msg_sender` -> https://github.com/OpenZeppelin/rust-contracts-stylus/blob/main/lib/motsu/src/shims.rs#L179-L183
> 
> it is still abit confusing how everything plugs in together; I added this piece of code frollowing the example;
> 
> ```rust
> /// Gets the chain ID of the current chain. The semantics are equivalent to
> /// that of the EVM's [`CHAINID`] opcode.
> /// 
> /// [`CHAINID`]: https://www.evm.codes/#46
> /// 
> /// # Panics
> /// Not Enough Gas
> /// Stack Overflow
> #[no_mangle]
> pub unsafe extern "C" fn chain_id(sender: *mut u8) {
>     let addr = const_hex::const_decode_to_array::<6>(CHAIN_ID).unwrap();
>     std::ptr::copy(addr.as_ptr(), sender, 6);
> }
> ```
> 
> but I am still getting the same error 😢

Hi @developeruche,
I have found and fixed to issues;
1. You did not implement `chainid` in a right way.
  - Docs says it should be like this -> https://github.com/OffchainLabs/stylus-sdk-rs/blob/stylus/stylus-sdk/src/hostio.rs#L131
  -  This is my fix -> https://github.com/OpenZeppelin/rust-contracts-stylus/pull/197/commits/78a1f811247972e23ffe46fda64632034f94ab08
2.  You were missing `contract_address` shim:
  - Required e.g. here -> https://github.com/developeruche/rust-contracts-stylus/blob/feat/eip712/contracts/src/cryptography/eip712.rs#L228
  - Docs here -> https://github.com/OffchainLabs/stylus-sdk-rs/blob/stylus/stylus-sdk/src/hostio.rs#L184-L188
  - My fix -> https://github.com/OpenZeppelin/rust-contracts-stylus/pull/197/commits/0e2664073b81ec32f8d7394099e7e4453c1f155a

bidzyyys:
@developeruche quick summary,
If you see the below error, it means that you are missing one of the shims from here -> https://github.com/OffchainLabs/stylus-sdk-rs/blob/stylus/stylus-sdk/src/hostio.rs
```
error: test failed, to rerun pass `-p openzeppelin-stylus --lib`

Caused by:
  process didn't exit successfully: ... (signal: 6, SIGABRT: process abort signal)
```

developeruche:
Thanks a lot... this is all clear now!

contracts/src/cryptography/eip712.rs
```diff
@@ -0,0 +1,265 @@
...
+        string _name;
+        /// The version
+        string _version;
+    }
+}
+
+#[external]
+impl EIP712 {
+    /// Returns the fields and values that describe the domain separator used by
+    /// this contract for EIP-712 signature.
```
bidzyyys, line 80: Missing `#Arguments` section like:
```
 /// # Arguments
 ///
```

bidzyyys: Applies to all functions when missing.

contracts/src/cryptography/message_hash_utils.rs
```diff
@@ -0,0 +1,87 @@
...
+    let encoded = TypeHashCoder::encode_packed(&(
+        String::from("\x19\x01"),
+        *domain_separator,
+        *struct_hash,
+    ));
+    keccak256(encoded)
+}
+
+#[cfg(all(test, feature = "std"))]
+mod tests {
```
bidzyyys, line 56: Do the values from tests come from verified Solidity contracts @developeruche ?

bidzyyys (commented):
Looks good, great job @developeruche! 🚀

bidzyyys:
@alexfertel @qalisander I am curious if we should add E2E tests for this contract.

contracts/src/cryptography/eip712.rs
```diff
@@ -0,0 +1,246 @@
...
+        /// The hashed name.
+        bytes32 _hashed_name;
+        /// The hashed version.
+        bytes32 _hashed_version;
+        /// The name.
+        string _name;
+        /// The version.
+        string _version;
+    }
+}
```
alexfertel, line 55: In the Solidity implementation, you can see that most state variables are in reality private immutables, set during contract construction:

```solidity
bytes32 private immutable _cachedDomainSeparator;
uint256 private immutable _cachedChainId;
address private immutable _cachedThis;

bytes32 private immutable _hashedName;
bytes32 private immutable _hashedVersion;

ShortString private immutable _name;
ShortString private immutable _version;
string private _nameFallback;
string private _versionFallback;

constructor(string memory name, string memory version) {
    _name = name.toShortStringWithFallback(_nameFallback);
    _version = version.toShortStringWithFallback(_versionFallback);
    _hashedName = keccak256(bytes(name));
    _hashedVersion = keccak256(bytes(version));

    _cachedChainId = block.chainid;
    _cachedDomainSeparator = _buildDomainSeparator();
    _cachedThis = address(this);
}
```

In reality, the only storage used is `_nameFallback` and `_versionFallback`. I think we can probably remove *all* that state:

- private immutables are just search and replace in the compiled binary. They are equivalent to Rust's `const` variables.
- `_nameFallback` and `_versionFallback` are only used in case the consumer doesn't properly set `name` and `version`, but if they are set at compile time, then we don't need them at all.

It turns out this EIP implementation doesn't even have to be a contract, since it doesn't need storage -- we can resolve these at compile time by making the user specify the variables. Something like the [ERC-20 example](https://github.com/OffchainLabs/stylus-sdk-rs/blob/17fc86a4615af0769e8f528f32276e067702d52e/examples/erc20/src/lib.rs#L21-L24) of the SDK, but not even requiring it to be a contract, just a plain ol' Rust generic trait.

Let's do this and move all of this code to `lib/crypto` please.

contracts/src/cryptography/eip712.rs
```diff
@@ -0,0 +1,246 @@
...
+};
+
+use super::message_hash_utils::to_typed_data_hash;
+
+/// `keccak256("EIP712Domain(string name,string version,uint256 chainId,address
+/// verifyingContract)");`
+pub const TYPE_HASH: B256 =
+    b256!("8b73c3c69bb8fe3d512ecc4cf759cc79239f7b179b0ffacaa9a75d522b39400f");
+pub const FIELDS: FixedBytes<1> = fixed_bytes!("15");
+pub const SALT: B256 = B256::ZERO;
```
alexfertel, line 31: By default we should try to use `alloy`'s types as least as possible to improve consistency and compatibility. This would also improve this code -- a tell sign is that you are having to dereference these variables when they are used.

```suggestion
pub const TYPE_HASH: [u8; 32] =
    hex!("8b73c3c69bb8fe3d512ecc4cf759cc79239f7b179b0ffacaa9a75d522b39400f");
pub const FIELDS: [u8; 2] = hex!("0f");
pub const SALT: [u8; 32] = [0u8; 32];
```

contracts/src/cryptography/eip712.rs
```diff
@@ -0,0 +1,246 @@
...
+        let version = "1";
+        contract._version.set_str(version);
+
+        assert_eq!(contract.eip712_version(), version);
+    }
+
+    #[motsu::test]
+    fn test_eip712_domain(contract: Eip712) {
+        let name = "EIP712";
+        let version = "1";
```
alexfertel, line 147: Let's extract these as `const`s.

contracts/src/cryptography/message_hash_utils.rs
```diff
@@ -0,0 +1,93 @@
...
+///
+/// NOTE: The `message_hash` parameter is intended to be the result of hashing a
+/// raw message with keccak256, although any bytes32 value can be safely used
+/// because the final digest will be re-hashed.
+///
+/// [eth_sign]: https://eth.wiki/json-rpc/API#eth_sign
+pub fn to_eth_signed_message_hash(message_hash: B256) -> B256 {
+    // "\x19Ethereum Signed Message:\n32" in bytes
+    let prefix =
+        bytes!("19457468657265756d205369676e6564204d6573736167653a0a3332");
```
alexfertel, line 38: Lets extract this to a `const`.

contracts/src/cryptography/message_hash_utils.rs
```diff
@@ -0,0 +1,93 @@
...
+/// The digest is calculated by prefixing a bytes32 `message_hash` with
+/// `"\x19Ethereum Signed Message:\n32"` and hashing the result. It corresponds
+/// with the hash signed when using the [eth_sign] JSON-RPC method.
+///
+/// NOTE: The `message_hash` parameter is intended to be the result of hashing a
+/// raw message with keccak256, although any bytes32 value can be safely used
+/// because the final digest will be re-hashed.
+///
+/// [eth_sign]: https://eth.wiki/json-rpc/API#eth_sign
+pub fn to_eth_signed_message_hash(message_hash: B256) -> B256 {
```
alexfertel, line 35: This is not doing exactly what the comment says. Let's try to replicate [this code instead](https://github.com/alloy-rs/core/blob/adcf7adfa1f35c56e6331bab85b8c56d32a465f1/crates/primitives/src/utils/mod.rs#L111-L143).

alexfertel: Note that we append `32` because that is the message length, but we should have the more general `eip191_message` and then use it as it should be used in this particular case for messages of length 32.

contracts/src/cryptography/message_hash_utils.rs
```diff
@@ -0,0 +1,93 @@
...
+use alloy_primitives::{bytes, keccak256, B256};
+use alloy_sol_types::{sol, SolType};
+
+type EthMessageCoder = sol! {
+    tuple(bytes, bytes32)
+};
+
+type TypeHashCoder = sol! {
+    tuple(string, bytes32, bytes32)
+};
```
alexfertel, line 21: I think it's better to co-locate this next to where they are used instead of at the top of the file. They should have docs comments as well.

contracts/src/cryptography/message_hash_utils.rs
```diff
@@ -0,0 +1,93 @@
...
+        );
+        let struct_hash = b256!(
+            "7379746c75730000000000000000000000000000000000000000000000000000"
+        );
+        let expected = b256!(
+            "cefc47137f8165d8270433dd62e395f5672966b83a113a7bb7b2805730a2197e"
+        );
+
+        assert_eq!(to_typed_data_hash(domain_separator, struct_hash), expected);
+    }
```
alexfertel, line 92: It's good practice to add a bit of explanation when values seem random so that the person reading the code understands assumptions and expectations. Please document where did these values come from.

lib/motsu/src/shims.rs
```diff
@@ -182,6 +189,29 @@ pub unsafe extern "C" fn msg_sender(sender: *mut u8) {
     std::ptr::copy(addr.as_ptr(), sender, 20);
 }
 
+/// Gets the address of the current program. The semantics are equivalent to
+/// that of the EVM's [`ADDRESS`] opcode.
+///
+/// [`ADDRESS`]: https://www.evm.codes/#30
```
alexfertel, line 195:
```suggestion
/// [`ADDRESS`]: https://www.evm.codes/#30
///
```

alexfertel (changes requested):
Overall looking good!

It turns out that in Rust, most of this behaviour can be implemented as a library, separate from the actual contract, so please, lets move the implementation to `lib/crypto` in a way that is easy to use from a contract.

The motivation for this contract to exist is to be used in conjunction with ECDSA to implement Permit. Thankfully, we can solve most of the state at compile-time, so we don't need constructors, or storage.

It should also try to use `alloy` as little as possible, so that it is compatible with both `alloy` & non-`alloy` Rust code as frictionless as possible.

Note how EIP-712 is only used as a parent contract, which means it is never really deployed on its own, so I don't think it makes sense to have it as a contract in our library.

contracts/src/cryptography/message_hash_utils.rs
developeruche: yeah, they are all values gotten from a solidity contract, would add some comments now

contracts/src/cryptography/message_hash_utils.rs
developeruche: sorry the comment was wrong.

```solidity
    function toEthSignedMessageHash(bytes32 messageHash) internal pure returns (bytes32 digest) {
        /// @solidity memory-safe-assembly
        assembly {
            mstore(0x00, "\x19Ethereum Signed Message:\n32") // 32 is the bytes-length of messageHash
            mstore(0x1c, messageHash) // 0x1c (28) is the length of the prefix
            digest := keccak256(0x00, 0x3c) // 0x3c is the length of the prefix (0x1c) + messageHash (0x20)
        }
    }
```

this is the solidity code I am trying to emulate;

developeruche: looking at this; 

https://github.com/developeruche/rust-contracts-stylus/blob/8d0954f9eb7de2a4056794b19029e725dcf61a8d/lib/crypto/src/message_hash_utils.rs#L31

should I still proceed?

contracts/src/cryptography/message_hash_utils.rs
developeruche: It came from a solidity smart contract but it is not verified...

https://gist.github.com/developeruche/3de82874ea23fe624c666ad20117dea3

contracts/src/cryptography/eip712.rs
developeruche: please correct me if I am wrong; 


This means we would rather introduce a `EIP712` trait, having the `eip712_domain` method;

```rust 
pub fn eip712_domain(
        &self,
    ) -> (FixedBytes<1>, String, String, u64, Address, B256, Vec<U256>);
pub fn eip712_name(&self) -> String;
pub fn eip712_version(&self) -> String;
```

and the rest of the methods would be extracted to be standalone; 
- `build_domain_separator`
- `domain_separator_v4`
- `hash_typed_data_v4`

lastly, the `NAME` and `VERSION` would be associated const types.

cc: @alexfertel

contracts/src/cryptography/message_hash_utils.rs
alexfertel: I think that looks good 👍

developeruche: alright

contracts/src/cryptography/eip712.rs
developeruche: please I need your assertion on this, @alexfertel

alexfertel: I think we only need one `eip712_domain`, cause that's the one which may require a custom implementation. Everything else can be associated consts. I may be wrong though! I'd need to see the actual implementation to know.

Correct me if I'm wrong, but we can compute everything at compile time based on the consumer's implementation. Let's give that I try. If you need a keccak at compile time, you can use `const_keccak`.

bidzyyys: From what I can see it looks like it should be doable and it would be a good gas optimization when having everything calculated at compile time. With this attitude, we would need to have integration tests (E2E) confirming that we are able to set & calculate everything in "dummy" parent contract.

bidzyyys (commented):
I like the idea of having `EIP712` resolved at compile time!

contracts/src/cryptography/eip712.rs
bidzyyys: What I can see in [ERC-20 Permit extension](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/extensions/ERC20Permit.sol), we need:
- `domain_separator_v4`
- `hash_typed_data_v4`

alexfertel: If we have everything ready at compile-time, we don't need those, since the first one only checks for the cached values (which will be there since we compute them at compile-time) and the second one is this:

```rust
        assembly {
            let ptr := mload(0x40)
            mstore(ptr, hex"19_01")
            mstore(add(ptr, 0x02), domainSeparator)
            mstore(add(ptr, 0x22), structHash)
            digest := keccak256(ptr, 0x42)
        }
```

which you can see is `keccak256([&[0x19, 0x01], &domain_hash[..], &struct_hash[..]].concat())`, which we could do at compile-time using `const_keccak`.

I may be wrong though! If at any point you find issues, let me know.

bidzyyys (approved):
Some minor fixed to be applied with #43

alexfertel (approved):

PR #199, opened 2024-07-16
feat: add ECDSA smart contract

### LINKED ISSUE #17: ECDSA

bidzyyys:
Resolves #17 

#### PR Checklist
- [x] Tests
- [x] Documentation


contracts/src/utils/cryptography/ecdsa.rs
```diff
@@ -0,0 +1,248 @@
...
+/// * `hash` - Hash of the message.
+/// * `v` - `v` value from the signature.
+/// * `r` - `r` value from the signature.
+/// * `s` - `s` value from the signature.
+fn prepare_calldata(hash: B256, v: u8, r: B256, s: B256) -> Vec<u8> {
+    let calldata = ECRecoverData { hash: *hash, v, r: *r, s: *s };
+    ECRecoverData::encode(&calldata)
+}
+
+/// Validates `S` value.
```
alexfertel, line 167:
```suggestion
/// Validates the `s` value of a signature.
```

contracts/src/utils/cryptography/ecdsa.rs
```diff
@@ -0,0 +1,248 @@
...
+///
+/// * If `ecrecover` precompile fail to execute.
+pub fn recover(
+    storage: &mut impl TopLevelStorage,
+    hash: B256,
+    v: u8,
+    r: B256,
+    s: B256,
+) -> Result<Address, Error> {
+    validate_s_value(&s)?;
```
alexfertel, line 96: nit: I'd say we inline this `fn`. it's only used once and it's ambiguous: what are we validating?

If we decide to keep it, what do you think of a name like: `s_in_range` or `check_if_malleable`.

contracts/src/utils/cryptography/ecdsa.rs
```diff
@@ -0,0 +1,248 @@
...
+    // If the signature is valid (and not malleable), return the signer address.
+    evm_recover(storage, hash, v, r, s)
```
alexfertel, line 98: I'd rather rename this to something else, like `_recover`.

I would also inline it, but I don't feel strongly about it, so feel free to ignore.

contracts/src/utils/cryptography/ecdsa.rs
```diff
@@ -0,0 +1,248 @@
...
+        // it should return `ECDSAInvalidSignature` error.
+        return Err(ECDSAInvalidSignature {}.into());
+    }
+
+    let recovered =
+        call::static_call(Call::new_in(storage), ECRECOVER_ADDR, &calldata)
+            .expect("should call `ecrecover` precompile");
+
+    let recovered = B256::from_slice(&recovered);
+    let recovered = Address::from_word(recovered);
```
alexfertel, line 146: I think there's probably an easier way to do this. Something like:

```suggestion
    let recovered = Address::from(recovered.as_slice()[12..]);
```

Or maybe `Address::from(&recovered[12..])` directly. Could you try that out?

contracts/src/utils/cryptography/ecdsa.rs
```diff
@@ -0,0 +1,248 @@
...
+use stylus_proc::SolidityError;
+use stylus_sdk::{
+    call::{self, Call},
+    storage::TopLevelStorage,
+};
+
+const ECRECOVER_ADDR: Address =
+    address!("0000000000000000000000000000000000000001");
+
+const EIP2_VALUE: U256 = uint!(
```
alexfertel, line 18: This should be more concrete about what it refers to because the name is a bit ambiguous. Maybe `SIGNATURE_S_UPPER_BOUND`? Also, probably should be public.

contracts/src/utils/cryptography/ecdsa.rs
```diff
@@ -0,0 +1,248 @@
...
+/// # Errors
+///
+/// * If the `s` value is grater than `EIP2_VALUE`, then the error
+/// [`Error::ECDSAInvalidSignatureS`] is returned.
+/// * If the recovered address is `Address::ZERO`, then the error
+/// [`Error::InvalidSignature`] is returned.
+///
+/// # Panics
+///
+/// * If `ecrecover` precompile fail to execute.
```
alexfertel, line 123:
```suggestion
/// * If the `ecrecover` precompile fails to execute.
```

contracts/src/utils/cryptography/ecdsa.rs
```diff
@@ -0,0 +1,248 @@
...
+    r: B256,
+    s: B256,
+) -> Result<Address, Error> {
+    let calldata = prepare_calldata(hash, v, r, s);
+
+    if v == 0 || v == 1 {
+        // `ecrecover` panics for this values
+        // but following the Solidity tests
+        // https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/test/utils/cryptography/ECDSA.test.js
+        // it should return `ECDSAInvalidSignature` error.
```
alexfertel, line 137:
```suggestion
        // `ecrecover` panics for these values
        // but following the Solidity tests
        // https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/test/utils/cryptography/ECDSA.test.js
        // it should return `ECDSAInvalidSignature` error.
```

contracts/src/utils/cryptography/ecdsa.rs
```diff
@@ -0,0 +1,248 @@
...
+
+/// Prepares call data for `ecrecover` EVM precompile.
+///
+/// # Arguments
+///
+/// * `hash` - Hash of the message.
+/// * `v` - `v` value from the signature.
+/// * `r` - `r` value from the signature.
+/// * `s` - `s` value from the signature.
+fn prepare_calldata(hash: B256, v: u8, r: B256, s: B256) -> Vec<u8> {
```
alexfertel, line 162: nit: maybe call this `abi_encode`? maybe `encode_calldata`? `prepare` is not a very useful word.

examples/cryptography/src/lib.rs
alexfertel, line 1: I think this example should be called `ecdsa` and not `cryptography`.

examples/cryptography/tests/abi/mod.rs
alexfertel, line 1: Same here, all of these should be `ecdsa` instead of `cryptography`.

examples/cryptography/tests/cryptography.rs
alexfertel, line 1: These look great 💯

alexfertel (changes requested):
Great work! Thanks for the effort and the patience here.

I mostly have nits and some naming concerns. Behaviour-wise, it looks perfect 👍

contracts/src/utils/cryptography/ecdsa.rs
bidzyyys: I prefer keeping validation functions so that they can be unit tested.  I will rename it.

examples/cryptography/src/lib.rs
bidzyyys: I see your point here. The reason is why I named it `cryptography` is that I think that this example should contain `EIP-712` and future cryptography contracts or libs.

examples/cryptography/tests/abi/mod.rs
bidzyyys: Answered here -> https://github.com/OpenZeppelin/rust-contracts-stylus/pull/199#discussion_r1698305608

examples/cryptography/src/lib.rs
bidzyyys: Wdyt @alexfertel?

contracts/src/utils/cryptography/ecdsa.rs
bidzyyys: Done.

contracts/src/utils/cryptography/ecdsa.rs
```diff
@@ -0,0 +1,248 @@
...
+/// # Arguments
+///
+/// * `s` - `s` value from the signature.
+///
+/// # Errors
+///
+/// * If the `s` value is grater than `EIP2_VALUE`, then the error
+/// [`Error::ECDSAInvalidSignatureS`] is returned.
+fn validate_s_value(s: &B256) -> Result<(), Error> {
+    if U256::from_be_slice(s.as_slice()) > EIP2_VALUE {
```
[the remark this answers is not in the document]
bidzyyys: We cannot compare `B256` unfortunately :(

qalisander (approved):
Looks good @bidzyyys
Great job!

examples/cryptography/src/lib.rs
qalisander: Seems minor thing to me. But in future we can rename it to `cryptography` or smth else once example will be updated. For now make sense to have it as `ecdsa`

alexfertel: Yeah, even if the example includes EIP-712, it still should be named `ecdsa` because that eip is only about how to hash structured things.

I'd rather we not use a general name that can lead to confusion!

alexfertel: For future crypto primitives that we include, like say, FHE, we should have separate examples of actual practical usage.

bidzyyys: Ok, thx guys, you convinced me, renaming to `ecdsa`.

contracts/src/utils/cryptography/ecdsa.rs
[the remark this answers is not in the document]
bidzyyys: Handled this way -> https://github.com/OpenZeppelin/rust-contracts-stylus/pull/199/commits/287c55da218c58945c7f247850eead0ca3851dfb

alexfertel (approved):
Let's goooo

PR #200, opened 2024-07-16
refactor: remove DefaultStorageLayout macro

qalisander:
Seems there is a more simple way to initialize contract for `motsu` unit tests.
Remove macro that we don't need to use.

Follows up #51

#### PR Checklist

- [-] Tests
- [-] Documentation


qalisander:
In brief this pr eliminates `DefaultStorageLayout` attribute since `StorageType` has method `StorageType::new(..)`. And it already implements construction in a "smart" recursive way

lib/motsu/src/context.rs
```diff
@@ -22,9 +24,22 @@ pub fn acquire_storage() -> MutexGuard<'static, ()> {
...
     reset_storage();
 }
+
+/// Initializes fields of contract storage and child contract storages with
+/// default values.
+pub trait DefaultStorage: StorageType {
+    /// Initializes fields of contract storage and child contract storages with
+    /// default values.
+    #[must_use]
+    fn default() -> Self {
```
ggonzalez94, line 40: Maybe I'm missing something... but this would effectively give an entire slot(32 bytes) to every contract variable instead of following the [evm storage layout](https://docs.soliditylang.org/en/latest/internals/layout_in_storage.html).

ggonzalez94 (commented):
This looks clean, but I'm not sure if it follows the storage layout for the EVM. Also I'm curious what contract(s) did you find a limit with the current implementation?

qalisander:
Those contracts that contains the other contracts inside.
Like this as I remember had issues with that macro
`sol_storage! {
    pub struct Erc721Consecutive {
        Erc721 erc721;
        Trace160 _sequential_ownership;
        BitMap _sequentian_burn;
        bool _initialized
    }
}`

lib/motsu/src/context.rs
qalisander: Are you assuming that layout should be different at stylus sdk side from [evm storage layout](https://docs.soliditylang.org/en/latest/internals/layout_in_storage.html)?
When the trait `StorageType` is implemented with `sol_storage!` macro. It seems for me that all inner types like `StorageMap` or `StorageBool` relying on the similar `SLOT_BYTES` constant. Same as it was beneath `DefaultStorageLayout` attribute. If I'm not mistaken of course

ggonzalez94: The function `new`  comes from the StorageType itself, so if we call it we need to pass the right slot and offset(that's why it is unsafe, because it is easy to mess up with). If we just call `new` like in the code above we are always setting slot 0 if you think about it. 
Because of how our unit tests are structured they don't fail even if the `default` implementation is not entirely correct(but for more complex tests that might not be the case). 
The tests that you removed on this PR were intended to check that(that the implementation of `default` was following the storage layout of the EVM), but they are also not perfect since there's no easy way to check the slot and offset of a given variable using the sdk.
**I would say that we should try to generalize the previous implementation to make it work with contracts that have other contracts inside if we need to, but we should make it following the EVM storage layout rules.** 
cc @alexfertel if you have more insights/opinion

ggonzalez94: One thing I like and we might want to keep is how you made it implicit(not needing to add the attribute on top of contract)

alexfertel: Yeah, understood why this works. You can see the recursion [here](https://github.com/OffchainLabs/stylus-sdk-rs/blob/17fc86a4615af0769e8f528f32276e067702d52e/stylus-proc/src/storage/mod.rs#L144-L154), and you can see the call for the root contract [here](https://github.com/OffchainLabs/stylus-sdk-rs/blob/17fc86a4615af0769e8f528f32276e067702d52e/stylus-proc/src/methods/entrypoint.rs#L37).

With these changes we would be doing exactly that when we call `default()`. 

One interesting thing to note here is that `offset` is completely unnecessary for the root call: it doesn't matter if we pass something that is not `0u8`, `offset` is not used.

```rust
unsafe fn new(mut root: stylus_sdk::alloy_primitives::U256, offset: u8) -> Self {
    use stylus_sdk::{storage, alloy_primitives};
    debug_assert!(offset == 0);

    let mut space: usize = 32;
    let mut slot: usize = 0;
    let accessor = Self {
        /// This is expanded per contract field.
		/// You can see `offset` is not used anywhere here.
        #ident: {
            let bytes = <#ty as storage::StorageType>::SLOT_BYTES;
            let words = <#ty as storage::StorageType>::REQUIRED_SLOTS;
            if space < bytes {
                space = 32;
                slot += 1;
            }
            space -= bytes;

            let root = root + alloy_primitives::U256::from(slot);
            let field = <#ty as storage::StorageType>::new(root, space as u8);
            if words > 0 {
                slot += words;
                space = 32;
            }
            field
        },
    };
    accessor
}
```

This would could impact the storage layout if we could modify inner recursive calls.

---

I think this also means that for types at the same level:
- If they are primitives that fit in the same slot they will overwrite each other.
- If they are composite types they will always be put in separate slots, even if they fit with each other.

I'm probably wrong, so I will work on a small PoC for this.

alexfertel (approved):
LGTM! Great find 🚀

qalisander:
@ggonzalez94 So yeah. It was definitely my guess that `DefaultStorageLayout` macro will not work with complex types. That is why I hasn't checked it. And it works in reality. Since `StorageType::new(..)` already has recursive implementation.
If you didn't knew before how `StorageType::new(..)` is implemented with `sol_storage` macro. I think you would also expect that it won't work in this case.
If this pr doesn't add any new feature. I think I'll change title from `feature` to `refactoring` cc @alexfertel

alexfertel:
Actually, do you mind fixing docs in this PR as well @qalisander. There are a few places where me have outdated comments like in the [README](https://github.com/OpenZeppelin/rust-contracts-stylus/tree/main/lib/motsu) and the lib.rs docs, etc.

lib/motsu/README.md
```diff
@@ -12,30 +12,14 @@ Japanese -- we hold a stylus in our hand.
...
-Note that we require contracts to implement `core::default::Default`. This
-implementation should match the way solidity would lay out the contract's state
-in storage, so that the tests are as close as possible to the real environment.
+Note that we require contracts to implement `stylus_sdk::prelude::StorageType`.
+This trait is typically implemented by default with `stylus_proc::sol_storage` macro.
 
 ```rust
 #[cfg(test)]
 mod tests {
     use contracts::token::erc20::Erc20;
```
alexfertel, line 21: Does this doc test pass without importing the new trait?

alexfertel (approved):
Beautiful! Great work, this is a much better approach 😌

bidzyyys (approved):
Great refactor, it simplifies the code a lot! 🚀 
LGTM!

lib/motsu/src/context.rs
alexfertel: Okay, had a gap in my knowledge: `struct`s are always put in the next slot regardless if there is space.

So everything works well!

ggonzalez94 (approved):
Great work! It pushed us to understand better how `StorageType` was implemented and certainly simplifies the code a lot :rocket:

PR #201, opened 2024-07-16
chore: move check-wasm.sh to scripts

qalisander:
Since check-wasm.sh is generic script and can be used not just from github actions workflow. Move it to ./scripts directory.


alexfertel (approved):

bidzyyys (approved):
LGTM! 🚀

PR #202, opened 2024-07-17
feat: erc721 consecutive extension

### LINKED ISSUE #68: ERC-721 Consecutive
```[tasklist]
### Tasks
- [ ] https://github.com/OpenZeppelin/rust-contracts-stylus/issues/76
- [ ] https://github.com/OpenZeppelin/rust-contracts-stylus/issues/77
```

qalisander:
This extension matches [the solidity implementation](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/483fc3f1f89ef6eb0803d578a4a17346d94359f7/contracts/token/ERC721/extensions/ERC721Consecutive.sol#L30) except one thing:
- to be able to override erc721 methods `_owner_of_inner` and ` _update` some functions are duplicated from erc721 token standard

Resolves #68

#### PR Checklist

- [x] Tests
- [x] Documentation


qalisander:
Code coverage anticipated to drop since not every function copied from erc721 covered with tests. Seems this approach will also require copy paste of unit tests from erc721:( cc @alexfertel @bidzyyys

contracts/src/token/erc721/extensions/consecutive.rs
```diff
@@ -0,0 +1,1079 @@
...
+    /// Override of [`Erc721::_owner_of_inner`] that checks the sequential
+    /// ownership structure for tokens that have been minted as part of a
+    /// batch, and not yet transferred.
+    ///
+    /// # Arguments
+    ///
+    /// * `&self` - Read access to the contract's state.
+    /// * `token_id` - Token id as a number.
+    pub fn _owner_of_inner(&self, token_id: U256) -> Address {
+        let owner = self.__owner_of_inner(token_id);
```
alexfertel, line 162: Why not inline `__owner_of_inner` to `self.erc721._owner_of_inner`? 

Also, I just realized that we don't need to suffix `_inner` in the Erc721 implementation.

contracts/src/token/erc721/extensions/consecutive.rs
```diff
@@ -0,0 +1,1079 @@
...
+            || token_id > U256::from(U96::MAX)
+            || token_id > U256::from(U96::MAX)
+            || token_id < U256::from(FIRST_CONSECUTIVE_ID)
+        {
+            return owner;
+        }
+
+        // Otherwise, check the token was not burned, and fetch ownership from
+        // the anchors.
+        // NOTE: no need for safe cast,
```
alexfertel, line 175: Let's add the full comment here, and move the note to the relevant line.

contracts/src/token/erc721/extensions/consecutive.rs
```diff
@@ -0,0 +1,1079 @@
...
+    ///
+    /// Emits a [`ConsecutiveTransfer`] event.
+    pub fn _mint_consecutive(
+        &mut self,
+        to: Address,
+        batch_size: U96,
+    ) -> Result<U96, Error> {
+        let next = self._next_consecutive_id();
+
+        if batch_size > U96::ZERO {
```
alexfertel, line 221: Let's keep the comments from the original implementation with proper capitalization and ending dots:

```solidity
        // minting a batch of size 0 is a no-op
        if (batchSize > 0) {
            if (address(this).code.length > 0) {
                revert ERC721ForbiddenBatchMint();
            }
            if (to == address(0)) {
                revert ERC721InvalidReceiver(address(0));
            }

            uint256 maxBatchSize = _maxBatchSize();
            if (batchSize > maxBatchSize) {
                revert ERC721ExceededMaxBatchMint(batchSize, maxBatchSize);
            }

            // push an ownership checkpoint & emit event
            uint96 last = next + batchSize - 1;
            _sequentialOwnership.push(last, uint160(to));

            // The invariant required by this function is preserved because the new sequentialOwnership checkpoint
            // is attributing ownership of `batchSize` new tokens to account `to`.
            _increaseBalance(to, batchSize);

            emit ConsecutiveTransfer(next, last, address(0), to);
        }
```

contracts/src/token/erc721/extensions/consecutive.rs
```diff
@@ -0,0 +1,1079 @@
...
+            }
+
+            if to.is_zero() {
+                return Err(erc721::Error::InvalidReceiver(
+                    ERC721InvalidReceiver { receiver: Address::ZERO },
+                )
+                .into());
+            }
+
+            if batch_size > MAX_BATCH_SIZE.to() {
```
alexfertel, line 233:
```suggestion
            if batch_size > MAX_BATCH_SIZE {
```

contracts/src/token/erc721/extensions/consecutive.rs
```diff
@@ -0,0 +1,1079 @@
...
+                return Err(ERC721ExceededMaxBatchMint {
+                    batchSize: U256::from(batch_size),
+                    maxBatch: U256::from(MAX_BATCH_SIZE),
+                }
+                .into());
+            }
+
+            let last = next + batch_size - uint!(1_U96);
+            self._sequential_ownership
+                .push(last, U160::from_be_bytes(to.into_array()))?;
```
alexfertel, line 243:
```suggestion
            self._sequential_ownership
                .push(last, to.into())?;
```

contracts/src/token/erc721/extensions/consecutive.rs
```diff
@@ -0,0 +1,1079 @@
...
+    /// Should be called to restrict consecutive mint after.
+    /// After this function being called, every call to
+    /// [`Self::_mint_consecutive`] will fail.
+    ///
+    /// # Arguments
+    ///
+    /// * `&self` - Write access to the contract's state.
+    pub fn _stop_mint_consecutive(&mut self) {
+        self._initialized.set(true);
+    }
```
alexfertel, line 265: nit: Let's inline this function at the call site. It's not called anywhere else.

contracts/src/token/erc721/extensions/consecutive.rs
```diff
@@ -0,0 +1,1079 @@
...
+                    _checkpoints: factory.create(),
+                },
+                _sequentian_burn: BitMap { _data: factory.create() },
+                _initialized: factory.create(),
+            }
+        }
+    }
+
+    const BOB: Address = address!("F4EaCDAbEf3c8f1EdE91b6f2A6840bc2E4DD3526");
+    const DAVE: Address = address!("0BB78F7e7132d1651B4Fd884B7624394e92156F1");
```
alexfertel, line 876:
```suggestion
```

contracts/src/token/erc721/extensions/consecutive.rs
```diff
@@ -0,0 +1,1079 @@
...
+        // only mint after construction
+        if previous_owner == Address::ZERO && !self._initialized.get() {
+            return Err(ERC721ForbiddenMint {}.into());
+        }
+
+        // record burn
+        if to == Address::ZERO // if we burn
+            && token_id < U256::from(self._next_consecutive_id()) // and the tokenId was minted in a batch
+            && !self._sequentian_burn.get(token_id)
+        // and the token was never marked as burnt
```
alexfertel, line 314:
```suggestion
        let next_id = self._next_consecutive_id();
        let is_burn = to == Address::ZERO;
        let token_minted_in_batch = token_id < U256::from(next_id);
        let not_marked_burned = !self._sequentian_burn.get(token_id);
        // Record burn.
        if is_burn && token_minted_in_batch && not_marked_burned {
            self._sequentian_burn.set(token_id);
        }
```

alexfertel (changes requested):
Looking good! First pass ✅

contracts/src/token/erc721/extensions/consecutive.rs
```diff
@@ -0,0 +1,1079 @@
...
+                    _operator_approvals: factory.create(),
+                },
+                _sequential_ownership: Trace160 {
+                    _checkpoints: factory.create(),
+                },
+                _sequentian_burn: BitMap { _data: factory.create() },
+                _initialized: factory.create(),
+            }
+        }
+    }
```
alexfertel, line 873: btw, I removed this code, and added the `#[cfg_attr(all(test, feature = "std"), derive(motsu::DefaultStorageLayout))]` to `Erc721Consecutive` and all the tests pass.

Do you mind sharing the issues you were having? cc: @ggonzalez94

qalisander: Tough to say. I suspect smth else was wrong or I hasn't checked and added default implementation. But yeah. It works just because `StorageType::new` for `erc721` was recursively implemented with `sol_storage!` macro. And that was unforeseen by us before.

contracts/src/token/erc721/extensions/consecutive.rs
qalisander: good catch

contracts/src/token/erc721/extensions/consecutive.rs
```diff
@@ -0,0 +1,1079 @@
...
+    /// * `&self` - Read access to the contract's state.
+    /// * `token_id` - Token id as a number.
+    pub fn _owner_of_inner(&self, token_id: U256) -> Address {
+        let owner = self.__owner_of_inner(token_id);
+        // If token is owned by the core, or beyond consecutive range, return
+        // base value
+        if owner != Address::ZERO
+            || token_id > U256::from(U96::MAX)
+            || token_id > U256::from(U96::MAX)
+            || token_id < U256::from(FIRST_CONSECUTIVE_ID)
```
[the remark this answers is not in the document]
qalisander: make sense, I'll change. But at solidity seems they don't care

contracts/src/token/erc721/extensions/consecutive.rs
```diff
@@ -0,0 +1,1054 @@
...
+            checkpoints::{Trace160, U96},
+        },
+    },
+};
+
+sol_storage! {
+    /// State of an [`Erc72Erc721Consecutive`] token.
+    #[cfg_attr(all(test, feature = "std"), derive(motsu::DefaultStorageLayout))]
+    pub struct Erc721Consecutive {
+        /// Erc721 contract storage.
```
bidzyyys, line 64: I am not sure if we should put each contract(struct) name in the docs in `[]` or not, like `[Erc721]`

contracts/src/token/erc721/extensions/consecutive.rs
```diff
@@ -0,0 +1,1054 @@
...
+
+    /// Batch burn is not supported.
+    #[derive(Debug)]
+    #[allow(missing_docs)]
+    error ERC721ForbiddenBatchBurn();
+}
+
+#[derive(SolidityError, Debug)]
+pub enum Error {
+    /// Error type from erc721 contract [`erc721::Error`]
```
bidzyyys, line 116: `[Erc721]` ?

contracts/src/token/erc721/extensions/consecutive.rs
```diff
@@ -0,0 +1,1079 @@
...
+// TODO: add an option to override these constants
+
+/// Maximum size of a batch of consecutive tokens. This is designed to limit
+/// stress on off-chain indexing services that have to record one entry per
+/// token, and have protections against "unreasonably large" batches of tokens.
+pub const MAX_BATCH_SIZE: U96 = uint!(5000_U96);
+
+/// Used to offset the first token id in
+/// [`Erc721Consecutive::_next_consecutive_id`]
+pub const FIRST_CONSECUTIVE_ID: U96 = uint!(0_U96);
```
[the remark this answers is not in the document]
bidzyyys: +1

contracts/src/token/erc721/extensions/consecutive.rs
```diff
@@ -0,0 +1,1054 @@
...
+    /// CAUTION: Does not invoke `onERC721Received` on the receiver.
+    ///
+    /// # Arguments
+    ///
+    /// * `&self` - Write access to the contract's state.
+    /// * `token_id` - Token id as a number.
+    ///
+    /// # Errors
+    ///
+    /// If to is [`Address::ZERO`] error [`rc721::Error::InvalidReceiver`] is
```
bidzyyys, line 209: Nit: IMHO we should follow the formatting and structure for code docs, like here https://github.com/OpenZeppelin/rust-contracts-stylus/blob/main/contracts/src/token/erc721/mod.rs#L277-L289

contracts/src/token/erc721/extensions/consecutive.rs
```diff
@@ -0,0 +1,1054 @@
...
+                .into());
+            }
+
+            // Push an ownership checkpoint & emit event.
+            let last = next + batch_size - uint!(1_U96);
+            self._sequential_ownership.push(last, to.into())?;
+
+            // The invariant required by this function is preserved because the
+            // new sequentialOwnership checkpoint is attributing
+            // ownership of `batchSize` new tokens to account `to`.
```
bidzyyys, line 251: `batch_size`

contracts/src/token/erc721/extensions/consecutive.rs
```diff
@@ -0,0 +1,1054 @@
...
+        self._require_owned(token_id)?;
+        Ok(self.erc721._get_approved_inner(token_id))
+    }
+
+    fn is_approved_for_all(&self, owner: Address, operator: Address) -> bool {
+        self.erc721.is_approved_for_all(owner, operator)
+    }
+}
+
+// erc721 related implementation:
```
bidzyyys, line 438: `ERC-721`

contracts/src/token/erc721/extensions/consecutive.rs
```diff
@@ -0,0 +1,1054 @@
...
+    /// mints (or burns) if the current owner (or `to`) is the `Address::ZERO`.
+    /// Returns the owner of the `token_id` before the update.
+    ///
+    /// The `auth` argument is optional. If the value passed is non-zero, then
+    /// this function will check that `auth` is either the owner of the
+    /// token, or approved to operate on the token (by the owner).
+    ///
+    /// NOTE: If overriding this function in a way that tracks balances, see
+    /// also [`Self::_increase_balance`].
+    fn __update(
```
bidzyyys, line 464: Why do we need `__`? Is `_update` not enough? Tbh I am confused with `__update`, `_update` etc.

contracts/src/token/erc721/extensions/consecutive.rs
```diff
@@ -0,0 +1,1054 @@
...
+
+        let init_tokens_count = uint!(10_U96);
+        init(contract, vec![alice], vec![init_tokens_count]);
+
+        let balance1 = contract
+            .balance_of(alice)
+            .expect("should return the balance of Alice");
+        assert_eq!(balance1, initial_balance + U256::from(init_tokens_count));
+
+        // Check non-consecutive mint
```
bidzyyys, line 887: We agreed that comments should end with `.`

bidzyyys (commented):
Good job, left some comments.

contracts/src/token/erc721/extensions/consecutive.rs
alexfertel: > This function is assumed to be called a the end user's side

Note that `user` here is a developer, not a user of the deployed contract.

This is an artifact of having `_initialized` as we discussed previously -- functions called from a constructor in Solidity won't be included in the deployed contract, that's why `mint_consecutive` *cannot* be called after the contract is deployed, and why it's a requirement in Solidity, see [the comments](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/19a657bef8354f2a655900654955739b70dfbde9/contracts/token/ERC721/extensions/ERC721Consecutive.sol#L95).

---

This actually surfaces a different issue -- We expect consumers to implement a complex `init` function. I think we should instead implement `init` ourselves, and disallow calling by checking if `initialized == true` at the start of the function. I see we have this function in the example:

```rust
pub fn init(
    &mut self,
    receivers: Vec<Address>,
    batches: Vec<U256>,
) -> Result<(), Error> {
    let len = batches.len();
    for i in 0..len {
        let receiver = receivers[i];
        let batch = batches[i];
        let _ = self
            .erc721_consecutive
            ._mint_consecutive(receiver, U96::from(batch))?;
    }
    self.erc721_consecutive._stop_mint_consecutive();
    Ok(())
}
```

Shouldn't this live in our contract? And then users just call `self.erc721_consecutive.init()`, which in turn sets `initialized`?

---

I also see we have a `constructor.sol` defined for this example, but we don't really need it since we are expecting users to call `init()`?

I think we either choose to have a constructor, which I think is what we should do, or instead have the `init()` mechanism.

contracts/src/token/erc721/extensions/consecutive.rs
```diff
@@ -0,0 +1,1054 @@
...
+                ERC721NonexistentToken { token_id },
+            )
+            .into());
+        }
+        Ok(owner)
+    }
+}
+
+#[cfg(all(test, feature = "std"))]
+mod test {
```
alexfertel, line 829: Let's match the other modules and use `tests`.

contracts/src/token/erc721/extensions/consecutive.rs
```diff
@@ -0,0 +1,1054 @@
...
+/// stress on off-chain indexing services that have to record one entry per
+/// token, and have protections against "unreasonably large" batches of tokens.
+pub const MAX_BATCH_SIZE: U96 = uint!(5000_U96);
+
+/// Used to offset the first token id in
+/// [`Erc721Consecutive::_next_consecutive_id`]
+pub const FIRST_CONSECUTIVE_ID: U96 = uint!(0_U96);
+
+/// Consecutive extension related implementation:
+impl Erc721Consecutive {
```
alexfertel, line 156: Let's match the other modules and have the `#[external]` impl block appear first in the file.

examples/erc721-consecutive/src/constructor.sol
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: MIT
```
alexfertel, line 1: As mentioned in a previous comment, we either don't need this file or we don't need an `init()` function.

examples/erc721-consecutive/tests/erc721-consecutive.rs
```diff
@@ -0,0 +1,234 @@
...
+async fn burns(alice: Account) -> eyre::Result<()> {
+    let contract_addr = deploy(alice.url(), &alice.pk()).await?;
+    let contract = Erc721::new(contract_addr, &alice.wallet);
+
+    // Mint batch of 1000 tokens to Alice
+    let _ =
+        watch!(contract.init(vec![alice.address()], vec![uint!(1000_U256)]))?;
+    let first_consecutive_token_id = U256::from(FIRST_CONSECUTIVE_ID);
+
+    // Check consecutive token burn
```
alexfertel, line 183: nit: Let's add `.`s to all the comments 🙏

contracts/src/token/erc721/extensions/consecutive.rs
```diff
@@ -0,0 +1,1054 @@
...
+/// Maximum size of a batch of consecutive tokens. This is designed to limit
+/// stress on off-chain indexing services that have to record one entry per
+/// token, and have protections against "unreasonably large" batches of tokens.
+pub const MAX_BATCH_SIZE: U96 = uint!(5000_U96);
+
+/// Used to offset the first token id in
+/// [`Erc721Consecutive::_next_consecutive_id`]
+pub const FIRST_CONSECUTIVE_ID: U96 = uint!(0_U96);
+
+/// Consecutive extension related implementation:
```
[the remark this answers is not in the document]
alexfertel: Actually would make these more prominent. Something like:

```rust
// ************** Consecutive **************
```

I think we can explore what it would look like.

contracts/src/token/erc721/extensions/consecutive.rs
qalisander: As I mentioned in the pr, part of the code is duplicated from erc721. `__update` function name highlights function copied from erc721. `_update` overrides that function. Not the best naming though.. Which approach you think could look better @bidzyyys @alexfertel ?

contracts/src/token/erc721/extensions/consecutive.rs
qalisander: why not!

contracts/src/token/erc721/extensions/consecutive.rs
```diff
@@ -0,0 +1,1079 @@
...
+
+sol! {
+    /// Emitted when the tokens from `fromTokenId` to `toTokenId` are transferred from `fromAddress` to `toAddress`.
+    ///
+    /// * `fromTokenId` - First token being transferred.
+    /// * `toTokenId` - Last token being transferred.
+    /// * `fromAddress` - Address from which tokens will be transferred.
+    /// * `toAddress` - Address where the tokens will be transferred to.
+    event ConsecutiveTransfer(
+        uint256 indexed fromTokenId,
```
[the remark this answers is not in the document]
qalisander: valid

contracts/src/token/erc721/extensions/consecutive.rs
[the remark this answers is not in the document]
qalisander: Opt in to  `************** Erc721 **************` pattern

contracts/src/token/erc721/extensions/consecutive.rs
qalisander: Initially I kept all boilerplate related to Erc721 at the end of file. `#[external]` just adds external implementation for erc721 and nothing else for consecutive. But yeah. Move it to the top

contracts/src/token/erc721/extensions/consecutive.rs
qalisander: The other solution could be: Create small functions with corresponding names that will be lazily evaluated. Or leave it as it is right now (like in solidity).

alexfertel: Ah, you are totally right. Yeah, in this case I think being more performant is better. So let's just ignore 👍

examples/erc721-consecutive/src/constructor.sol
```diff
@@ -0,0 +1,23 @@
...
+pragma solidity ^0.8.21;
+
+contract Erc721Example {
```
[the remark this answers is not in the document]
qalisander: Just removed solidity constructor since it is not needed

contracts/src/token/erc721/extensions/consecutive.rs
```diff
@@ -0,0 +1,1079 @@
...
+    fn encode(self) -> alloc::vec::Vec<u8> {
+        self.into()
+    }
+}
+
+impl MethodError for checkpoints::Error {
+    fn encode(self) -> alloc::vec::Vec<u8> {
+        self.into()
+    }
+}
```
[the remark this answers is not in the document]
qalisander: opened an issue there https://github.com/OffchainLabs/stylus-sdk-rs/issues/135

contracts/src/token/erc721/extensions/consecutive.rs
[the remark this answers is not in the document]
qalisander: added PoC to the current branch. If it seems bad for you guys. I'll revert commit

contracts/src/token/erc721/extensions/consecutive.rs
alexfertel: Yeah, I gave this a bit more thought and I am more convinced that the way to go is to just use a Solidity constructor. It doesn't matter if they are too big.

This is better because:

- It's consistent with the rest of the library.
- We would want it to get audited.
- A user would probably not change much in the constructor of this particular contract.
- It's more secure. (Less vectors vs initializer)

Let's give it a shot 🙏

ggonzalez94: @qalisander wouldn't this be more work? Since we are just adding the contract I think we should aim to do it with a constructor(for consistency and all the other reasons that @alexfertel detailed above). Could you give it a shot so we can see how feasible it is and the level of effort? Thanks!

qalisander: @ggonzalez94 I'm just offering more granular way to make changes. Since I've already tried an approach with solidity constructor and it wasn't compiling that simple. It can take a bit more time. But yeah I'll definitely revisit it in the other branch

contracts/src/token/erc721/extensions/consecutive.rs
alexfertel: I gave this another look and I think I don't understand very well why we *need* to duplicate it, since we can just call `self.erc721._update()`, right?

qalisander: I think no, since every time we call an `__update()` function `_owner_of_inner()` is called inside of it.
`_owner_of_inner()` is a bit tweaked function, that also uses `_sequential_burn` and `_sequential_ownership` storage fields. That is how I've tried to emulate solidity inheritance

qalisander: How do you think guys it can be done better?

alexfertel: Yeah, I see it now. We need to use the new `_owner_of_inner`, so we can't call `self.erc721._update` because it uses the old `_owner_of_inner`.

I see `__update` only gets called once in line 408, inside `_update`. We have two options:
- Rename `__update` to something else, like for example: `_update_consecutive`.
- Inline `__update`.

wdyt?

qalisander: Yeah! As I remember `__update` was unlined before. Then I thought it would be convenient for maintenance to split unique code of consecutive and erc721 duplicated one. And extracted it as an `__update` function.
I think the public function should have the same name `_update` as erc721 base implementation has. To not surprise users.
`__update` is private right now and we are free to use any name. How do u feel about  `_update_base` or `_update_super` ?

alexfertel: Okay, let's go with `_update_base`. Let's also remove the `NOTE` comment since we can't override it, and let's document that this is a copy of `_update`, but we have to duplicate it because

> We need to use the new _owner_of_inner, so we can't call self.erc721._update because it uses the old _owner_of_inner.

bidzyyys (commented):
Looks good in a second pass. I agree with comments left by @alexfertel that are still to be resolved.

contracts/src/token/erc721/extensions/consecutive.rs
qalisander: Do you think it is better to put "Erc721" into [`Erc721`]?

alexfertel: I think that for struct fields it is fine if we don't

contracts/src/token/erc721/mod.rs
```diff
@@ -1853,40 +1862,40 @@ mod tests {
     }
 
     #[motsu::test]
-    fn owner_of_inner_works(contract: Erc721) {
+    fn _owner_of_works(contract: Erc721) {
```
alexfertel, line 1865: Why do we prefix these functions with an underscore?

alexfertel (approved):
LGTM! Just one comment regarding test function names.

contracts/src/token/erc721/extensions/consecutive.rs
```diff
@@ -0,0 +1,1007 @@
...
+    /// Any batch mint not emitting the [`Transfer`] event outside of the constructor
+    /// is non ERC-721 compliant.
+    #[derive(Debug)]
+    #[allow(missing_docs)]
+    error ERC721ForbiddenBatchMint();
+
+    /// Exceeds the max number of mints per batch.
+    #[derive(Debug)]
+    #[allow(missing_docs)]
+    error ERC721ExceededMaxBatchMint(uint256 batchSize, uint256 maxBatch);
```
bidzyyys, line 99: We use sneak_case for error's attributes, check https://github.com/OpenZeppelin/rust-contracts-stylus/blob/main/contracts/src/token/erc721/mod.rs#L54-L120

bidzyyys (approved):
Great job @qalisander!
I left some minor comments and nits.

contracts/src/token/erc721/mod.rs
qalisander: Pointless. I agree

contracts/src/token/erc721/extensions/consecutive.rs
```diff
@@ -0,0 +1,1007 @@
...
+        let owner = contract
+            .owner_of(token_id)
+            .expect("should return the owner of the token");
+        assert_eq!(owner, alice);
+
+        let balance2 = contract
+            .balance_of(alice)
+            .expect("should return the balance of Alice");
+
+        assert_eq!(balance2, balance1 + uint!(1_U256));
```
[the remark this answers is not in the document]
qalisander: I'm usually leaning towards `assert_eq!(<current_value>, <expected>)`. But may be not all the time. I guess same as you in [pausable contract](https://github.com/OpenZeppelin/rust-contracts-stylus/blob/4fb347f4c20d9990f29c6dc9b6351038060b2ece/contracts/src/utils/pausable.rs#L143). I think we can vote (if it will make sense) later which formatting is preferable and change it everywhere at once.

PR #203, opened 2024-07-17
ref: rework gas usage reporting

### LINKED ISSUE #170: Refactor benchmarking infra

alexfertel:
Partially completes #170

### DIFF
--- a/benches/src/access_control.rs
+++ b/benches/src/access_control.rs
@@ -8,7 +8,7 @@ use alloy::{
 };
 use e2e::{receipt, Account};
 
-use crate::ArbOtherFields;
+use crate::{report::Report, ArbOtherFields};
 
 sol!(
     #[sol(rpc)]
@@ -33,7 +33,7 @@ const ROLE: [u8; 32] =
 const NEW_ADMIN_ROLE: [u8; 32] =
     hex!("879ce0d4bfd332649ca3552efe772a38d64a315eb70ab69689fd309c735946b5");
 
-pub async fn bench() -> eyre::Result<()> {
+pub async fn bench() -> eyre::Result<Report> {
     let alice = Account::new().await?;
     let alice_addr = alice.address();
     let alice_wallet = ProviderBuilder::new()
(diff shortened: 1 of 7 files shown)


benches/src/report.rs
```diff
@@ -0,0 +1,55 @@
...
+pub struct Reports(Vec<Report>);
+
+impl Reports {
+    pub fn merge_with(mut self, report: Report) -> Self {
+        self.0.push(report);
+        self
+    }
+}
+
+impl Display for Reports {
```
qalisander, line 29: That looks cool!

qalisander (approved):
Cool refactoring. Good job!

benches/src/access_control.rs
```diff
@@ -65,51 +65,17 @@ pub async fn bench() -> eyre::Result<()> {
...
-    );
-    println!(
-        "|{}+--------+--------+----------------------|",
-        "-".repeat(name_width + 2)
-    );
-
-    // Print each row.
-    for (func_name, receipt) in receipts {
+    let mut report = Report::new("AccessControl");
+    for (signature, receipt) in receipts {
```
bidzyyys, line 69: IMHO this loop is repeated in each benchmark; Can we have a common code for it? If not, it is fine to leave as it is.

benches/src/report.rs
bidzyyys, line 29: I like the attitude 🚀

bidzyyys (approved):
LGTM 🚀

benches/src/access_control.rs
alexfertel: yes, that's a second refactor!

bidzyyys: Maybe implementing `From<T>` trait for `Report`? I mean `Report::from(receipts)`

alexfertel:
@qalisander @bidzyyys Since in the end we didn't need this today, I'll complete the refactor by tomorrow instead.

benches/src/access_control.rs
alexfertel: We can't cause we need the contract name.

bidzyyys (approved):
Looks awesome 🚀

benches/src/access_control.rs
bidzyyys: Missed that one.

qalisander (approved):
Looks good!

PR #209, opened 2024-07-23
feat: add ERC-20 Permit extension

### LINKED ISSUE #43: ERC-20 Permit

bidzyyys:
Resolves #43 

- [x] Tests
- [x] Documentation


contracts/src/token/erc20/extensions/permit.rs
```diff
@@ -0,0 +1,196 @@
...
+            return Err(ERC2612InvalidSigner { signer, owner }.into());
+        }
+
+        let _ = erc20._approve(owner, spender, value);
+        Ok(())
+    }
+}
+
+#[cfg(all(test, feature = "std"))]
+mod tests {}
```
alexfertel, line 196: 👀

contracts/src/token/erc20/extensions/permit.rs
```diff
@@ -0,0 +1,196 @@
...
+
+use crate::{
+    token::erc20::IErc20Internal,
+    utils::{
+        cryptography::{ecdsa, eip712::IEIP712},
+        nonces::Nonces,
+    },
+};
+
+const PERMIT_TYPEHASH: B256 =
```
alexfertel, line 25:
```suggestion
// keccak256("Permit(address owner,address spender,uint256 value,uint256 nonce,uint256 deadline)")
const PERMIT_TYPEHASH: B256 =
```

contracts/src/token/erc20/extensions/permit.rs
```diff
@@ -0,0 +1,196 @@
...
+) -> B256 {
+    keccak256(StructHashTuple::encode_params(&(
+        *PERMIT_TYPEHASH,
+        owner,
+        spender,
+        value,
+        nonce,
+        deadline,
+    )))
+}
```
alexfertel, line 55: I'm not sure this should be part of our public API. We should probably inline it and then duplicate the logic in our tests, the same way the Solidity counterpart does.

contracts/src/token/erc20/extensions/permit.rs
```diff
@@ -0,0 +1,196 @@
...
+    }
+}
+
+/// NOTE: Implementation of [`TopLevelStorage`] to be able use `&mut self` when
+/// calling other contracts and not `&mut (impl TopLevelStorage +
+/// BorrowMut<Self>)`. Should be fixed in the future by the Stylus team.
+unsafe impl<T: IEIP712 + StorageType> TopLevelStorage for Permit<T> {}
+
+#[external]
+impl<T: IEIP712 + StorageType> Permit<T> {
```
alexfertel, line 96: Why do we require a `StorageType` bound here?

contracts/src/token/erc20/extensions/permit.rs
```diff
@@ -0,0 +1,196 @@
...
+    ///
+    /// # Requirements
+    ///
+    /// * `spender` cannot be the ``Address::ZERO``.
+    /// * `deadline` must be a timestamp in the future.
+    /// * `v`, `r` and `s` must be a valid secp256k1 signature from `owner`
+    /// over the EIP712-formatted function arguments.
+    /// * the signature must use `owner`'s current nonce.
+    #[allow(clippy::too_many_arguments)]
+    pub fn permit(
```
alexfertel, line 158: This should be in `#[external]`.

contracts/src/token/erc20/extensions/permit.rs
```diff
@@ -0,0 +1,196 @@
...
+    InvalidSigner(ERC2612InvalidSigner),
+}
+
+sol_storage! {
+    /// State of a Permit Contract.
+    pub struct Permit<T: IEIP712 + StorageType>{
+        /// Nonces contract.
+        Nonces nonces;
+        /// EIP-712 contract. Must implement [`IEIP712`] trait.
+        T eip712;
```
alexfertel, line 86: Could you expand on why this needs to be in storage? Why can't we just implement the trait for `Permit` and that's it?

contracts/src/token/erc20/extensions/permit.rs
```diff
@@ -0,0 +1,196 @@
...
+    pub fn permit(
+        &mut self,
+        owner: Address,
+        spender: Address,
+        value: U256,
+        deadline: U256,
+        v: u8,
+        r: B256,
+        s: B256,
+        erc20: &mut impl IErc20Internal,
```
alexfertel, line 167: Do you think we could implement this differently? I don't think the API should be like this.

contracts/src/token/erc20/extensions/permit.rs
```diff
@@ -0,0 +1,196 @@
...
+            spender,
+            value,
+            self.nonces.use_nonce(owner),
+            deadline,
+        );
+
+        let hash: B256 = self.eip712.hash_typed_data_v4(struct_hash);
+
+        let signer: Address = ecdsa::recover(self, hash, v, r, s)
+            .expect("should recover signer's address");
```
alexfertel, line 184: Why do we expect here instead of `?`?

contracts/src/token/erc20/extensions/permit.rs
```diff
@@ -0,0 +1,196 @@
...
+
+        if signer != owner {
+            return Err(ERC2612InvalidSigner { signer, owner }.into());
+        }
+
+        let _ = erc20._approve(owner, spender, value);
```
alexfertel, line 190: Should we `?` here instead of ignoring the result?

contracts/src/token/erc20/mod.rs
```diff
@@ -221,57 +222,34 @@ pub trait IErc20 {
...
-    }
-
-    fn allowance(&self, owner: Address, spender: Address) -> U256 {
-        self._allowances.get(owner).get(spender)
-    }
-
-    fn approve(
+/// Interface of an [`Erc20`] contract methods that can be used by the
+/// extensions.
+pub trait IErc20Internal {
```
alexfertel, line 227: This feels a bit awkward, could you explain why we need it?

examples/erc20/src/lib.rs
alexfertel, line 1: I think we should explore implementing `Permit` as a trait and having a separate example for it instead of here. I think this is counterintuitive -- having an empty `Eip712` in storage is not what I would expect.

examples/erc20/src/lib.rs
```diff
@@ -26,11 +26,20 @@ sol_storage! {
...
         #[borrow]
         Pausable pausable;
+        #[borrow]
+        Permit<Eip712> permit;
     }
+
+    struct Eip712 {}
+}
+
+impl IEIP712 for Eip712 {
```
alexfertel, line 36: Should it be `IEip712`?

examples/erc20/tests/erc20.rs
```diff
@@ -32,6 +49,29 @@ async fn deploy(
     e2e::deploy(rpc_url, private_key, Some(args)).await
 }
 
+fn prepare_typed_data_hash(domain_separator: B256, struct_hash: B256) -> B256 {
```
alexfertel, line 52: `prepare` is generally an ambiguous word, cause it doesn't really say much about what the preparation entails, so it's better to avoid it.

I would just rename it to `to_typed_data_hash`.

examples/erc20/tests/erc20.rs
```diff
@@ -32,6 +49,29 @@ async fn deploy(
...
+    openzeppelin_stylus::utils::cryptography::eip712::to_typed_data_hash(
+        &domain_separator,
+        &struct_hash,
+    )
+}
+
+fn prepare_struct_hash(
```
alexfertel, line 59: Same here -- `permit_struct_hash`?

examples/erc20/tests/erc20.rs
```diff
@@ -18,6 +18,23 @@ const TOKEN_NAME: &str = "Test Token";
 const TOKEN_SYMBOL: &str = "TTK";
 const CAP: U256 = uint!(1_000_000_U256);
 
+// Saturday, 1 January 2000 00:00:00
+const EXPIRED_DEADLINE: U256 = uint!(946_684_800_U256);
+
+// Wednesday, 1 January 3000 00:00:00
+const FAIR_DEADLINE: U256 = uint!(32_503_680_000_U256);
```
alexfertel, line 25: Should these be computed dynamically instead? Like `block::timestamp() -/+ something`?

examples/erc20/tests/erc20.rs
```diff
@@ -1216,3 +1256,237 @@ async fn error_when_transfer_from(alice: Account, bob: Account) -> Result<()> {
...
+            typed_data_hash.as_slice(),
+        ))
+        .await;
+
+    let err = send!(contract_alice.permit(
+        alice_addr,
+        bob_addr,
+        balance,
+        EXPIRED_DEADLINE,
+        signature.v().y_parity_byte_non_eip155().unwrap(),
```
alexfertel, line 1298: Are you sure this is correct? I think this should be `signature.v().to_u64() as u8` instead, to account for all transaction types.

examples/erc20/tests/erc20.rs
```diff
@@ -1216,3 +1256,237 @@ async fn error_when_transfer_from(alice: Account, bob: Account) -> Result<()> {
...
+        U256::ZERO,
+        FAIR_DEADLINE,
+    );
+
+    let typed_data_hash =
+        prepare_typed_data_hash(domain_separator!(contract_alice), struct_hash);
+    let signature = bob
+        .sign_hash(&alloy::primitives::B256::from_slice(
+            typed_data_hash.as_slice(),
+        ))
```
alexfertel, line 1472: Ah, this is probably because the primitives version that `alloy` uses is not the same as the `alloy_primitives` we use. Let's create an issue to fix this.

alexfertel (changes requested):
Good job!

I think we have to change a few things before we can merge. For example, I think the Eip712/Permit/Erc20 relation is quite awkward. Let's try to improve it before merging it. Otherwise this is looking very good!

contracts/src/token/erc20/extensions/permit.rs
qalisander: Seems we want to let user define his own `NAME` and `VERSION` constants. Those constants are used inside other functions of `IEip712` trait.
The other approach could be: have those fields as strings inside contract storage as `Metadata` extension. And init them through `koba` solidity constructor.

examples/erc20/src/lib.rs
qalisander: I think @bidzyyys has built an alternative to generically typed contract with `PhantomData<..>` inside contract's storage. But technically same, but looks a bit better.

contracts/src/token/erc20/extensions/permit.rs
qalisander: Basically it looks similar to [Erc721Enumerable::_add_token_to_owner_enumeration](https://github.com/OpenZeppelin/rust-contracts-stylus/blob/22ea8d6606137809f880deb50f2a0cd76f720c05/contracts/src/token/erc721/extensions/enumerable.rs#L157). Since we accepted this approach once.. 
But yeah. Tough to say what should be better for users. Since stylus-sdk is quite limited for smart contract's extensions.

qalisander (commented):
Good job @bidzyyys!
Added comments for the hot topics.

contracts/src/token/erc20/extensions/permit.rs
bidzyyys: Ok.

contracts/src/token/erc20/extensions/permit.rs
bidzyyys: Resolved with #235

examples/erc20/tests/erc20.rs
bidzyyys: Unfortunately we need to wait for #179 to have same `alloy` version...

examples/erc20/tests/erc20.rs
bidzyyys: Yep, we require `v` to be either `27` or `28` following behaviour from Solidity -> https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/test/utils/cryptography/ECDSA.test.js#L108
With `signature.v().to_u64() as u8` it can be also `0` or `1`.

contracts/src/token/erc20/mod.rs
bidzyyys: Resolved with #235

contracts/src/token/erc20/extensions/permit.rs
bidzyyys: Yep, I want to let developers define own `NAME` and `VERSION` while using our `Permit`.

contracts/src/token/erc20/extensions/permit.rs
bidzyyys:
```
error[E0277]: the trait bound `T: stylus_sdk::prelude::StorageType` is not satisfied
  --> contracts/src/token/erc20/extensions/permit.rs:57:1
   |
57 | / sol_storage! {
58 | |     /// State of a Permit Contract.
59 | |     pub struct Erc20Permit<T: IEip712>{
60 | |         /// ERC-20 contract.
...  |
68 | |     }
69 | | }
   | |_^ the trait `stylus_sdk::prelude::StorageType` is not implemented for `T`
```

examples/erc20/tests/erc20.rs
bidzyyys: I do not see any strong reason for this... Besides that, we would need to query blockchain that makes tests more complicated.

examples/erc20/src/lib.rs
bidzyyys: I just wanted to let developers define own `NAME` and `VERSION` and do not require from them implementing whole `Permit` logic.

examples/erc20/tests/erc20.rs
alexfertel: Oh, I see, I didn't expect that. Good one 👍

examples/erc20/tests/erc20.rs
alexfertel: yeah, 😢

contracts/src/token/erc20/extensions/permit.rs
```diff
@@ -0,0 +1,328 @@
...
+
+        let hash: B256 = self.eip712.hash_typed_data_v4(struct_hash);
+
+        let signer: Address = ecdsa::recover(self, hash, v, r, s)?;
+
+        if signer != owner {
+            return Err(ERC2612InvalidSigner { signer, owner }.into());
+        }
+
+        let _ = self.erc20._approve(owner, spender, value)?;
```
alexfertel, line 190: I think this wouldn't warn anymore, right?

```suggestion
        self.erc20._approve(owner, spender, value)?;
```

alexfertel (approved):
Looks good to me! 🚀 

Only a smol comment.

qalisander (approved):
LGTM
Good job!

PR #229, opened 2024-08-02
build: pin stylus-sdk version

alexfertel:
Need to pin so that new installs don't use broken versions.


bidzyyys (approved):
LGTM!

qalisander (approved):
Looks good

PR #235, opened 2024-08-08
ref: rework ERC-20 Permit extension

bidzyyys:
Rework of ERC-20 Permit extension.


contracts/src/token/erc20/extensions/permit.rs
```diff
@@ -57,8 +57,12 @@ pub enum Error {
 sol_storage! {
     /// State of a Permit Contract.
     pub struct Permit<T: IEip712 + StorageType>{
```
qalisander, line 59: I think that the name in this case should be like `Erc20Permit` since rn it is tightly coupled to `Erc20` contract.

qalisander (commented):
Good job working on PoC!
It looks similar to an approach with Consecutive, but doesn't have that much code duplication. I was expecting worse:)

PR #248, opened 2024-08-19
feat: rename crypto crate

alexfertel:
Decided to go for `openzeppelin-crypto-rs`, but obv open to something different.

UPDATE: In the end went for `openzeppelin-crypto`.

### DIFF
--- a/lib/crypto/Cargo.toml
+++ b/lib/crypto/Cargo.toml
@@ -1,5 +1,5 @@
 [package]
-name = "crypto"
+name = "openzeppelin-crypto-rs"
 categories = ["cryptography", "algorithms", "no-std", "wasm"]
 description = "Cryptography Utilities"
 edition.workspace = true
--- a/lib/crypto/Cargo.toml
+++ b/lib/crypto/Cargo.toml
@@ -1,5 +1,5 @@
 [package]
-name = "openzeppelin-crypto-rs"
+name = "openzeppelin-crypto"
 categories = ["cryptography", "algorithms", "no-std", "wasm"]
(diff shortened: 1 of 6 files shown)


lib/crypto/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "crypto"
+name = "openzeppelin-crypto-rs"
```
qalisander, line 2: Isn't it contradicting to the [official guidelines](https://rust-lang.github.io/api-guidelines/naming.html)?

> Crate names should not use -rs or -rust as a suffix or prefix. Every crate is Rust! It serves no purpose to remind users of this constantly.

qalisander: Of course there are many crates that doesn't follow that convention. But seems it is an official recomendation

alexfertel: Yes! I'll remove the suffix, was thinking something else. thx

bidzyyys: I am not a fun of this suffix; I vote for just  `openzeppelin-crypto`.

bidzyyys (commented):
I would leave `openzeppelin-crypto`

lib/crypto/Cargo.toml
qalisander: `openzeppelin-crypto` yeah. This one seems good also for me

bidzyyys (approved):
LGTM!

ggonzalez94 (approved):

PR #250, opened 2024-08-20
fix(ci): cargo-stylus and benchmarks should not share cache

### LINKED ISSUE #249: [Bug]: cargo stylus is building again in ci
### What happened?

When pr opened. Cargo stylus build step runs every time we push to the branch. And take about 6 min. But logically should be cached.
image

### platform

- [ ] linux
- [ ] windows
- [ ] macos

### Expected behavior

_No response_

### Contribution Guidelines

qalisander:
Cache key in [Swatinem/rust-cache@v2](https://github.com/Swatinem/rust-cache/tree/v2#readme) action generated by gh job_id and hashes of Cargo.toml/Cargo.lock and other files.
But we have two gh actions with the same job name `required`: e2e-tests.yml and gas-bench.yml.
And recent caches looks like [this](https://github.com/OpenZeppelin/rust-contracts-stylus/actions/caches).
image
I guess two jobs reuse the same cache key. And one that was finished earlier will update the cache entry.
We should make cache keys different for those jobs.
Also made caching less aggressive since failed(suspened) build can corrupt cache or occupy entry with not fully cached build data.

Resolves #249


alexfertel (approved):
LGTM!

Makes sense, nice catch! 👍

bidzyyys (approved):
great catch!

PR #251, opened 2024-08-20
ref: add associated error types for contract traits

qalisander:
Allows to the library users to reuse our traits with their own error types.
And it is more "rusty" approach having associated error types for traits with fallible functions.
Adds new functional macro for the `MethodError` trait.

### DIFF
--- a/contracts/src/token/erc20/extensions/permit.rs
+++ b/contracts/src/token/erc20/extensions/permit.rs
@@ -12,7 +12,9 @@
 use alloy_primitives::{b256, keccak256, Address, B256, U256};
 use alloy_sol_types::{sol, SolType};
 use stylus_proc::{external, sol_storage, SolidityError};
-use stylus_sdk::{block, prelude::StorageType, storage::TopLevelStorage};
+use stylus_sdk::{
+    block, call::MethodError, prelude::StorageType, storage::TopLevelStorage,
+};
 
 use crate::{
     token::erc20::{self, Erc20, IErc20},
--- a/contracts/src/token/erc20/extensions/permit.rs
+++ b/contracts/src/token/erc20/extensions/permit.rs
@@ -12,9 +12,7 @@
 use alloy_primitives::{b256, keccak256, Address, B256, U256};
 use alloy_sol_types::{sol, SolType};
 use stylus_proc::{external, sol_storage, SolidityError};
-use stylus_sdk::{
-    block, call::MethodError, prelude::StorageType, storage::TopLevelStorage,
-};
+use stylus_sdk::{block, prelude::StorageType, storage::TopLevelStorage};
 
 use crate::{
     token::erc20::{self, Erc20, IErc20},
--- a/contracts/src/token/erc20/extensions/permit.rs
+++ b/contracts/src/token/erc20/extensions/permit.rs
@@ -60,18 +60,6 @@ pub enum Error {
     ECDSA(ecdsa::Error),
 }
 
-impl MethodError for erc20::Error {
-    fn encode(self) -> alloc::vec::Vec<u8> {
-        self.into()
-    }
-}
-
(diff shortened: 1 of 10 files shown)


contracts/src/token/erc20/extensions/permit.rs
```diff
@@ -60,18 +60,6 @@ pub enum Error {
     ECDSA(ecdsa::Error),
 }
 
-impl MethodError for erc20::Error {
```
qalisander, line 63: `MethodError` implementation is moved to the corresponding contract

contracts/src/utils/structs/checkpoints.rs
```diff
@@ -29,6 +32,12 @@ pub enum Error {
     CheckpointUnorderedInsertion(CheckpointUnorderedInsertion),
 }
 
+impl MethodError for Error {
```
bidzyyys, line 35: Can we have #[derive(MethodError)]` macro?

bidzyyys (commented):
Looks good but I do not like having `impl MethodError` many times...

contracts/src/utils/structs/checkpoints.rs
qalisander: Also don't like it. Basically I have proposed it in [pr to stylus-sdk](https://github.com/OffchainLabs/stylus-sdk-rs/pull/136). I hope they will do smth with it:)

qalisander: Also we don't have a proc macro library for contracts. Do you think we should create it for this temporary macro?

qalisander: Instead of separate crate for `derive` macro. I've added functional macro that looks like this: `impl_method_error!(Error);`. Wdt @bidzyyys?

bidzyyys (approved):
Looks good!

contracts/src/token/erc20/extensions/permit.rs
alexfertel: This means that this import can be removed from this file, right?

contracts/src/token/erc721/extensions/consecutive.rs
```diff
@@ -130,18 +130,6 @@ pub enum Error {
 
 unsafe impl TopLevelStorage for Erc721Consecutive {}
 
-impl MethodError for erc721::Error {
```
alexfertel, line 133: Same as above

contracts/src/token/erc721/mod.rs
```diff
@@ -150,6 +153,8 @@ pub enum Error {
     InvalidOperator(ERC721InvalidOperator),
 }
 
+impl_method_error!(Error);
```
alexfertel, line 156: I prefer we write it here explicitly, since this should be something that the SDK exposes as a `derive` or by default. It makes it easier to change later.

contracts/src/token/erc721/mod.rs
```diff
@@ -193,8 +198,9 @@ unsafe impl TopLevelStorage for Erc721 {}
 
 /// Required interface of an [`Erc721`] compliant contract.
 pub trait IErc721 {
-    /// The error type associated to this ERC-721 trait implementation.
+    /// The error type associated with ERC-721 contract.
```
alexfertel, line 201: I think what we had here before was better. Same with the other comments.

P.s. this should be `with the ERC-721 contract` if we keep it.

alexfertel (changes requested):
If I understand correctly this change, I would expect it to affect our examples, right? What is the practical result of this change?

contracts/src/utils/structs/checkpoints.rs
qalisander: decided with @alexfertel to write it expicitly

PR #253, opened 2024-08-21
ci: bump cargo-stylus version

qalisander:


bidzyyys:
We need to make sure that every build is run on the same version; Would be great to have a single place to change `cargo-stylus` version.

qalisander:
@bidzyyys how do you think implementation should look like?

bidzyyys (approved):
LGTM! Let's leave as it is right now.

PR #254, opened 2024-08-22
feat: cache contracts before benching

alexfertel:
Use the `cargo stylus cache` to cache contracts in the `CacheManager` before running benches.

### DIFF
--- a/benches/src/access_control.rs
+++ b/benches/src/access_control.rs
@@ -50,7 +50,9 @@ pub async fn bench() -> eyre::Result<Report> {
         .wallet(EthereumWallet::from(bob.signer.clone()))
         .on_http(bob.url().parse()?);
 
-    let contract_addr = deploy(&alice).await;
+    let contract_addr = deploy(&alice).await?;
+    crate::cache_contract(&alice, contract_addr)?;
+
     let contract = AccessControl::new(contract_addr, &alice_wallet);
     let contract_bob = AccessControl::new(contract_addr, &bob_wallet);
 
@@ -72,7 +74,7 @@ pub async fn bench() -> eyre::Result<Report> {
     Ok(report)
 }
 
-async fn deploy(account: &Account) -> Address {
+async fn deploy(account: &Account) -> eyre::Result<Address> {
     let args = AccessControl::constructorCall {};
     let args = alloy::hex::encode(args.abi_encode());
     crate::deploy(account, "access-control", Some(args)).await
--- a/benches/src/access_control.rs
+++ b/benches/src/access_control.rs
@@ -8,7 +8,10 @@ use alloy::{
 };
 use e2e::{receipt, Account};
 
-use crate::report::Report;
+use crate::{
+    report::{ContractReport, FunctionReport},
+    CacheOpt,
+};
 
 sol!(
     #[sol(rpc)]
@@ -33,7 +36,22 @@ const ROLE: [u8; 32] =
 const NEW_ADMIN_ROLE: [u8; 32] =
     hex!("879ce0d4bfd332649ca3552efe772a38d64a315eb70ab69689fd309c735946b5");
 
-pub async fn bench() -> eyre::Result<Report> {
+pub async fn bench() -> eyre::Result<ContractReport> {
+    let receipts = run_with(CacheOpt::None).await?;
+    let report = receipts
+        .into_iter()
+        .try_fold(ContractReport::new("AccessControl"), ContractReport::add)?;
+
+    let cached_receipts = run_with(CacheOpt::Bid(0)).await?;
+    let report = cached_receipts
+        .into_iter()
+        .try_fold(report, ContractReport::add_cached)?;
+
+    Ok(report)
+}
+pub async fn run_with(
+    cache_opt: CacheOpt,
+) -> eyre::Result<Vec<FunctionReport>> {
     let alice = Account::new().await?;
     let alice_addr = alice.address();
     let alice_wallet = ProviderBuilder::new()
@@ -50,8 +68,7 @@ pub async fn bench() -> eyre::Result<Report> {
         .wallet(EthereumWallet::from(bob.signer.clone()))
         .on_http(bob.url().parse()?);
 
-    let contract_addr = deploy(&alice).await?;
-    crate::cache_contract(&alice, contract_addr)?;
+    let contract_addr = deploy(&alice, cache_opt).await?;
 
     let contract = AccessControl::new(contract_addr, &alice_wallet);
     let contract_bob = AccessControl::new(contract_addr, &bob_wallet);
@@ -68,14 +85,17 @@ pub async fn bench() -> eyre::Result<Report> {
         (setRoleAdminCall::SIGNATURE, receipt!(contract.setRoleAdmin(ROLE.into(), NEW_ADMIN_ROLE.into()))?),
     ];
 
-    let report = receipts
+    receipts
         .into_iter()
-        .try_fold(Report::new("AccessControl"), Report::add)?;
-    Ok(report)
+        .map(FunctionReport::new)
+        .collect::<eyre::Result<Vec<_>>>()
 }
 
-async fn deploy(account: &Account) -> eyre::Result<Address> {
+async fn deploy(
+    account: &Account,
+    cache_opt: CacheOpt,
+) -> eyre::Result<Address> {
     let args = AccessControl::constructorCall {};
     let args = alloy::hex::encode(args.abi_encode());
-    crate::deploy(account, "access-control", Some(args)).await
+    crate::deploy(account, "access-control", Some(args), cache_opt).await
 }
--- a/benches/src/main.rs
+++ b/benches/src/main.rs
@@ -13,6 +13,7 @@ async fn main() -> eyre::Result<()> {
     let report =
         reports.into_iter().fold(Reports::default(), Reports::merge_with);
 
+    println!();
     println!("{report}");
 
     Ok(())
--- a/benches/src/main.rs
+++ b/benches/src/main.rs
@@ -1,20 +1,19 @@
 use benches::{
     access_control, erc20, erc721, merkle_proofs, report::BenchmarkReport,
 };
+use futures::FutureExt;
 
 #[tokio::main]
 async fn main() -> eyre::Result<()> {
-    let reports = tokio::join!(
-        access_control::bench(),
-        erc20::bench(),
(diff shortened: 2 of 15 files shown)


bidzyyys (approved):
Looks good

alexfertel:
> Looks good

Doesn't work 😢

benches/src/access_control.rs
```diff
@@ -50,7 +50,9 @@ pub async fn bench() -> eyre::Result<Report> {
         .wallet(EthereumWallet::from(bob.signer.clone()))
         .on_http(bob.url().parse()?);
 
-    let contract_addr = deploy(&alice).await;
+    let contract_addr = deploy(&alice).await?;
+    crate::cache_contract(&alice, contract_addr)?;
```
alexfertel, line 54: Maybe this can be part of `deploy`? And maybe change `deploy`'s signature to receive a `enum CachePolicy { None, Bid(u64) }` which determines whether we cache or not? (this may seem overengineered, but it is forward-compatible)

benches/src/main.rs
```diff
@@ -13,6 +13,7 @@ async fn main() -> eyre::Result<()> {
     let report =
         reports.into_iter().fold(Reports::default(), Reports::merge_with);
 
+    println!();
     println!("{report}");
```
alexfertel, line 17:
```suggestion
    println!("\n{report}");
```

lib/e2e/src/account.rs
```diff
@@ -23,7 +25,7 @@ pub struct Account {
 }
 
 impl Account {
-    /// Create a new account.
+    /// Create a new account with default funding of 100 ETH.
```
alexfertel, line 28:
```suggestion
    /// Create a new account with a default funding of `DEFAULT_FUNDING_ETH`.
```

scripts/bench.sh
```diff
@@ -5,9 +5,14 @@ MYDIR=$(realpath "$(dirname "$0")")
 cd "$MYDIR"
 cd ..
 
-NIGHTLY_TOOLCHAIN=${NIGHTLY_TOOLCHAIN:-nightly}
+NIGHTLY_TOOLCHAIN=${NIGHTLY_TOOLCHAIN:-nightly-2024-01-01}
```
alexfertel, line 8: Maybe use a more recent nightly like `2024-09-01`?

scripts/bench.sh
```diff
@@ -5,9 +5,14 @@ MYDIR=$(realpath "$(dirname "$0")")
...
 cargo +"$NIGHTLY_TOOLCHAIN" build --release --target wasm32-unknown-unknown -Z build-std=std,panic_abort -Z build-std-features=panic_immediate_abort
 
 export RPC_URL=http://localhost:8547
-cargo run --release -p benches
+
+# No need to run benchmark infrastructure with release.
+# Just compilation will take longer.
+# Contracts already built with release.
+cargo run -p benches
```
alexfertel, line 16: If the contracts are already built with release, then this would skip compilation as well? I think the comment is a bit misleading, could we rephrase it?

If the reason is that `--release` has no impact here because only compiling the contracts with `--release` is what matters, then I think something like "No need to compile benchmarks with `--release` since gas measurement happens on the contract side".

scripts/nitro-testnode.sh
```diff
@@ -18,7 +18,14 @@ do
       shift
       ;;
     -q|--quit)
-      docker container stop $(docker container ls -q --filter name=nitro-testnode)
+      NITRO_CONTAINERS=$(docker container ls -q --filter name=nitro-testnode)
+
+      if [ -z "$NITRO_CONTAINERS" ]; then
+          echo "No running nitro test node containers"
```
alexfertel, line 24:
```suggestion
          echo "No nitro-testnode containers running"
```

alexfertel (commented):
It's great to see that this now works!

At first glance it seems like a 20k reduction in gas usage? Nice!

I can't approve cause I opened the PR lol

alexfertel:
I also think that panicking is fine here, since this is not user-facing, but not a strong opinion!

scripts/bench.sh
qalisander: Alas, recent nightly version do not work. That is another problem why we're pinned. Also I've checked that on stable not just binary size is smaller, but benchmarks perform %2-3 worse

lib/e2e/src/account.rs
qalisander: make sense!

scripts/bench.sh
qalisander: Yeah. You've got it correct. Will make it clear

qalisander:
@alexfertel thanks for the review! It seems for me that caching doesn't make that big difference. But still

ggonzalez94:
@alexfertel this looks great! Thanks for continuing the work on this.
One thing is that I believe we should have benchmarks for both cached and uncached runs of the contracts(ideally on a table that compares them side by side).

Can you add this to the PR? If you are busy we can take over and add it ourselves.

qalisander:
@ggonzalez94 yeah. I've been planning to do it. Check out gas benchmark report now

scripts/bench.sh
```diff
@@ -5,9 +5,16 @@ MYDIR=$(realpath "$(dirname "$0")")
...
 
-NIGHTLY_TOOLCHAIN=${NIGHTLY_TOOLCHAIN:-nightly}
+NIGHTLY_TOOLCHAIN=${NIGHTLY_TOOLCHAIN:-nightly-2024-01-01}
 cargo +"$NIGHTLY_TOOLCHAIN" build --release --target wasm32-unknown-unknown -Z build-std=std,panic_abort -Z build-std-features=panic_immediate_abort
 
 export RPC_URL=http://localhost:8547
-cargo run --release -p benches
+
+# No need to compile benchmarks with `--release`
+# since gas measurement happens on the contract side
```
ggonzalez94, line 14: I think this comment can be a bit clearer. Not sure if it means that the gas happens from outisde the node or what?

benches/src/main.rs
```diff
@@ -10,9 +12,11 @@ async fn main() -> eyre::Result<()> {
     );
 
     let reports = [reports.0?, reports.1?, reports.2?, reports.3?];
```
ggonzalez94, line 14: This is a nit, but I don't love the idea of accessing the values of the tuple by index(hard to scale and avoid missing out some of the values). Even if it is more verbose this is better IMO and would allow the compiler to highlight when there's an unused value.
```
    let (access_control_report, erc20_report, erc721_report, merkle_proofs_report) = tokio::join!(
        access_control::bench(),
        erc20::bench(),
        erc721::bench(),
        merkle_proofs::bench()
    );

    let reports = [access_control_report?, erc20_report?, erc721_report?, merkle_proofs_report?];
```

scripts/nitro-testnode.sh
```diff
@@ -41,10 +48,10 @@ then
   cd "$MYDIR" || exit
   cd ..
 
-  git clone --recurse-submodules https://github.com/OffchainLabs/nitro-testnode.git
+  git clone -b release --recurse-submodules https://github.com/OffchainLabs/nitro-testnode.git
   cd ./nitro-testnode || exit
-  # `release` branch.
-  git checkout 8cb6b84e31909157d431e7e4af9fb83799443e00 || exit
+  git pull origin release --recurse-submodules
+  git checkout 43861b001713db3526b74e905e383d41e9272760 || exit
```
ggonzalez94, line 54: Not important.... But do we need to clone the release branch if we are doing a checkout on a specific commit? Why don't we just checkout and run from release branch? Is it because we are afraid of breaking changes?(I still think that it is probably better find out right away)

ggonzalez94 (approved):
Looks awesome! Just a few nits and a questions. But looks ready to go 🚀

scripts/bench.sh
qalisander: The comment was like that [before](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/254#discussion_r1771079049). @alexfertel offered to slightly improve it

benches/src/main.rs
qalisander: If you're curious, there is a functional and a bit overengineered solution:
```rust
    let report = futures::future::try_join_all([
        access_control::bench().boxed(),
        erc20::bench().boxed(),
        erc721::bench().boxed(),
        merkle_proofs::bench().boxed(),
    ])
    .await?
    .into_iter()
    .fold(BenchmarkReport::default(), BenchmarkReport::merge_with);
```
I can change to this

scripts/bench.sh
ggonzalez94: If I understand correctly the reason we don't need the `--release`flag is because this is running the benchmark infrastructure and not the smart contracts that are being measured? If that's the case may I offer an improved wording?
```suggestion
# since this only runs the benchmarking code and the contracts have already been compiled with `--release`
```

benches/src/main.rs
ggonzalez94: I like it! :rocket:

scripts/nitro-testnode.sh
ggonzalez94: After our talk today do you want to just run the latest from the release branch or prefer to keep the specific commit? Regardless this looks ready to go!

scripts/bench.sh
alexfertel: beautiful! I'm so bad with words 😅

scripts/nitro-testnode.sh
qalisander: I've just found one downside of this approach applicable for benchmarks. You can check out how gas fees have changed after I've updated the nitro test node with a bug fix. Improvement is significant. And since I've just lifted the nitro-test-node's version it is obvious what is the reason.
But in future when we will see some random uplift/downgrade of benchmarks, we can whether blame our library changes or hidden changes at nitro test node.
I think it is better to keep the commit hash for now. And in future record in benchmark's report metadata about nitro-test-node revision. And use just a release branch. That will help us to analyze why performance improved/degraded
Wdt?

qalisander:
@ggonzalez94 @alexfertel It happened! Check out gas benchmarks now:) Just a small bug with pricing was fixed inside nitro node

alexfertel:
oooooh, they look much closer to the Solidity ones now! 🚀 

Interesting that the merkle proof stayed the same. Overall looks great

qalisander:
Merkle proof doesn't write to storage. The bug was connected with charging fees multiple times for on-chain storage access

PR #257, opened 2024-08-23
ref: code cleanup

bidzyyys:
Cleanup of existing warnings.


qalisander (approved):
LGTM

contracts/src/utils/structs/checkpoints.rs
```diff
@@ -16,7 +16,9 @@ use crate::utils::math::alloy::Math;
 
 // TODO: add generics for other pairs (uint32, uint224) and (uint48, uint208).
 // Logic should be the same.
+/// "[`Uint`] for 96 bits."
```
qalisander, line 19: Why we need double quotes (") here?

bidzyyys: Good catch!

PR #273, opened 2024-09-02
build: bump Stylus SDK to v0.6.0 and cargo-stylus to v0.5.1

bidzyyys:
Bump:
- Stylus SDK to `v0.6.0`
- cargo-stylus to `v0.5.1`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -60,11 +60,11 @@ all = "warn"
 
 [workspace.dependencies]
 # stylus-related
-stylus-sdk = { version = "=0.5.0", default-features = false }
-stylus-proc = { version = "=0.5.0", default-features = false }
+stylus-sdk = { version = "=0.6.0", default-features = false }
+stylus-proc = { version = "=0.6.0", default-features = false }
 mini-alloc = "0.4.2"
 
-alloy = { version = "0.1.1", features = [
+alloy = { version = "0.1.4", features = [
   "contract",
   "network",
   "providers",
@@ -77,8 +77,8 @@ alloy = { version = "0.1.1", features = [
 # Even though `alloy` includes `alloy-primitives` and `alloy-sol-types` we need
 # to keep both versions for compatibility with the Stylus SDK. Once they start
 # using `alloy` we can remove these.
-alloy-primitives = { version = "0.3.1", default-features = false }
-alloy-sol-types = { version = "0.3.1", default-features = false }
+alloy-primitives = { version = "0.7.6", default-features = false }
+alloy-sol-types = { version = "0.7.6", default-features = false }
 
 const-hex = { version = "1.11.1", default-features = false }
 eyre = "0.6.8"
--- a/contracts/src/token/erc721/extensions/consecutive.rs
+++ b/contracts/src/token/erc721/extensions/consecutive.rs
@@ -26,7 +26,7 @@ use alloc::vec;
 
 use alloy_primitives::{uint, Address, U256};
 use alloy_sol_types::sol;
-use stylus_proc::{external, sol_storage, SolidityError};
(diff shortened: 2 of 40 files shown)


qalisander (approved):
LGTM

contracts/src/token/erc721/mod.rs
```diff
@@ -1112,8 +1112,13 @@ impl Erc721 {
...
         let call = Call::new_in(self);
         let data = data.to_vec();
-        let result =
-            receiver.on_erc_721_received(call, operator, from, token_id, data);
+        let result = receiver.on_erc_721_received(
+            call,
+            operator,
+            from,
+            token_id,
+            data.to_vec().into(),
```
ggonzalez94, line 1120: I think we are doing an unnecessary conversion here. The function takes in `data` as Bytes, then converts it to a vec and now converts it back to Bytes. I think we can remove line 1114 and just pass in data here

contracts/src/token/erc721/extensions/consecutive.rs
```diff
@@ -168,7 +168,7 @@ impl IErc721 for Erc721Consecutive {
             from,
             to,
             token_id,
-            &data,
+            data,
```
ggonzalez94, line 171: This is more a question than a comment. I'm curious what happened here. I see that `IERC721Receiver` always took bytes as a parameter, but we were sending the borrowed variable through the codebase.

ggonzalez94 (commented):
Looks great overall! I left a couple comments and one question.

contracts/src/token/erc721/extensions/consecutive.rs
qalisander: Since `Bytes` type just wraps Vec<u8>, that is referential type by itself. We are not copying underlying bytes itself passing this type by value. I would say it is more simple to have it without reference if we can do it

contracts/src/token/erc721/mod.rs
qalisander: Alas, those are different `Bytes` types. One is from `alloy-primitives` and the other one from `stylus-sdk`

qalisander: Yeah, you correct @ggonzalez94. I've noticed the lane `let data = data.to_vec();`. `to_vec` method call is duplicated

Cargo.toml
```diff
@@ -77,8 +77,8 @@ alloy = { version = "0.1.1", features = [
 # Even though `alloy` includes `alloy-primitives` and `alloy-sol-types` we need
 # to keep both versions for compatibility with the Stylus SDK. Once they start
 # using `alloy` we can remove these.
-alloy-primitives = { version = "0.3.1", default-features = false }
-alloy-sol-types = { version = "0.3.1", default-features = false }
+alloy-primitives = { version = "0.7.6", default-features = false }
```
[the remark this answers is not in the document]
qalisander: Initially it fixed a bug with serialization. Now as I've checked `e2e-tests` work without `tiny-keccak` feature. Removed it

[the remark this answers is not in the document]
bidzyyys: Yep, it was needed in `hotfix-v0.5.3`, great catch @ggonzalez94 & thanks for checking this @qalisander

contracts/src/token/erc721/mod.rs
ggonzalez94: I see. Should we remove it an merge the PR? I think it is the only thing missing @qalisander @bidzyyys

bidzyyys: Adding this right now.

qalisander (approved):
Good job!
Ship it

PR #277, opened 2024-09-06
docs: e2e tests setup guide

qalisander:
Extend guidelines for e2e tests and adjust ux of few scripts.

### DIFF
--- a/GUIDELINES.md
+++ b/GUIDELINES.md
@@ -226,7 +226,7 @@ Also, you should have the cargo stylus tool:
 cargo install cargo-stylus
 ```
 
-Since most of the e2e tests using koba for deploying contracts, you need to install solidity compiler v0.8.21, like
+Since most of the e2e tests use [koba](https://github.com/OpenZeppelin/koba) for deploying contracts, you need to install the solidity compiler (`v0.8.21`). You can do so, following these steps:
 this:
 
 ```shell
--- a/GUIDELINES.md
+++ b/GUIDELINES.md
@@ -235,7 +235,7 @@ sudo mv solc-static-linux /usr/bin/solc
 sudo chmod a+x /usr/bin/solc
 ```
 
-To run e2e tests, you need to have a local nitro test node set up.
+To run e2e tests, you need to have a local nitro test node up and running.
 Run the following command and wait till script exit successfully:
 
 ```shell
--- a/GUIDELINES.md
+++ b/GUIDELINES.md
@@ -131,7 +131,7 @@ Use 1-word names for function parameters or variables that have larger scopes:
 
 ```rust
 pub fn commit(repo: &Repository, sig: &Signature) -> Result<Commit, Error> {
-...
+    ...
 }
 ```
 
@@ -207,6 +207,47 @@ Make sure all tests are passing with:
 
     $ cargo test --all-features
 
+### Running end-to-end tests
+
+In order to run end-to-end (e2e) tests you need to have a specific nightly toolchain.
+"Nightly" is necessary to use optimization compiler flags and have contract wasm small enough to be eligible for
+deployment.
+
+Run the following commands to install the necessary toolchain:
+
+```shell
+rustup install nightly-2024-01-01
+rustup component add rust-src
+```
+
+Also, you should have the cargo stylus tool:
+
+```shell
+cargo install cargo-stylus
+```
+
+Since most of the e2e tests using koba for deploying contracts, you need to install solidity compiler v0.8.21, like
+this:
+
+```shell
+curl -LO https://github.com/ethereum/solidity/releases/download/v0.8.21/solc-static-linux
+sudo mv solc-static-linux /usr/bin/solc
+sudo chmod a+x /usr/bin/solc
+```
+
+To run e2e tests, you need to have a local nitro test node set up.
+Run the following command and wait till script exit successfully:
+
+```shell
+./scripts/nitro-testnode.sh -i -d
+```
+
+Then you will be able to run e2e tests:
+
+```shell
+./scripts/e2e-tests.sh
+```
+
 ### Checking the docs
 
 If you make documentation changes, you may want to check whether there are any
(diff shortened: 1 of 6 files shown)


GUIDELINES.md
```diff
@@ -207,6 +207,47 @@ Make sure all tests are passing with:
...
+rustup component add rust-src
+```
+
+Also, you should have the cargo stylus tool:
+
+```shell
+cargo install cargo-stylus
+```
+
+Since most of the e2e tests using koba for deploying contracts, you need to install solidity compiler v0.8.21, like
```
ggonzalez94, line 229:
```suggestion
Since most of the e2e tests use [koba](https://github.com/OpenZeppelin/koba) for deploying contracts, you need to install the solidity compiler (`v0.8.21`). You can do so, following these steps:
```

ggonzalez94: Do people need that exact version of the compiler? Or are we of with `0.8.21`+?

GUIDELINES.md
```diff
@@ -207,6 +207,47 @@ Make sure all tests are passing with:
...
+this:
+
+```shell
+curl -LO https://github.com/ethereum/solidity/releases/download/v0.8.21/solc-static-linux
+sudo mv solc-static-linux /usr/bin/solc
+sudo chmod a+x /usr/bin/solc
```
ggonzalez94, line 235: Isn't it easier if people use npm? 
`npm install -g solc`

GUIDELINES.md
```diff
@@ -207,6 +207,47 @@ Make sure all tests are passing with:
...
+```
+
+To run e2e tests, you need to have a local nitro test node set up.
```
ggonzalez94, line 238:
```suggestion
To run e2e tests, you need to have a local nitro test node up and running.
```

ggonzalez94 (commented):
Looks good! Just left a few suggestion to improve the guidelines

GUIDELINES.md
qalisander: I remember that koba doesn't work with latest solidity version

GUIDELINES.md
```diff
@@ -207,6 +207,46 @@ Make sure all tests are passing with:
...
+
+```shell
+rustup install nightly-2024-01-01
+rustup component add rust-src
+```
+
+Also, you should have the cargo stylus tool:
+
+```shell
+cargo install cargo-stylus
```
bidzyyys, line 226: You should specify version of `cargo-stylus`

GUIDELINES.md
```diff
@@ -207,6 +207,46 @@ Make sure all tests are passing with:
...
+```
+
+Since most of the e2e tests use [koba](https://github.com/OpenZeppelin/koba) for deploying contracts, you need to
+install the solidity compiler (`v0.8.21`). You can do so, following these steps:
+this:
+
+```shell
+npm install -g solc@0.8.21
```
bidzyyys, line 234: IMHO missing information how to install `npm`

scripts/nitro-testnode.sh
```diff
@@ -46,7 +46,7 @@ then
   # `release` branch.
   git checkout 8cb6b84e31909157d431e7e4af9fb83799443e00 || exit
 
-  ./test-node.bash --no-run --init --no-tokenbridge || exit
+  ./test-node.bash --no-run --init || exit
```
bidzyyys, line 49: Why do we need `tokenbridge`?

scripts/nitro-testnode.sh
```diff
@@ -17,15 +17,15 @@ do
       HAS_DETACH=true
       shift
       ;;
-    -down|--shutdown)
-      docker container stop "$(docker container ls -q --filter name=nitro-testnode)"
+    -q|--quit)
```
bidzyyys, line 20: I am not a fan of `-q/--quit`; IMHO `-down` is better.
Even in `docker compose` you have `down` instead of `quit`.

bidzyyys (commented):
Left some comments.

GUIDELINES.md
ggonzalez94: True.... but I don't think we should teach people how to install every tool as part of this guide. I'm happy with just specifying it is a prerequisite. But @qalisander can decide what works best.

ggonzalez94 (approved):
Thanks for adding this!

scripts/nitro-testnode.sh
qalisander: In this version of the test node `--no-tokenbridge` behaviour is by default.
Before it was different

GUIDELINES.md
qalisander: I'd like to have fewer places with version indication. Latest cargo-stylus usually works.

scripts/nitro-testnode.sh
qalisander: in docker compose `down` is not a flag. It is a subcommand and looks like this: `docker compose up` and `docker compose down`.
We should pass to this script short and long flags. But `-down` flag is already quite long:)

GUIDELINES.md
qalisander: It was an idea of @ggonzalez94 to use npm. I would prefer to have a straightforward installation guideline. And if user want to use `npm`, `yarn`, `brew` or `apt-get` he will decide. Wdt guys?

ggonzalez94: I'm in favor of npm, I just think that people can figure out how to install it themselves(or we can put a link to the official docs). But I'm happy just listing it a prerequisite :)

ggonzalez94 (approved):

scripts/nitro-testnode.sh
bidzyyys: Fair enough, thanks!

GUIDELINES.md
bidzyyys: `npm` is fine for me, I think that we need to mention that it is a prerequisite and link to the official docs.

GUIDELINES.md
bidzyyys: Fair enough.

GUIDELINES.md
bidzyyys: IMHO the best way is just to point that we require this specific version and leave link to this - https://docs.soliditylang.org/en/latest/installing-solidity.html#

scripts/nitro-testnode.sh
bidzyyys: Ok, can agree on this.

GUIDELINES.md
alexfertel: It should work with up to `0.8.24` because that version introduces the `MCOPY` opcode, and `nitro-testnode` didn't support that at the time. [Issue](https://github.com/OpenZeppelin/koba/issues/4) from the koba repo.

Now that it does, the fix should be pretty simple, since I had already added `MCOPY` on [this commit](https://github.com/OpenZeppelin/koba/commit/607d36aae8ae43054d7215a1df7727ad3ad5f368). Probably only need to test with a newer Solidity version and that's it. However, there may be new stuff that nitro-testnode doesn't support.

cc: @bidzyyys @ggonzalez94

GUIDELINES.md
alexfertel: wouldn't that require having `npm` installed, as opposed to `curl` (which I believe comes in every UNIX system)?

bidzyyys (approved):
Looks good to me!

GUIDELINES.md
qalisander: I would also prefer smth without external deps. But we have settled on official solc installation guide.

PR #281, opened 2024-09-11
feat: erc165 support interface

### LINKED ISSUE #33: Add support interface ERC165 to our contracts

Implement a function selector that is same as [ERC165](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/cb2aaaa04a292887c49839cd958b08a83979d746/contracts/utils/introspection/ERC165.sol#L20) for solidity.

Also think about ways to to retrieve function selectors and interface id (that is basically xor of all function selectors in the interface).
Would be great to have a convenient function selector support from stylus sdk as it is at solidity.

Helpful links:
- function_selector! [macro](https://github.com/OffchainLabs/stylus-sdk-rs/blob/290326c9f28057f20cbe881738b913ed34c20755/stylus-sdk/src/abi/mod.rs#L78).
- function [selectors](https://docs.rs/alloy-sol-types/latest/alloy_sol_types/trait.SolInterface.html#method.selectors) in the alloy crates.

Add the standard to the following contracts:
- [ ] ERC721URIStorage (#56)
- [ ] ERC721 (#28)
- [ ] ERC721Enumerable (#67)

### Contribution Guidelines

qalisander:
Adds erc165 standard altogether with a proc macro, that lets to compute interface id in a simple way.

Resolves #33

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/contracts-proc/Cargo.toml
+++ b/contracts-proc/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "openzeppelin-stylus-proc"
 description = "Procedural macros for OpenZeppelin Stylus contracts"
-version = "0.1.0"
+version.workspace = true
 authors.workspace = true
 edition.workspace = true
 license.workspace = true
--- a/contracts-proc/Cargo.toml
+++ b/contracts-proc/Cargo.toml
@@ -1,0 +1,22 @@
+[package]
+name = "openzeppelin-stylus-proc"
+description = "Procedural macros for OpenZeppelin Stylus contracts"
+version = "0.1.0"
+authors.workspace = true
+edition.workspace = true
+license.workspace = true
+keywords.workspace = true
+repository.workspace = true
+
+
+[dependencies]
+proc-macro2.workspace = true
+quote.workspace = true
+syn.workspace = true
+convert_case = "0.6.0"
+
+[lints]
+workspace = true
+
+[lib]
+proc-macro = true
--- a/contracts-proc/src/interface.rs
+++ b/contracts-proc/src/interface.rs
@@ -3,7 +3,7 @@ use std::mem;
 use convert_case::{Case, Casing};
 use proc_macro::TokenStream;
 use proc_macro2::Ident;
-use quote::{quote, ToTokens};
+use quote::quote;
 use syn::{
     parse::{Parse, ParseStream},
     parse_macro_input, FnArg, ItemTrait, LitStr, Result, Token, TraitItem,
--- a/contracts-proc/src/interface.rs
+++ b/contracts-proc/src/interface.rs
@@ -1,0 +1,118 @@
+use std::mem;
+
+use convert_case::{Case, Casing};
+use proc_macro::TokenStream;
+use proc_macro2::Ident;
+use quote::{quote, ToTokens};
+use syn::{
+    parse::{Parse, ParseStream},
+    parse_macro_input, FnArg, ItemTrait, LitStr, Result, Token, TraitItem,
+};
+
+/// TODO#q: description
+pub fn interface(_attr: TokenStream, input: TokenStream) -> TokenStream {
+    let mut input = parse_macro_input!(input as ItemTrait);
+    let mut output = quote! {};
+
+    let mut selectors = Vec::new();
+    for item in &mut input.items {
+        let TraitItem::Fn(func) = item else {
+            continue;
+        };
+
+        let mut override_name = None;
+        for attr in mem::take(&mut func.attrs) {
+            let Some(ident) = attr.path().get_ident() else {
+                func.attrs.push(attr);
+                continue;
+            };
+            if *ident == "selector" {
+                if override_name.is_some() {
+                    error!(attr.path(), "more than one selector attribute");
+                }
+                let args: SelectorArgs = match attr.parse_args() {
+                    Ok(args) => args,
+                    Err(error) => error!(ident, "{}", error),
+                };
+                override_name = Some(args.name);
+                continue;
+            }
+            func.attrs.push(attr);
+        }
+
+        let sol_name = override_name.unwrap_or_else(|| {
+            func.sig.ident.clone().to_string().to_case(Case::Camel)
+        });
+
+        let args = func.sig.inputs.iter();
+        let arg_types: Vec<_> = args
+            .filter_map(|arg| match arg {
+                FnArg::Typed(t) => Some(t.ty.clone()),
+                _ => None,
+            })
+            .collect();
+
+        let selector = quote! { u32::from_be_bytes(stylus_sdk::function_selector!(#sol_name #(, #arg_types )*)) };
+        selectors.push(selector);
+    }
+
+    let name = input.ident.clone();
+    let vis = input.vis.clone();
+    let attrs = input.attrs.clone();
+    let trait_items = input.items.clone();
+    let (_impl_generics, ty_generics, where_clause) =
+        input.generics.split_for_impl();
+
+    output.extend(quote! {
+        #(#attrs)*
+        #vis trait #name #ty_generics #where_clause {
+            #(#trait_items)*
+
+            /// Solidity interface id associated with current trait.
+            const INTERFACE_ID: u32 = {
+                #(#selectors)^*
+            };
+        }
+    });
+
+    output.into()
+}
+
+struct SelectorArgs {
+    name: String,
+}
+
+impl Parse for SelectorArgs {
+    fn parse(input: ParseStream) -> Result<Self> {
+        let mut name = None;
+
+        if input.is_empty() {
+            error!(@input.span(), "missing id or text argument");
+        }
+
+        while !input.is_empty() {
+            let ident: Ident = input.parse()?;
+            let _: Token![=] = input.parse()?;
+
+            match ident.to_string().as_str() {
+                "name" => {
+                    let lit: LitStr = input.parse()?;
+                    if name.is_some() {
+                        error!(@lit, r#"only one "name" is allowed"#);
+                    }
+                    name = Some(lit.value());
+                }
+                _ => error!(@ident, "Unknown selector attribute"),
+            }
+
+            // allow a comma
+            let _: Result<Token![,]> = input.parse();
+        }
+
+        if let Some(name) = name {
+            Ok(Self { name })
+        } else {
+            error!(@input.span(), r#""name" is required"#);
+        }
+    }
+}
--- a/contracts-proc/src/interface.rs
+++ b/contracts-proc/src/interface.rs
@@ -9,8 +9,8 @@ use syn::{
     parse_macro_input, FnArg, ItemTrait, LitStr, Result, Token, TraitItem,
 };
 
-/// TODO#q: description
-pub fn interface(_attr: TokenStream, input: TokenStream) -> TokenStream {
+/// Computes interface id as an associated constant for the trait.
+pub(crate) fn interface(_attr: TokenStream, input: TokenStream) -> TokenStream {
     let mut input = parse_macro_input!(input as ItemTrait);
     let mut output = quote! {};
 
--- a/contracts-proc/src/interface_id.rs
+++ b/contracts-proc/src/interface_id.rs
@@ -17,7 +17,6 @@ pub(crate) fn interface_id(
     input: TokenStream,
 ) -> TokenStream {
     let mut input = parse_macro_input!(input as ItemTrait);
-    let mut output = quote! {};
 
     let mut selectors = Vec::new();
     for item in &mut input.items {
@@ -68,7 +67,7 @@ pub(crate) fn interface_id(
     let (_impl_generics, ty_generics, where_clause) =
         input.generics.split_for_impl();
 
-    output.extend(quote! {
+    quote! {
         #(#attrs)*
         #vis trait #name #ty_generics #where_clause {
             #(#trait_items)*
@@ -78,9 +77,8 @@ pub(crate) fn interface_id(
                 #(#selectors)^*
             };
         }
-    });
-
-    output.into()
+    }
+    .into()
 }
 
 struct SelectorArgs {
--- a/contracts-proc/src/interface_id.rs
+++ b/contracts-proc/src/interface_id.rs
@@ -11,7 +11,7 @@ use syn::{
     parse_macro_input, FnArg, ItemTrait, LitStr, Result, Token, TraitItem,
 };
 
-/// Computes interface id as an associated constant for the trait.
+/// Computes an interface id as an associated constant for the trait.
 pub(crate) fn interface_id(
     _attr: &TokenStream,
     input: TokenStream,
--- a/contracts-proc/src/interface_id.rs
+++ b/contracts-proc/src/interface_id.rs
@@ -1,3 +1,5 @@
+//! Defines the `#[interface_id]` procedural macro.
+
 use std::mem;
 
 use convert_case::{Case, Casing};
--- a/contracts-proc/src/interface_id.rs
+++ b/contracts-proc/src/interface_id.rs
@@ -10,7 +10,10 @@ use syn::{
 };
 
 /// Computes interface id as an associated constant for the trait.
-pub(crate) fn interface(_attr: TokenStream, input: TokenStream) -> TokenStream {
+pub(crate) fn interface_id(
+    _attr: &TokenStream,
+    input: TokenStream,
+) -> TokenStream {
     let mut input = parse_macro_input!(input as ItemTrait);
     let mut output = quote! {};
 
@@ -48,7 +51,7 @@ pub(crate) fn interface(_attr: TokenStream, input: TokenStream) -> TokenStream {
         let arg_types: Vec<_> = args
             .filter_map(|arg| match arg {
                 FnArg::Typed(t) => Some(t.ty.clone()),
-                _ => None,
+                FnArg::Receiver(_) => None,
             })
             .collect();
 
--- a/contracts-proc/src/interface_id.rs
+++ b/contracts-proc/src/interface_id.rs
@@ -73,6 +73,7 @@ pub(crate) fn interface_id(
             #(#trait_items)*
 
             #[doc = concat!("Solidity interface id associated with ", stringify!(#name), " trait.")]
+            #[doc = "Computed as a XOR of selectors for each function in the trait."]
             const INTERFACE_ID: u32 = {
                 #(#selectors)^*
             };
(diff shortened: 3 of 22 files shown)


examples/erc20/src/lib.rs
```diff
@@ -105,4 +105,12 @@ impl Erc20Example {
         self.pausable.when_not_paused()?;
         self.erc20.transfer_from(from, to, value).map_err(|e| e.into())
     }
+
+    fn supports_interface(
+        interface_id: FixedBytes<4>,
+    ) -> Result<bool, Vec<u8>> {
+        let interface_id = u32::from_be_bytes(*interface_id);
+        let supported = interface_id == <Erc20 as IErc20>::INTERFACE_ID;
```
qalisander, line 113: It also should check for IErc165 interface id

ggonzalez94: I'm sure there are better ways to achieve the same, but something like this should work

contracts-proc/src/interface.rs
```diff
@@ -0,0 +1,118 @@
...
+use proc_macro::TokenStream;
+use proc_macro2::Ident;
+use quote::quote;
+use syn::{
+    parse::{Parse, ParseStream},
+    parse_macro_input, FnArg, ItemTrait, LitStr, Result, Token, TraitItem,
+};
+
+/// Computes interface id as an associated constant for the trait.
+pub(crate) fn interface(_attr: TokenStream, input: TokenStream) -> TokenStream {
```
ggonzalez94, line 13: Maybe `erc165_interface` is a better name. Why? There might be other type of interface like macros we want to add in the future or the sdk itself might add one with this name

ggonzalez94 (commented):
This is looking good! I like the approach of making this a procedural macro. I still need to go through the macro implementation, but already left a couple comments

contracts-proc/src/interface.rs
bidzyyys: Totally agree with @ggonzalez94 in terms of better name for the macro attribute. For me it is not enough self-explanatory. I would name `#[erc165]`

bidzyyys (commented):
Great job @qalisander! 
I am not a fan of introducing `contracts-proc` lib only for a single macro. 
I am thinking if is there any way to achieve the same only with default trait impls like:
```
pub trait IInterface {
    const INTERFACE_ID: FixedBytes<4>;
}
pub trait IErc165 + IInterface { 
    const ERC165_INTERFACE_ID: FixedBytes<4> = fixed_bytes!("01ffffff"); 
    
    fn supports_interface(&self, interface_id: FixedBytes<4>) -> bool { 
        interface_id == Self::ERC165_INTERFACE_ID || interface_id == Self::INTERFACE_ID
    } 
}
```

Could you try this out? Or if you see any blockers let me know before spending time on it.

contracts-proc/src/interface_id.rs
```diff
@@ -0,0 +1,123 @@
...
+        input.generics.split_for_impl();
+
+    output.extend(quote! {
+        #(#attrs)*
+        #vis trait #name #ty_generics #where_clause {
+            #(#trait_items)*
+
+            /// Solidity interface id associated with current trait.
+            const INTERFACE_ID: u32 = {
+                #(#selectors)^*
```
qalisander, line 78: This one produces code that XOR all available selectors.
Like:
```rust
const INTERFACE_ID: u32 = SELECTOR_1 ^ SELECTOR_2 ^ SELECTOR_3 ^ ...;
```

contracts-proc/src/interface_id.rs
```diff
@@ -0,0 +1,123 @@
...
+            continue;
+        };
+
+        let mut override_name = None;
+        for attr in mem::take(&mut func.attrs) {
+            let Some(ident) = attr.path().get_ident() else {
+                func.attrs.push(attr);
+                continue;
+            };
+            if *ident == "selector" {
```
qalisander, line 34: Here we retrieve all attributes of type: `#[selector(name = "actualSelectorName")` in case actual selector's name should be different for abi

contracts-proc/src/interface.rs
qalisander: decided to go with `#[interface_id]` macro name

examples/erc20/src/lib.rs
```diff
@@ -105,4 +105,9 @@ impl Erc20Example {
         self.pausable.when_not_paused()?;
         self.erc20.transfer_from(from, to, value).map_err(|e| e.into())
     }
+
+    fn supports_interface(interface_id: FixedBytes<4>) -> bool {
+        Erc20::supports_interface(interface_id)
+            || Erc20Metadata::supports_interface(interface_id)
```
bidzyyys, line 111: Why `Erc20Metadata`?

examples/erc721/src/lib.rs
```diff
@@ -151,4 +151,9 @@ impl Erc721Example {
 
         Ok(())
     }
+
+    pub fn supports_interface(interface_id: FixedBytes<4>) -> bool {
+        Erc721::supports_interface(interface_id)
+            || Enumerable::supports_interface(interface_id)
```
bidzyyys, line 157: Why `Enumerable`?

contracts/src/token/erc721/extensions/enumerable.rs
```diff
@@ -144,6 +149,13 @@ impl IErc721Enumerable for Erc721Enumerable {
     }
 }
 
+impl IErc165 for Erc721Enumerable {
```
bidzyyys, line 152: I am not sure if we want to have it implemented always.

contracts/src/token/erc721/extensions/metadata.rs
```diff
@@ -58,3 +61,24 @@ impl IErc721Metadata for Erc721Metadata {
         self._base_uri.get_string()
     }
 }
+
+impl IErc165 for Erc721Metadata {
```
bidzyyys, line 65: I am not sure if we want to have it implemented always.

contracts/src/token/erc721/extensions/metadata.rs
```diff
@@ -58,3 +61,24 @@ impl IErc721Metadata for Erc721Metadata {
...
+        <Self as IErc721Metadata>::INTERFACE_ID
+            == u32::from_be_bytes(*interface_id)
+    }
+}
+
+#[cfg(all(test, feature = "std"))]
+mod tests {
+    // use crate::token::erc721::extensions::{Erc721Metadata, IErc721Metadata};
+
+    // TODO: IErc721Metadata should be refactored to have same api as solidity
```
bidzyyys, line 76: TODO

contracts/src/token/erc721/mod.rs
```diff
@@ -549,6 +555,13 @@ impl IErc721 for Erc721 {
     }
 }
 
+impl IErc165 for Erc721 {
```
bidzyyys, line 558: I am not sure if we want to have it implemented always.

contracts/src/utils/introspection/erc165.rs
```diff
@@ -0,0 +1,64 @@
...
+/// Implementers can declare support of contract interfaces, which others can
+/// query.
+///
+/// For an implementation, see [`Erc165`].
+///
+/// [ERC]: https://eips.ethereum.org/EIPS/eip-165
+#[interface_id]
+pub trait IErc165 {
+    /// Returns true if this contract implements the interface defined by
+    /// `interfaceId`. See the corresponding [ERC section]
```
bidzyyys, line 19:
```suggestion
    /// `interface_id`. See the corresponding [ERC section]
```

contracts/src/token/erc20/mod.rs
```diff
@@ -286,6 +290,13 @@ impl IErc20 for Erc20 {
     }
 }
 
+impl IErc165 for Erc20 {
```
bidzyyys, line 293: I am not sure if we want to have it implemented always.

contracts/src/token/erc20/extensions/metadata.rs
```diff
@@ -76,3 +81,22 @@ impl IErc20Metadata for Erc20Metadata {
         DEFAULT_DECIMALS
     }
 }
+
+impl IErc165 for Erc20Metadata {
```
bidzyyys, line 85: I am not sure if we want to have it implemented always.

bidzyyys (commented):
Looks good overall, left comments that bothers me.

examples/erc721/src/lib.rs
qalisander: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/4764ea50750d8bda9096e833706beba86918b163/contracts/token/ERC721/extensions/ERC721Enumerable.sol#L39

contracts/src/token/erc721/extensions/enumerable.rs
qalisander: Basically it is [implemented](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/530179a71f435e85ae9df9b9f12b5637cf229e5c/contracts/token/ERC721/extensions/ERC721Enumerable.sol#L39) for Erc721Enumerable in solidity

contracts/src/token/erc721/mod.rs
qalisander: It is [implemented](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/530179a71f435e85ae9df9b9f12b5637cf229e5c/contracts/token/ERC721/ERC721.sol#L19) in solidity for Erc721

contracts/src/token/erc721/extensions/metadata.rs
qalisander: It is also [implemented](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/530179a71f435e85ae9df9b9f12b5637cf229e5c/contracts/token/ERC721/ERC721.sol#L50) for Erc721Metadata at solidity sources

contracts/src/token/erc721/extensions/metadata.rs
qalisander: You implemented metadata extension differentl from solidity. It has `base_uri` function instead of `token_uri`. And now it exposes "baseUri()" function into abi that is not like that in solidity. I think we will refactor it later

contracts-proc/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "openzeppelin-stylus-proc"
+description = "Procedural macros for OpenZeppelin Stylus contracts"
```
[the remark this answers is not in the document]
qalisander: Do you think we should also change description of contracts from `OpenZeppelin Contracts for Stylus` to `OpenZeppelin smart contracts library for Stylus`? Just seeking for consistency among projects.
cc @ggonzalez94

examples/erc721/src/lib.rs
qalisander: Basically it is [implemented](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/530179a71f435e85ae9df9b9f12b5637cf229e5c/contracts/token/ERC721/extensions/ERC721Enumerable.sol#L39) for Erc721Enumerable in solidity

contracts/src/token/erc20/mod.rs
qalisander: Basically `Erc20` compare to `Erc721` doesn't have `Erc165` support. I think original issue is [here](https://github.com/OpenZeppelin/openzeppelin-contracts/issues/1073#issuecomment-478674541) was closed due to "complexity". Some other implementations of Erc20 standard (not open zeppelin) have it.
Also seeking around github by actual Erc20 interface id `0x36372b07`, I could found find projects that rely on the Erc20 with interface support.
In fact at stylus contracts we are not forcing users to expose `supportsInterface(bytes4)`. But they have an option to do so.
@ggonzalez94 what do you think?

examples/erc20/src/lib.rs
qalisander: discussion in the other comment

contracts-proc/src/interface_id.rs
```diff
@@ -0,0 +1,122 @@
...
+            let _: Result<Token![,]> = input.parse();
+        }
+
+        if let Some(name) = name {
+            Ok(Self { name })
+        } else {
+            error!(@input.span(), r#""name" is required"#);
+        }
+    }
+}
```
ggonzalez94, line 122: Maybe this is a dumb question, but if we are only expecting the "name" attribute when parsing can't we just rely on `syn` parsing and do something simpler like this?
```
impl Parse for SelectorArgs {
    fn parse(input: ParseStream) -> Result<Self> {
        let ident: Ident = input.parse()?;
        let _: Token![=] = input.parse()?;
        let lit: LitStr = input.parse()?;

        if ident != "name" {
            return Err(syn::Error::new_spanned(ident, "expected 'name'"));
        }

        Ok(SelectorArgs { name: lit.value() })
    }
}
```

I run the unit tests and all are passing with this.

contracts-proc/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+        return error.to_compile_error().into();
+    }};
+    (@ $tokens:expr, $($msg:expr),+ $(,)?) => {{
+        return Err(syn::Error::new(syn::spanned::Spanned::span(&$tokens), format!($($msg),+)))
+    }};
+}
+
+mod interface_id;
+
+/// Computes interface id as an associated constant `INTERFACE_ID` for the trait
```
ggonzalez94, line 22:
```suggestion
/// Computes the interface id as an associated constant `INTERFACE_ID` for the trait
```

contracts-proc/src/interface_id.rs
ggonzalez94: Maybe it is worth adding some comments to the parts of the macro that are not obvious? For future reference and maintainers

ggonzalez94 (commented):
Great job here! I just did a second pass and is looking good.
I left a few comments/suggestions and tomorrow I'll go through the open questions

contracts-proc/src/interface_id.rs
qalisander: `selector` attribute parsing was copied from stylus-sdk :)
Just errors are better there.. 
e.g if you would write `#[selector(selector ! "mySelector")]` the first error your version would show is `expected '='`. Or if there are some additional tokens just general error.
But I agree it can be refactored!

contracts-proc/src/interface_id.rs
```diff
@@ -0,0 +1,122 @@
...
+                }
+                let args: SelectorArgs = match attr.parse_args() {
+                    Ok(args) => args,
+                    Err(error) => error!(ident, "{}", error),
+                };
+                override_name = Some(args.name);
+                continue;
+            }
+            func.attrs.push(attr);
+        }
```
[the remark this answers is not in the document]
qalisander: Changed, but saved `mem::take(&mut ..)`. It is common pattern in rust. May be when they add `drain_filter(..)` or smth would make sense to refactor

contracts/src/token/erc721/mod.rs
ggonzalez94: We will have to decide on an approach for all the standards in our library, so we can use this thread to do so(I've also opened an internal chat to get our Solidity team opinion). I'm leaning more on not adding it by default and letting the user opt in when they want. 
That being said when the standard defines that `ERC165` should be part of the contract we should do it. That's the case with [ERC721](https://eips.ethereum.org/EIPS/eip-721)

contracts/src/token/erc20/mod.rs
ggonzalez94: https://github.com/OpenZeppelin/rust-contracts-stylus/pull/281#discussion_r1772048963
^^

contracts/src/token/erc721/mod.rs
ggonzalez94: @qalisander after discussing internally and doing some research it seems `ERC165` is not as widely used that is worth adding it to all our contract. What I suggest we do:
1. Add it automatically to **only** the standards that require it as part of the ERC(e.g ERC721)
2. Add a section in our docs that explains how to add it to your own contract. This could be under the API reference for ERC165 or similar.

What do you think?

qalisander: I think that would look a bit inconsistent. I even wondering do we need `ERC165` that much just for `ERC721` .. Also I suspect, it is not widely used mostly, because it is not implemented for the contracts like erc20.
1.  Consider that rust is different from solidity. And with current implementation just to make `support_interface` function visible, library user should import `openzeppelin_stylus::utils::introspection::erc165::IErc165` module in the scope (otherwise it won't be even compiled). And then reexport it within `#[public]` macro. So we are not "automatically" implementing the standard as I see, but making it simple to implement for users.
2. That's a great point. I need to add a guide about erc165 to antora docs. But seems for me the `ERC165` guide (for erc20) would be more simple if we implement `IErc165` trait for `ERC20` also.

qalisander: And one more point. There is an "orphan rule" in rust that disallows to user implement foreign trait (like 'IErc165` in our case) for a foreign contract type (like `Erc20`). In this case the `IErc165` trait seems even less practical for users if they would try to implement it for Erc20. May be it would even make sense to remove it..

ggonzalez94: > I think that would look a bit inconsistent. 
I don't think it is inconsistent if we explicitly say: We will only add it to standards that require it and give the user the option to add it to any contract(s) they think is worth it.

> I even wondering do we need ERC165 that much just for ERC721
It wouldn't be for ERC721 only, since the implementation would be available for anyone wanting to add this to contracts they would like to specify which interfaces they support.

> And then reexport it within #[public] macro. So we are not "automatically" implementing the standard as I see, but making it simple to implement for users.
What do you mean here? We are adding it to the #[public] impl block of contracts. That means it would automatically be part of the inherited contract api right?

qalisander: > What do you mean here? We are adding it to the #[public] impl block of contracts. That means it would automatically be part of the inherited contract api right?
We are doing so inside examples. User also can opt out to export it to the abi, even if trait `IErc165` is implemented in our library. That is the difference from solidity, where as I see `Erc721` will always have `supportInterface(...)` abi public

ggonzalez94: Ok... after discussing this today I agree it makes sense to add it to implementations of main interfaces and it doesn't come at any cost to users that don't want it(since it is not part of the public abi and they still need to explicitely opt in). Only potential downside is a bit of gas overhead when inheriting from multiple contracts.

ggonzalez94 (approved):
@qalisander this looks ready to go IMO. The only thing left to do is documentation, but you can decide if you want to include it as part of this PR or create a new one for that

PR #289, opened 2024-09-17
feat: SafeErc20 utility

### LINKED ISSUE #279: [Feature]: SafeERC20 utility

Wrappers around ERC-20 operations that throw on failure.
Reference implementation: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/utils/SafeERC20.sol

### Contribution Guidelines

0xNeshi:
Resolves #279

I think I did everything that is required. Let me know if I missed something.

If you prefer a different code style, naming convention and/or anything similar, let me know and I'll update (e.g. check out how tests for this are organized). 

Used Solidity version [SafeERC20](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/4764ea50750d8bda9096e833706beba86918b163/contracts/token/ERC20/utils/SafeERC20.sol#L12) as inspiration for both implementation and tests.
 
#### PR Checklist

- [x] Tests
- [x] Documentation


contracts/src/token/erc20/utils/safe_erc20.rs
```diff
@@ -0,0 +1,153 @@
...
+    /// contract), relaxing the requirement on the return value: the return
+    /// value is optional (but if data is returned, it must not be false).
+    /// @param token The token targeted by the call.
+    /// @param data The call data (encoded using abi.encode or one of its
+    /// variants).
+    ///
+    /// This is a variant of {_callOptionalReturnBool} that reverts if call
+    /// fails to meet the requirements.
+    fn call_optional_return(&self, data: Vec<u8>) -> Result<(), Error> {
+        match RawCall::new()
```
0xNeshi, line 95: Certain things are unclear to me here as the docs aren't clear enough:
1. How can one get the address of a contract (in this case the ERC20 token) without passing it explicitly? Similar to how in Solidity one does `address(contract)`.
2. My intuition says that [RawCall + limit_return_data](https://docs.arbitrum.io/stylus/reference/rust-sdk-guide#rawcall-and-unsafe-calls) is more performant here, but I'm a bit confused by this sentence in the docs: ``Note that the [call](https://docs.rs/stylus-sdk/latest/stylus_sdk/call/struct.RawCall.html#method.call) method is `unsafe` when reentrancy is enabled.``. May be a dumb question, but better safe than sorry - does this mean the `call` is safe to use here, but we let the caller of the function worry about reentrancy?

qalisander (commented):
Hey @0xNeshi and thanks for your contribution!
Appreciate it!

Hard to tell for me right now what kind of design should be the best for `SafeErc20` with current state of the stylus-sdk.

IMHO we can try to build a working example first with end-to-end tests. Then later dive into optimization if it will make sense.
There is no good interoperability between rust struct, that defines a contract in stylus and an external call to that contract (deployed at some address) as it is in solidity.
So we can try to add `Erc20` solidity abi into `sol_interface!` macro like [IERC721Receiver example](https://github.com/OpenZeppelin/rust-contracts-stylus/blob/e4ba4e583824c0d0b4e78791e8349e689cb14436/contracts/src/token/erc721/mod.rs#L159).
And use call contract's type generated by `sol_interface!` as a first argument like it is in [solidity SafeErc20 example](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/4764ea50750d8bda9096e833706beba86918b163/contracts/token/ERC20/utils/SafeERC20.sol#L34).
But we should process boolean responses from underlying `Erc20` contract differently as it is specified in the [solidity counterpart](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/4764ea50750d8bda9096e833706beba86918b163/contracts/token/ERC20/utils/SafeERC20.sol#L12).

Then we can use [IErc20 trait](https://github.com/OpenZeppelin/rust-contracts-stylus/blob/e4ba4e583824c0d0b4e78791e8349e689cb14436/contracts/src/token/erc20/mod.rs#L115) to create wrapped Erc20 contract. It should route most calls externally using to the `Call::new(..)` api from the stylus-sdk (like [Erc721Receiver example](https://github.com/OpenZeppelin/rust-contracts-stylus/blob/e4ba4e583824c0d0b4e78791e8349e689cb14436/contracts/src/token/erc721/mod.rs#L1113))

How do you feel about this plan?

contracts/src/token/erc20/utils/safe_erc20.rs
qalisander: Replied in the other comment about different approach

0xNeshi:
@qalisander sounds like a good plan!

I implemented what I understood to be what you had imagined. See the most recent changes. 

Note: I'm trying to implement a working solution for the regular ERC20 `transfer` function (i.e. working on SafeERC20's `safe_transfer` function), which returns a `bool`. Once successful, I will handle the general case where the function might return a random type or not even return anything (e.g. USDT).

I also implemented one "happy flow" test for `safe_transfer`, but I have been banging my head against one problem for the last 6 hours though. It seems that calling the function like [erc20.transfer(call, to, value)](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/289/files#diff-8af311813d2aea1c03c4655d40bc21aa45b778e22e48d767bbf70d1f71d031e9R73) doesn't pass the expected msg sender. **In other words, ERC20's `msg.sender` does not seem to be the same as SafeErc20's `msg::sender()`.**

For context, I implemented a [mock ERC20 contract](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/289/files#diff-2e628b32ee8ddbb95bb4ef464de07acca28552b95845819edc543b6adfb85862) that on calling `transfer` just updates the balances as expected (w/o checks), i.e. it reduces the amount from `msg.sender` and increases it for `to` (address). But the test keeps failing and I managed to narrow it down is that the error is of type [stylus_sdk::call::Error::Reverted](https://docs.rs/stylus-sdk/latest/stylus_sdk/call/enum.Error.html#variant.Revert) (it's a bit difficult to debug Stylus contracts for an amateur, maybe it's just a skill issue).
 
To test my hypothesis that `msg.sender` within `ERC20` is not the same as `msg::sender()` within `SafeErc20`, I changed mock ERC20's `transfer` function to accept an `address from` argument (basically turning it into `transferFrom`), and passed `msg::sender()` as the value for this argument from within `SafeErc20` like [erc20.transfer(call, msg::sender(), to, value)](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/289/files#diff-8af311813d2aea1c03c4655d40bc21aa45b778e22e48d767bbf70d1f71d031e9R73) - after this the test started passing!

I tried another approach by using `RawCall::new()`, passing the encoded calldata and invoking the ERC20 contract that way, but the result was the same (a failing test).

Might you have an idea what the problem might be here? Is this an SDK issue or have I misconfigured something?

scripts/e2e-tests.sh
```diff
@@ -11,4 +11,4 @@ cargo +"$NIGHTLY_TOOLCHAIN" build --release --target wasm32-unknown-unknown -Z b
 export RPC_URL=http://localhost:8547
 # We should use stable here once nitro-testnode is updated and the contracts fit
 # the size limit. Work tracked [here](https://github.com/OpenZeppelin/rust-contracts-stylus/issues/87)
-cargo +"$NIGHTLY_TOOLCHAIN" test --features std,e2e --test "*"
+cargo +"$NIGHTLY_TOOLCHAIN" test --features e2e -p safe-erc20-example --test safe-erc20
```
0xNeshi, line 14: Will revert this once ready for merge, using this for easier testing

qalisander: Yeah. If yours IDE allows to set build settings for tests, you can also run specific integration test from IDE's UI.

0xNeshi: Regarding the [msg sender issue above](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/289#issuecomment-2362200392), I hope the explanation was clear enough. The same issue is there for `safe_transfer_from`, as the msg sender isn't the one expected (the original sender of tx).

examples/safe-erc20/tests/mock/erc20.rs
```diff
@@ -0,0 +1,40 @@
...
+        function balanceOf(address account) public override view returns (uint256) {
+            return super.balanceOf(account);
+        }
+
+        function mint(address account, uint256 value) public {
+            super._mint(account, value);
+        }
+
+        function transfer(address to, uint256 amount) public override returns (bool) {
+            return super.transfer(to, amount);
```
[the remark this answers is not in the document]
0xNeshi: I think I just realized that this is _the expected_ behavior. Looking at how [Solidity SafeERC20](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/4764ea50750d8bda9096e833706beba86918b163/contracts/token/ERC20/utils/SafeERC20.sol#L12) is implemented, it now seems obvious that it was never intended to be that the original tx sender is the msg.sender for the nested call.

contracts/src/token/erc20/utils/safe_erc20.rs
```diff
@@ -0,0 +1,225 @@
...
+impl SafeErc20 {
+    /// Imitates a Solidity high-level call, relaxing the requirement on the return value:
+    /// if data is returned, it must not be `false`, otherwise calls are assumed to be successful.
+    fn call_optional_return(
+        &self,
+        token: Address,
+        data: &[u8],
+    ) -> Result<(), Error> {
+        match RawCall::new()
+            .gas(gas_left())
```
0xNeshi, line 203: It's not really clear to me whether explicitly setting the gas remaining is necessary, so I set it just in case.

contracts/src/token/erc20/utils/safe_erc20.rs
```diff
@@ -0,0 +1,225 @@
...
+
+impl SafeErc20 {
+    /// Imitates a Solidity high-level call, relaxing the requirement on the return value:
+    /// if data is returned, it must not be `false`, otherwise calls are assumed to be successful.
+    fn call_optional_return(
+        &self,
+        token: Address,
+        data: &[u8],
+    ) -> Result<(), Error> {
+        match RawCall::new()
```
0xNeshi, line 202: As far as I understand, [RawCall is unsafe](https://docs.arbitrum.io/stylus/reference/rust-sdk-guide#rawcall-and-unsafe-calls) and should be called with `flush_storage_cache` when reentrancy is enabled. Let me know how you'd like to do this.

0xNeshi, line 202: The reason I opted for `RawCall` here is that it enables us to set the return data size to just 32 bytes, which assumes the underlying ERC20 returns a boolean. It also does not make assumptions about the return value of the underlying ERC20.

examples/safe-erc20/Cargo.toml
bidzyyys, line 1: Try to make dependencies consistent with our examples like for [Erc20](https://github.com/OpenZeppelin/rust-contracts-stylus/blob/main/examples/erc20/Cargo.toml)

examples/safe-erc20/src/constructor.sol
bidzyyys, line 1: With this [PR](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/295) merged we do not need empty Solidity constructor.

qalisander:
@0xNeshi we're quite busy currently with security audit fixes and another huge pr review.
We will come and give a thorough review of your pr soon, once we finalize other things!
cc @bidzyyys

0xNeshi:
No problem!

bidzyyys:
> No problem!

Thanks @0xNeshi for your patience, we should have more time later this week!

contracts/src/token/erc20/utils/safe_erc20.rs
```diff
@@ -0,0 +1,229 @@
...
+    pub fn safe_decrease_allowance(
+        &mut self,
+        token: Address,
+        spender: Address,
+        requested_decrease: U256,
+    ) -> Result<(), Error> {
+        let erc20 = IERC20::new(token);
+        let call = Call::new_in(self);
+        let current_allowance =
+            erc20.allowance(call, address(), spender).or({
```
bidzyyys, line 136: Refactor `erc20.allowance` call into separate fn for both `safe_decrease_allowance` and `safe_increase_allowance`.

contracts/src/token/erc20/utils/safe_erc20.rs
```diff
@@ -0,0 +1,229 @@
...
+            value: U256,
+            selector: &[u8],
+        ) -> Vec<u8> {
+            let encoded_args = (spender, value).abi_encode_params();
+            [&selector[..4], &encoded_args].concat()
+        }
+
+        // Try performing the approval with the desired value
+        let approve_data = build_approve_calldata(spender, value, &selector);
+        if self._call_optional_return(token, &approve_data).is_ok() {
```
bidzyyys, line 183: You need to make sure it returns `true`

contracts/src/token/erc20/utils/safe_erc20.rs
bidzyyys: For Stylus re-entrancy is disabled by default.

contracts/src/token/erc20/utils/safe_erc20.rs
bidzyyys: No need for doing this I would say.

contracts/src/token/erc20/utils/safe_erc20.rs
```diff
@@ -0,0 +1,229 @@
...
+            _ => {
+                Err(Error::SafeErc20FailedOperation(SafeErc20FailedOperation {
+                    token,
+                }))
+            }
+        }
+    }
+}
+
+fn encodes_true(data: &[u8]) -> bool {
```
bidzyyys, line 226: Write unit tests for this fn.

contracts/src/token/erc20/utils/safe_erc20.rs
```diff
@@ -0,0 +1,229 @@
...
+        token: Address,
+        data: &[u8],
+    ) -> Result<(), Error> {
+        match RawCall::new()
+            .gas(gas_left())
+            .limit_return_data(0, 32)
+            .call(token, data)
+        {
+            Ok(data)
+                if (data.is_empty() && Address::has_code(&token))
```
bidzyyys, line 212: For me `Address::has_code(&token)` should be checked before doing the `Call`. No need to continue when the address is not the address of Erc20 token.

contracts/src/token/erc20/utils/safe_erc20.rs
```diff
@@ -0,0 +1,229 @@
...
+                    || (!data.is_empty() && encodes_true(&data)) =>
+            {
+                Ok(())
+            }
+            _ => {
+                Err(Error::SafeErc20FailedOperation(SafeErc20FailedOperation {
```
bidzyyys, line 218: I would unpack data with `.map_err(|_| Error::SafeErc20FailedOperation(SafeErc20FailedOperation )?` and then perform the verification later.

contracts/src/token/erc20/utils/safe_erc20.rs
bidzyyys, line 226: Missing docs for the fn.

examples/safe-erc20/tests/er20_that_doesnt_return.rs
bidzyyys, line 1: rename `doesnt` -> `does_not`

examples/safe-erc20/tests/er20_that_doesnt_return.rs
```diff
@@ -0,0 +1,456 @@
...
+        let _ = watch!(erc20_alice.mint(safe_erc20_addr, balance));
+
+        let initial_safe_erc20_balance =
+            erc20_alice.balanceOf(safe_erc20_addr).call().await?._0;
+        let initial_bob_balance =
+            erc20_alice.balanceOf(bob_addr).call().await?._0;
+        assert_eq!(initial_safe_erc20_balance, balance);
+        assert_eq!(initial_bob_balance, U256::ZERO);
+
+        let _ = receipt!(safe_erc20_alice.safeTransfer(
```
bidzyyys, line 39: Check `Transfer` event here.

examples/safe-erc20/tests/er20_that_doesnt_return.rs
```diff
@@ -0,0 +1,456 @@
...
+        let _ = watch!(erc20_alice.approve(safe_erc20_addr, value));
+
+        let initial_alice_balance =
+            erc20_alice.balanceOf(alice_addr).call().await?._0;
+        let initial_bob_balance =
+            erc20_alice.balanceOf(bob_addr).call().await?._0;
+        assert_eq!(initial_alice_balance, balance);
+        assert_eq!(initial_bob_balance, U256::ZERO);
+
+        let _ = receipt!(safe_erc20_alice.safeTransferFrom(
```
bidzyyys, line 81: Check erc20 event here.

bidzyyys (commented):
Great job @0xNeshi, left some comments.
What I am missing in E2E tests are:
- [x] missing checks for Erc20 events on successful operations
- [ ] missing checks for arithmetic overflow for increasing approval
- [ ] missing checks for failed operations like failed transfer due to insufficient balance or approval

contracts/src/token/erc20/utils/safe_erc20.rs
```diff
@@ -0,0 +1,284 @@
...
+        interface IErc20 {
+            function allowance(address owner, address spender) external view returns (uint256);
+            function approve(address spender, uint256 value) external returns (bool);
+            function transfer(address to, uint256 value) external returns (bool);
+            function transferFrom(address from, address to, uint256 value) external returns (bool);
+        }
+    }
+}
+sol_storage! {
+    /// State of the SafeErc20 Contract.
```
bidzyyys, line 70:
```suggestion
    /// State of the [`SafeErc20`] Contract.
```

contracts/src/token/erc20/utils/safe_erc20.rs
```diff
@@ -0,0 +1,284 @@
...
+}
+
+/// NOTE: Implementation of [`TopLevelStorage`] to be able use `&mut self` when
+/// calling other contracts and not `&mut (impl TopLevelStorage +
+/// BorrowMut<Self>)`. Should be fixed in the future by the Stylus team.
+unsafe impl TopLevelStorage for SafeErc20 {}
+
+/// Required interface of an [`SafeErc20`] utility contract.
+pub trait ISafeErc20 {
+    /// The error type associated to this ERC-20 trait implementation.
```
bidzyyys, line 81:
```suggestion
    /// The error type associated to this trait implementation.
```

contracts/src/token/erc20/utils/safe_erc20.rs
```diff
@@ -0,0 +1,284 @@
...
+    SafeErc20FailedOperation(SafeErc20FailedOperation),
+    /// Indicates a failed [`ISafeErc20::decrease_allowance`] request.
+    SafeErc20FailedDecreaseAllowance(SafeErc20FailedDecreaseAllowance),
+}
+
+pub use token::*;
+#[allow(missing_docs)]
+mod token {
+    alloy_sol_types::sol! {
+        /// Interface of the ERC-20 standard as defined in the ERC.
```
bidzyyys, line 60:
```suggestion
        /// Interface of the ERC-20 token.
```

contracts/src/token/erc20/utils/safe_erc20.rs
```diff
@@ -0,0 +1,284 @@
...
+        }
+
+        match RawCall::new()
+            .gas(gas_left())
+            .limit_return_data(0, 32)
+            .call(token, &call.abi_encode())
+        {
+            Ok(data)
+                if data.is_empty()
+                    || (!data.is_empty() && encodes_true(&data)) =>
```
bidzyyys, line 249: Optimize this `if`

contracts/src/token/erc20/utils/safe_erc20.rs
```diff
@@ -0,0 +1,284 @@
...
+                    token,
+                })
+            })?;
+
+        Ok(U256::from_be_slice(&allowance))
+    }
+}
+
+fn encodes_true(data: &[u8]) -> bool {
+    data[..data.len() - 1].iter().all(|&byte| byte == 0)
```
bidzyyys, line 282: Refactor to non-panic and easier to read.

contracts/src/token/erc20/utils/safe_erc20.rs
```diff
@@ -0,0 +1,280 @@
...
+}
+
+/// A [`SafeErc20`] error.
+#[derive(SolidityError, Debug)]
+pub enum Error {
+    /// Error type from [`erc20::Erc20`] contract [`erc20::Error`].
+    Erc20(erc20::Error),
+    /// An operation with an ERC-20 token failed.
+    SafeErc20FailedOperation(SafeErc20FailedOperation),
+    /// Indicates a failed [`ISafeErc20::decrease_allowance`] request.
```
0xNeshi, line 52:
```suggestion
    /// Indicates a failed [`ISafeErc20::safe_decrease_allowance`] request.
```

contracts/src/token/erc20/utils/safe_erc20.rs
```diff
@@ -0,0 +1,280 @@
...
+
+sol! {
+    /// An operation with an ERC-20 token failed.
+    ///
+    /// * `token` - Address of the ERC-20 token.
+    #[derive(Debug)]
+    #[allow(missing_docs)]
+    error SafeErc20FailedOperation(address token);
+
+    /// Indicates a failed [`ISafeErc20::decrease_allowance`] request.
```
0xNeshi, line 31:
```suggestion
    /// Indicates a failed [`ISafeErc20::safe_decrease_allowance`] request.
```

examples/safe-erc20/tests/er20_that_doesnt_return.rs
0xNeshi: Just noticed that `er20_...` should also be `erc20_...`

bidzyyys (approved):

qalisander (approved):
Good job!

PR #295, opened 2024-09-20
fix(e2e): deploy contracts without Solidity constructor when not needed

TucksonDev:
This PR changes the `e2e` testing library to allow deployments of contracts without a constructor., when the constructor file is not present. This is already allowed by [koba](https://github.com/OpenZeppelin/koba/blob/main/src/generator.rs#L27-L30).

~Resolves #???~

#### PR Checklist

- [x] Tests
- [ ] ~Documentation~ (not required)

### DIFF
--- a/examples/ecdsa/src/constructor.sol
+++ b/examples/ecdsa/src/constructor.sol
@@ -1,6 +1,0 @@
-// SPDX-License-Identifier: MIT
-pragma solidity ^0.8.21;
-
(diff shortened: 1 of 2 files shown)


bidzyyys (commented):
Looks fine for me 👍

qalisander (approved):
Hey Jose! Thanks for the contribution.
I've fixed final formatting issue and I think we are ready to merge

PR #297, opened 2024-09-21
chore: update keywords for crates.io

qalisander:
Adds more relevant and different keywords for different crates.


contracts/Cargo.toml
```diff
@@ -1,9 +1,9 @@
 [package]
 name = "openzeppelin-stylus"
-categories = ["no-std", "wasm"]
+categories = ["cryptography::cryptocurrencies", "no-std", "wasm"]
 description = "OpenZeppelin Contracts for Stylus"
 edition.workspace = true
-keywords.workspace = true
+keywords = ["arbitrum", "ethereum", "stylus", "smart-contracts", "web3"]
```
ggonzalez94, line 6:
```suggestion
keywords = ["arbitrum", "ethereum", "stylus", "smart-contracts", "standards"]
```

lib/e2e/Cargo.toml
```diff
@@ -2,10 +2,11 @@
 name = "e2e"
 description = "End-to-end Testing for Stylus"
 version = "0.1.0"
+categories = ["development-tools::testing", "cryptography::cryptocurrencies"]
+keywords = ["arbitrum", "ethereum", "stylus", "tests", "openzeppelin"]
```
ggonzalez94, line 6: We're not including OZ as a keyword in the other packages, and I think is not the best practice to include company name(and it is already part of the package name and owners)
```suggestion
keywords = ["arbitrum", "ethereum", "stylus", "tests"]
```

lib/motsu-proc/Cargo.toml
```diff
@@ -2,7 +2,6 @@
 name = "motsu-proc"
 description = "Mostu's Procedural Macros"
 edition.workspace = true
```
ggonzalez94, line 4: This will be published right? Shouldn't we add keywords as well?

lib/motsu/Cargo.toml
```diff
@@ -2,7 +2,8 @@
 name = "motsu"
 description = "Unit Testing for Stylus"
 edition.workspace = true
-keywords.workspace = true
+categories = ["development-tools::testing", "cryptography::cryptocurrencies"]
+keywords = ["arbitrum", "ethereum", "stylus", "tests", "openzeppelin"]
```
ggonzalez94, line 6: We're not including OZ as a keyword in the other packages, and I think is not the best practice to include company name(and it is already part of the package name and owners)

qalisander: It is not the part of the name "motsu", that is why I included it. But yeah, owner is already "openzeppelin". Will remove it

lib/motsu-proc/Cargo.toml
qalisander: But it cost nothing to duplicate keywords here also. I'll update them

lib/motsu/Cargo.toml
qalisander: We are allowed to have up to 5 keywords per crate. So added "unit-tests" for `motsu` and "integration-testing" for `e2e` instead of "openzeppelin". Those exact tags are frequently used at crates.io to describe corresponding crates.

ggonzalez94 (approved):
LGTM! Nice addition adding keywords for unit and integration tests.

Issue #62, opened 2024-05-15
ERC165


bidzyyys:
Implement ERC165 Standard
Check https://docs.openzeppelin.com/contracts/5.x/api/utils#ERC165

Add the standard to the following contracts:
- [ ] ERC721URIStorage
- [ ] ERC721

PR #300, opened 2024-09-27
fix: docs.rs url

ggonzalez94:
Fixes #299 

- We have to include the version because the default one is empty(until we publish 0.1.0 at least), but anyways it is important to explicitly include the version.
- Add crate suffix after version, otherwise the link is still broken
- Fix crypto urls
- Add GH workflow that checks that links are not broken on `adoc`and `md` files :)

PR #304, opened 2024-10-02
docs: erc165 + structs

qalisander:
Follows #62

#### PR Checklist

- [ ] Tests
- [x] Documentation

### DIFF
--- a/docs/modules/ROOT/pages/utilities.adoc
+++ b/docs/modules/ROOT/pages/utilities.adoc
@@ -9,7 +9,7 @@ Here are some of the more popular ones.
 
 It's frequently helpful to know whether a contract supports an interface you'd like to use.
 https://eips.ethereum.org/EIPS/eip-165[`ERC-165`] is a standard that helps do runtime interface detection.
-Contracts for Stylus provides helpers for implementing ERC-165 in your contracts:
+Contracts for Stylus provide helpers for implementing ERC-165 in your contracts:
 
 * https://docs.rs/openzeppelin-stylus/0.1.0-rc/openzeppelin_stylus/utils/introspection/erc165/trait.IErc165.html[`IERC165`] — this is the ERC-165 trait that defines https://docs.rs/openzeppelin-stylus/0.1.0-rc/openzeppelin_stylus/utils/introspection/erc165/trait.IErc165.html#tymethod.supports_interface[`supportsInterface`]. In order to implement ERC-165 interface detection, you should manually expose https://docs.rs/openzeppelin-stylus/0.1.0-rc/openzeppelin_stylus/utils/introspection/erc165/trait.IErc165.html#tymethod.supports_interface[`supportsInterface`] function in your contract.
 
--- a/docs/modules/ROOT/pages/utilities.adoc
+++ b/docs/modules/ROOT/pages/utilities.adoc
@@ -1,0 +1,43 @@
+= Utilities
+
+The OpenZeppelin Stylus Contracts provide a ton of useful utilities that you can use in your project.
+For a complete list, check out the https://docs.rs/openzeppelin_stylus/utils/index.html[API Reference].
+Here are some of the more popular ones.
+
+[[introspection]]
+== Introspection
+
+As well as in Solidity, at Stylus, it's frequently helpful to know whether a contract supports an interface you'd like to use.
+https://eips.ethereum.org/EIPS/eip-165[`ERC-165`] is a standard that helps do runtime interface detection.
+Contracts provide helpers both for implementing ERC-165 in your contracts and querying other contracts:
+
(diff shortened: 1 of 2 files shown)


docs/modules/ROOT/pages/utilities.adoc
```diff
@@ -0,0 +1,43 @@
+= Utilities
+
+The OpenZeppelin Stylus Contracts provide a ton of useful utilities that you can use in your project.
+For a complete list, check out the https://docs.rs/openzeppelin_stylus/utils/index.html[API Reference].
```
ggonzalez94, line 4: Shouldn't we pin to a specific version? In this case `0.1.0-rc`, otherwise when we publish new releases the docs for the old version will point to the new api docs

docs/modules/ROOT/pages/utilities.adoc
```diff
@@ -0,0 +1,43 @@
...
+Here are some of the more popular ones.
+
+[[introspection]]
+== Introspection
+
+As well as in Solidity, at Stylus, it's frequently helpful to know whether a contract supports an interface you'd like to use.
```
ggonzalez94, line 10: We can keep the writing simpler
```suggestion
it's frequently helpful to know whether a contract supports an interface you'd like to use.
```

bidzyyys (approved):
Looks good to me!

docs/modules/ROOT/pages/utilities.adoc
```diff
@@ -0,0 +1,43 @@
...
+https://eips.ethereum.org/EIPS/eip-165[`ERC-165`] is a standard that helps do runtime interface detection.
+Contracts provide helpers both for implementing ERC-165 in your contracts and querying other contracts:
```
ggonzalez94, line 12: Slight rephrase, and I don't think we are providing support for checking interfaces right?
```suggestion
Contracts for Stylus provides helpers for implementing ERC-165 in your contracts:
```

docs/modules/ROOT/pages/utilities.adoc
```diff
@@ -0,0 +1,43 @@
...
+
+* https://docs.rs/openzeppelin_stylus/utils/introspection/erc165/trait.IErc165.html[`IERC165`] — this is the ERC-165 trait that defines https://docs.rs/openzeppelin_stylus/utils/introspection/erc165/trait.IErc165.html#tymethod.supports_interface[`supportsInterface`]. In order to implement ERC-165 interface detection, you should manually expose https://docs.rs/openzeppelin_stylus/utils/introspection/erc165/trait.IErc165.html#tymethod.supports_interface[`supportsInterface`] function in your contract.
```
ggonzalez94, line 14: Same here, we should pin to a specific version in docs.rs

docs/modules/ROOT/pages/utilities.adoc
```diff
@@ -0,0 +1,43 @@
...
+        Erc721::supports_interface(interface_id)
+    }
+}
+
+----
+
+[[structures]]
+== Structures
+
+Some use cases require more powerful data structures than arrays and mappings offered natively in Stylus sdk.
```
ggonzalez94, line 39: This is a nit, so feel free to ignore it.
```suggestion
Some use cases require more powerful data structures than arrays and mappings offered natively in alloy and the Stylus sdk.
```

docs/modules/ROOT/pages/utilities.adoc
```diff
@@ -0,0 +1,43 @@
...
+Contracts provides these libraries for enhanced data structure management:
```
ggonzalez94, line 40: I think we should name it Contracts for Stylus to be more consistent with the name of the library. Alternatively I'm happy to use this crate, or this library....
```suggestion
Contracts for Stylus provides these libraries for enhanced data structure management:
```

docs/modules/ROOT/pages/utilities.adoc
```diff
@@ -0,0 +1,43 @@
...
+
+- https://docs.rs/openzeppelin_stylus/utils/structs/bitmap/index.html[`BitMaps`]: Store packed booleans in storage.
```
ggonzalez94, line 42: Same here, we should pin to a specific version in docs.rs

ggonzalez94 (commented):
Looks good overall. Left some suggestions.
We should be careful with docs version and format and follow the format at #300. The current links on this PR don't work.

docs/modules/ROOT/pages/utilities.adoc
qalisander: This pr to the `main` branch. In this case make sense to change it everywhere in the `main` branch.

ggonzalez94: Should be resolved once we merge #300 back to main

docs/modules/ROOT/pages/utilities.adoc
qalisander: Yeah. Just use `Contracts for Stylus provide` without "s". Since it has already plural form

ggonzalez94: This is a nit. But Actually it is a singular entity even if it ends with "s"

> Grammatical correctness: "OpenZeppelin Contracts for Stylus" is treated as a singular entity, representing a single library or project. Therefore, the singular verb "exists" is the correct form to use

ggonzalez94 (approved):
Let's go 🚀  can you check the last nit I reopened on the comments?

docs/modules/ROOT/pages/utilities.adoc
qalisander: So be it. But I think there're many artifacts of "mine approach" around the codebase :)

PR #310, opened 2024-10-04
fix(L-03): Missing Checks in ERC-20 Implementation

### LINKED ISSUE #313: [Audit]: L-03 Missing Checks in ERC-20 Implementation

Some validation checks are missing in the current ERC-20 implementation when compared to
the Solidity library:
• In mod.rs , the _approve() function checks the spender's address but does not
verify that the owner's address is not zero, unlike in the Solidity library.
• The function directly sets the allowance in storage without
calling the internal   function, as is done in the Solidity version. This bypasses the address validation checks (owner and spender) that _approve() provides.
• The function currently calls _spend_allowance() before
, where the latter includes non-zero address checks for both the sender
and recipient. However, these checks should be performed prior to updating the allowance, as is done in the Solidity implementation, to prevent unnecessary storage writes in the event of a revert.
Consider adding the aforementioned missing checks to improve the security of the library.

### Contribution Guidelines

qalisander:
Add additional check to `approve` + emit_event flag like in solidity. Rely on that `approve` inside `_spend_allowance`.

Resolves #313

### DIFF
--- a/contracts/src/token/erc20/extensions/burnable.rs
+++ b/contracts/src/token/erc20/extensions/burnable.rs
@@ -158,7 +158,7 @@ mod tests {
     }
 
     #[motsu::test]
-    fn burns_from_errors_when_invalid_sender(contract: Erc20) {
+    fn burns_from_errors_when_invalid_approver(contract: Erc20) {
         let one = uint!(1_U256);
 
         contract
--- a/contracts/src/token/erc20/extensions/burnable.rs
+++ b/contracts/src/token/erc20/extensions/burnable.rs
@@ -168,7 +168,7 @@ mod tests {
             .set(one);
 
         let result = contract.burn_from(Address::ZERO, one);
-        assert!(matches!(result, Err(Error::InvalidSender(_))));
+        assert!(matches!(result, Err(Error::InvalidApprover(_))));
     }
 
     #[motsu::test]
--- a/contracts/src/token/erc20/mod.rs
+++ b/contracts/src/token/erc20/mod.rs
@@ -835,7 +835,7 @@ mod tests {
     }
 
     #[motsu::test]
-    fn transfer_from_errors_when_invalid_sender(contract: Erc20) {
+    fn transfer_from_errors_when_invalid_approver(contract: Erc20) {
         let alice = address!("A11CEacF9aa32246d767FCCD72e02d6bCbcC375d");
         let one = uint!(1_U256);
         contract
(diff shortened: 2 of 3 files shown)


contracts/src/token/erc20/mod.rs
```diff
@@ -825,7 +844,7 @@ mod tests {
             .setter(msg::sender())
             .set(one);
         let result = contract.transfer_from(Address::ZERO, alice, one);
-        assert!(matches!(result, Err(Error::InvalidSender(_))));
+        assert!(matches!(result, Err(Error::InvalidApprover(_))));
```
bidzyyys, line 847: Right now it looks weird with the test name. Maybe better test name?

contracts/src/token/erc20/extensions/burnable.rs
```diff
@@ -168,7 +168,7 @@ mod tests {
             .set(one);
 
         let result = contract.burn_from(Address::ZERO, one);
-        assert!(matches!(result, Err(Error::InvalidSender(_))));
+        assert!(matches!(result, Err(Error::InvalidApprover(_))));
```
bidzyyys, line 171: Right now it looks weird with the test name. Maybe better test name?

contracts/src/token/erc20/mod.rs
bidzyyys, line 1: Missing `transfer_from` improvement.

bidzyyys (commented):
Left few comments.

contracts/src/token/erc20/extensions/burnable.rs
qalisander: Changed `*_sender` to `*_approver` that is not ideal though, especially when names here follows our previous error convention "burns_from_errors_when_invalid_approver"
I think we should later refactor test names for `erc20`

contracts/src/token/erc20/mod.rs
qalisander: changed and replied in the other comment

contracts/src/token/erc20/mod.rs
qalisander: `transfer_from` is referenced as an example of function that will get behavior different then solidity. We don't need to changed 
fyi: how [transfer_from](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/cceac54953ccda8a9a028d0b9bf4378605fdf67e/contracts/token/ERC20/ERC20.sol#L149) is implemented in solidity.
image
Same as it is now.

bidzyyys (approved):
LGTM!

ggonzalez94 (approved):

Issue #282, opened 2024-09-12
Refactor functions that don't need to return a result 


ggonzalez94:
### What is the feature you would like to see?

Previously, all public methods were required to return a Result type with Vec<u8> as the error type. This is now optional. Specifically, if a method is "infallible" (i.e., it cannot produce an error), it does not need to return a Result type.

### Contribution Guidelines

- [X] I agree to follow this project's Contribution Guidelines

0xNeshi:
@ggonzalez94 @bidzyyys @qalisander there don't seem to be any "infallible" methods that are left to refactor, can you point out any?

Also, in [this comment](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/311#discussion_r1788083697) Gustavo suggests we update all examples to return explicit `Error` types instead of `Vec<u8>`.
@emnul you can do this as part of this PR - basically, wherever you see that `Result<[return-type], Vec<u8>>` is returned, update it to return an explicit `Result<[return-type], Error>` type, where `Error` is either imported from the inherited contract (e.g. `erc20::Error`), or is a wrapper around multiple errors.

To see how to create wrapper errors, check:

- [Erc20Permit wrapper error definition](https://github.com/OpenZeppelin/rust-contracts-stylus/blob/9192c1849680f0c0f60822ac288b172a8b276c89/contracts/src/token/erc20/extensions/permit.rs#L62-L72)
- [converting an inherited error into a wrapper error](https://github.com/OpenZeppelin/rust-contracts-stylus/blob/main/examples/merkle-proofs/src/lib.rs#L30)

emnul:
Thanks for taking another look at this one @0xNeshi ! I will turn my attention to #327 at this time

0xNeshi:
@emnul I'll assign you to that issue then, and take over this one :+1:

PR #311, opened 2024-10-04
fix(L-04): Public Exposure of pause and unpause Functions

### LINKED ISSUE #314: [Audit]: L-04 Public Exposure of pause and unpause Functions

The   contract exposes the pause and unpause functions as part of a public implementation block. This design automatically makes these functions available
in any inheriting contract's public ABI, without any access control restrictions. In the Solidity version of the Pausable contract, the corresponding functions are marked as internal . This forces developers to explicitly declare external or public functions that interact with
pause and unpause , and ensures that proper access control measures are applied.
Consider placing the pause and unpause functions in a non-public implementation block, preventing their exposure via the ABI by default. This will enforce explicit access control, reducing the risk of unintended function access.

### Contribution Guidelines

qalisander:
Methods `pause` and `unpause` are not exposed by default now.

Resolves #314

### DIFF
--- a/contracts/src/utils/pausable.rs
+++ b/contracts/src/utils/pausable.rs
@@ -8,8 +8,7 @@
 //! which can be added to the functions of your contract.
 //!
 //! Note that they will not be pausable by simply including this module,
-//! only once functions [`Pausable::when_not_paused`] and
-//! [`Pausable::when_paused`] are put in place.
+//! only once function [`Pausable::when_not_paused`] is put in place.
 //!
 //! Note that [`Pausable::pause`] and [`Pausable::unpause`] methods are not
 //! exposed by default.
--- a/contracts/src/utils/pausable.rs
+++ b/contracts/src/utils/pausable.rs
@@ -9,6 +9,10 @@
 //!
 //! Note that they will not be pausable by simply including this module,
 //! only once the modifiers are put in place.
+//!
+//! Note that [`Pausable::pause`] and [`Pausable::unpause`] methods are not
+//! exposed by default.
+//! You should expose them manually in your contract's abi.
 
 use alloy_sol_types::sol;
 use stylus_proc::{public, sol_storage, SolidityError};
@@ -68,40 +72,6 @@ impl Pausable {
         self._paused.get()
     }
 
-    /// Triggers `Paused` state.
-    ///
-    /// # Arguments
-    ///
-    /// * `&mut self` - Write access to the contract's state.
-    ///
-    /// # Errors
-    ///
-    /// If the contract is in `Paused` state, then the error
-    /// [`Error::EnforcedPause`] is returned.
-    pub fn pause(&mut self) -> Result<(), Error> {
-        self.when_not_paused()?;
-        self._paused.set(true);
-        evm::log(Paused { account: msg::sender() });
-        Ok(())
-    }
-
-    /// Triggers `Unpaused` state.
-    ///
-    /// # Arguments
-    ///
-    /// * `&mut self` - Write access to the contract's state.
-    ///
-    /// # Errors
-    ///
-    /// If the contract is in `Unpaused` state, then the error
-    /// [`Error::ExpectedPause`] is returned.
-    pub fn unpause(&mut self) -> Result<(), Error> {
-        self.when_paused()?;
-        self._paused.set(false);
-        evm::log(Unpaused { account: msg::sender() });
-        Ok(())
-    }
-
     /// Modifier to make a function callable only when the contract is NOT
     /// paused.
     ///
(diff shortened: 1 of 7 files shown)


contracts/src/utils/pausable.rs
```diff
@@ -9,6 +9,10 @@
 //!
 //! Note that they will not be pausable by simply including this module,
 //! only once the modifiers are put in place.
+//!
+//! Note that [`Pausable::pause`] and [`Pausable::unpause`] methods are not
+//! exposed by default.
+//! You should expose them manually in your contract's abi.
```
bidzyyys, line 15:
```suggestion
//! You should expose them manually in your contract.
```

examples/erc721/src/lib.rs
```diff
@@ -151,4 +151,12 @@ impl Erc721Example {
 
         Ok(())
     }
+
+    pub fn pause(&mut self) -> Result<(), Error> {
```
bidzyyys, line 155: `Error` is inconsistent with other functions.

examples/erc721/src/lib.rs
```diff
@@ -151,4 +151,12 @@ impl Erc721Example {
...
+        self.pausable.pause()
+    }
+
+    pub fn unpause(&mut self) -> Result<(), Error> {
```
bidzyyys, line 159: `Error` is inconsistent with other functions.

examples/erc20/src/lib.rs
```diff
@@ -105,4 +105,12 @@ impl Erc20Example {
         self.pausable.when_not_paused()?;
         self.erc20.transfer_from(from, to, value).map_err(|e| e.into())
     }
+
+    pub fn pause(&mut self) -> Result<(), Error> {
+        self.pausable.pause()
+    }
+
+    pub fn unpause(&mut self) -> Result<(), Error> {
```
bidzyyys, line 113: `Error` is inconsistent with other functions.

examples/erc20/src/lib.rs
```diff
@@ -105,4 +105,12 @@ impl Erc20Example {
         self.pausable.when_not_paused()?;
         self.erc20.transfer_from(from, to, value).map_err(|e| e.into())
     }
+
+    pub fn pause(&mut self) -> Result<(), Error> {
```
bidzyyys, line 109: `Error` is inconsistent with other functions.

bidzyyys (commented):
Left some comments @qalisander

examples/erc721/src/lib.rs
qalisander: I think it is fine. One less conversion to `Vec<u8>` in code.
cc @ggonzalez94

contracts/src/utils/pausable.rs
qalisander: Isn't previous way more clear for novice library user?
wdt @ggonzalez94 ?

contracts/src/utils/pausable.rs
```diff
@@ -68,73 +72,75 @@ impl Pausable {
...
     /// # Errors
     ///
-    /// If the contract is in `Paused` state, then the error
+    /// If the contract is in the `Paused` state, then the error
     /// [`Error::EnforcedPause`] is returned.
-    pub fn pause(&mut self) -> Result<(), Error> {
-        self.when_not_paused()?;
-        self._paused.set(true);
-        evm::log(Paused { account: msg::sender() });
+    pub fn when_not_paused(&self) -> Result<(), Error> {
```
ggonzalez94, line 86: I know we already had this as part of the contract public abi, but do we want this? It seems like it would be better to move it and keep it internal to the contract.

contracts/src/utils/pausable.rs
```diff
@@ -68,73 +72,75 @@ impl Pausable {
...
     ///
     /// # Errors
     ///
     /// If the contract is in `Unpaused` state, then the error
     /// [`Error::ExpectedPause`] is returned.
-    pub fn unpause(&mut self) -> Result<(), Error> {
-        self.when_paused()?;
-        self._paused.set(false);
-        evm::log(Unpaused { account: msg::sender() });
+    pub fn when_paused(&self) -> Result<(), Error> {
```
ggonzalez94, line 104: Same question as above.
I know we already had this as part of the contract public abi, but do we want this? It seems like it would be better to move it and keep it internal to the contract.

ggonzalez94 (commented):
Looks good overall! Left one small comment and a questions that came to my mind as I was reviewing about the visibility of the "modifiers"

contracts/src/utils/pausable.rs
ggonzalez94: I also prefer to keep the explicit word abi.

examples/erc721/src/lib.rs
ggonzalez94: I don't like `Vec<u8>`, but for consistency inside this example I would stick with it. It looks rare and might be confusing if inside the same file there are some functions that return `Result<(), Error> ` and others that return `Result<(), Vec<u8>>`even if they are equivalent.
We can do a refactor of all return types later as part of this [issue](https://github.com/OpenZeppelin/rust-contracts-stylus/issues/282). WDYT?

bidzyyys: I agree on refactoring later.

contracts/src/utils/pausable.rs
bidzyyys: It's a good question. IMHO modifiers do not need to be callable so I would move them from `#[public]` impl block to the second one.

contracts/src/utils/pausable.rs
```diff
@@ -68,73 +72,75 @@ impl Pausable {
         self._paused.get()
     }
 
-    /// Triggers `Paused` state.
+    /// Modifier to make a function callable only when the contract is NOT
```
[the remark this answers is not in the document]
qalisander: this should be better: "Returns error when the contract is NOT paused."

contracts/src/utils/pausable.rs
qalisander: Made them private and updated tests

ggonzalez94 (approved):
Much cleaner interface, love it.

bidzyyys (approved):
LGTM!

PR #321, opened 2024-10-04
Avoid unnecessary swap operation

### LINKED ISSUE #317: [Audit]: N-04 Unnecessary Swap Operation

The `commutative_hash_pair` function in hash.rs performs a comparison of a and b values, swapping them if `a > b` However, it would be more efficient and readable to simply reverse the order of arguments in the call to hash_pair if the comparison is true.

To improve code readability, consider modifying the `commutative_hash_pair` function to implement the aforementioned optimization.

### Contribution Guidelines

ggonzalez94:
Resolves #317

### DIFF
--- a/lib/crypto/src/hash.rs
+++ b/lib/crypto/src/hash.rs
@@ -142,16 +142,16 @@ where
 /// Sort the pair `(a, b)` and hash the result with `state`. Frequently used
 /// when working with merkle proofs.
 #[inline]
-pub fn commutative_hash_pair<S, H>(mut a: H, mut b: H, state: S) -> S::Output
+pub fn commutative_hash_pair<S, H>(a: &H, b: &H, state: S) -> S::Output
 where
     H: Hash + PartialOrd,
     S: Hasher,
 {
     if a > b {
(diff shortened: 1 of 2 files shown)


lib/crypto/src/hash.rs
```diff
@@ -142,16 +142,16 @@ where
...
     H: Hash + PartialOrd,
     S: Hasher,
 {
     if a > b {
-        core::mem::swap(&mut a, &mut b);
+        return hash_pair(b, a, state);
     }
 
-    hash_pair(&a, &b, state)
+    hash_pair(a, b, state)
```
bidzyyys, line 154: Nit: I would refactor to:
```
if a > b {
  hash_pair(b, a, state)
}
else {
  hash_pair(a, b, state)
}
```

bidzyyys (commented):
Left small comment.

lib/crypto/src/hash.rs
bidzyyys: To make it more redeable

bidzyyys (approved):
LGTM!

PR #322, opened 2024-10-04
fix(N-05) improve comments for modifiers

### LINKED ISSUE #318: [Audit]:N-05 Misleading Reference to Modifiers in Pausable Module

The Pausable module makes reference to modifiers before the when_not_paused and when_paused functions, but true modifiers are not supported by the Stylus SDK. Consider
changing the wording of these comments and any others in the codebase to ensure that there are references to features that are unique to Solidity.

### Contribution Guidelines

ggonzalez94:
Resolves #318 

I used the term `helpers`, but open to better wording if someone has suggestions. This touches the same file as #311, so we might need to update it after we merge that one.

### DIFF
--- a/contracts/src/utils/pausable.rs
+++ b/contracts/src/utils/pausable.rs
@@ -7,8 +7,8 @@
 //! and [`Pausable::when_paused`],
 //! which can be added to the functions of your contract.
 //!
-//! Note that your contract will not be pausable by simply including this module,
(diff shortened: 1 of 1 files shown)


bidzyyys (approved):
I like term `helper`.

qalisander (approved):
Looks good!

PR #323, opened 2024-10-04
fix(L-01): Incomplete Implementation for checkpoints.rs

qalisander:
Extends checkpoints implementation without code duplications, albeit it has a bit complex generics


contracts/src/utils/structs/checkpoints/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+    /// Zero value of the number.
+    const ZERO: Self;
+}
+
+impl<const B: usize, const L: usize> Num for Uint<B, L> {
+    const ZERO: Self = Self::ZERO;
+}
+
+/// Abstracts accessor inside the checkpoint contract
+pub trait Accessor {
```
qalisander, line 74: IMHO names names are not best for that kind of abstractions. But it is hard to come up with smth else for me

contracts/src/utils/structs/checkpoints/generic_size.rs
```diff
@@ -0,0 +1,89 @@
+//! Contains generic size utilities for checkpoint storage contract.
+
+use core::ops::{Add, Div, Mul, Sub};
+
+use alloy_primitives::Uint;
+use stylus_sdk::prelude::*;
```
bidzyyys, line 6: Import only what we need.

contracts/src/utils/structs/checkpoints/generic_size.rs
```diff
@@ -0,0 +1,89 @@
...
+    fn get(&self) -> Self::Wraps;
+
+    /// Sets underlying element [`Self::Wraps`] in persistent storage.
+    fn set(&mut self, value: Self::Wraps);
+}
+
+impl<const B: usize, const L: usize> Accessor
+    for stylus_sdk::storage::StorageUint<B, L>
+{
+    type Wraps = Uint<B, L>;
```
bidzyyys, line 80: Nit: I do not like `Wraps`, I would remove.

bidzyyys: Or at least make it more self-descripting

contracts/src/utils/structs/checkpoints/generic_size.rs
```diff
@@ -0,0 +1,89 @@
...
+
+impl Size for S208 {
+    type Key = <Self::KeyStorage as Accessor>::Wraps;
+    type KeyStorage = stylus_sdk::storage::StorageUint<48, 1>;
+    type Value = <Self::ValueStorage as Accessor>::Wraps;
+    type ValueStorage = stylus_sdk::storage::StorageUint<208, 4>;
+}
+
+/// Abstracts number inside the checkpoint contract.
+pub trait Num: Add + Sub + Mul + Div + Ord + Sized + Copy {
```
bidzyyys, line 56: I think the name should be more self-descripting.

contracts/src/utils/structs/checkpoints/generic_size.rs
bidzyyys, line 1: I would add macros for `Size` generation similar to [these](https://docs.rs/ruint/1.12.3/src/ruint/aliases.rs.html#51-63)

bidzyyys (commented):
Left some comments.

contracts/src/utils/structs/checkpoints/generic_size.rs
qalisander: `StorageType::Wraps` has the same name for associated abi type. What other ideas you have?

contracts/src/utils/structs/checkpoints/generic_size.rs
qalisander: Not ideal, but what you think is better? `AbiNum`?

contracts/src/utils/structs/checkpoints/generic_size.rs
```diff
@@ -0,0 +1,89 @@
...
+    /// Zero value of the number.
+    const ZERO: Self;
+}
+
+impl<const B: usize, const L: usize> Num for Uint<B, L> {
+    const ZERO: Self = Self::ZERO;
+}
+
+/// Abstracts accessor inside the checkpoint contract.
+pub trait Accessor {
```
qalisander, line 66: Think about having `Accessor` trait private

contracts/src/utils/structs/checkpoints/generic_size.rs
qalisander, line 56: Think about having this trait private

contracts/src/utils/structs/checkpoints/generic_size.rs
qalisander: What's the purpose of `prelude` module in the crate then? It should be meant to auto export all basic things. To not spell them all over again. If auto exported name collides with one of declared within your module, then export them explicitly.
Otherwise I see usage of `::prelude::*` more helpful and less polluting the code-base.
@ggonzalez94 wdt?

contracts/src/utils/structs/checkpoints/generic_size.rs
qalisander: I've made this trait `pub(crate)`. Naming shouldn't be that important

contracts/src/utils/structs/checkpoints/generic_size.rs
qalisander: I've made this trait `pub(crate)`. Naming shouldn't be that important

bidzyyys (approved):
LGTM!

PR #329, opened 2024-10-09
fix(L-05): remove redundant error from ECDSA smart contract

### LINKED ISSUE #328: [Audit]: L-05 Missing Signature Length Check

The ECDSA library does not utilize the  `ECDSAInvalidSignatureLength`  error. According to
the Solidity library, the concatenation of `v` ,  `r` , and  `s`  must have a valid signature length of 65
bytes, otherwise, an error is returned. This validation is missing in the  `_recover`  function
within  ecdsa.rs , where the signature length is not checked.
Consider adding this validation check for the correct signature length in the  `_recover`
function. In addition, consider correcting the misplaced comment referencing this error in line
189.

### Contribution Guidelines

bidzyyys:
Resolves #328

### DIFF
--- a/contracts/src/utils/cryptography/ecdsa.rs
+++ b/contracts/src/utils/cryptography/ecdsa.rs
@@ -30,13 +30,6 @@ sol! {
     #[allow(missing_docs)]
     error ECDSAInvalidSignature();
 
-    /// The signature has an invalid length.
-    ///
-    /// * `length` - Length of the signature.
-    #[derive(Debug)]
-    #[allow(missing_docs)]
-    error ECDSAInvalidSignatureLength(uint256 length);
-
     /// The signature has an `S` value that is in the upper half order.
     ///
     /// * `s` - Invalid `S` value.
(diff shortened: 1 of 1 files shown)


ggonzalez94 (approved):

contracts/src/utils/cryptography/ecdsa.rs
```diff
@@ -186,7 +177,7 @@ fn encode_calldata(hash: B256, v: u8, r: B256, s: B256) -> Vec<u8> {
 /// and for v in (302): v ∈ {27, 28}.
 ///
 /// Most signatures from current libraries generate a unique signature
-/// with an s-value in the lower half order.ECDSAInvalidSignatureLength
+/// with an s-value in the lower half order.
```
qalisander, line 180: isn't here slightly different meaning?

qalisander: I see It wasn't grammatically correct before..

qalisander (approved):
LGTM

contracts/src/utils/cryptography/ecdsa.rs
bidzyyys: `ECDSAInvalidSignatureLength` should not be placed there 😅

ggonzalez94 (approved):

PR #330, opened 2024-10-09
fix(N-01) Remove TODOs and FixMes

### LINKED ISSUE #315: [Audit]: N-01 Unaddressed TODOs and FIXMEs

Throughout the codebase, multiple instances of TODO and FIXME comments were identified:
 - line59 of ERC-20 extension metadata.rs
- line73 of ERC-20 extension enumerable.rs
- line11 of ERC-721 extension enumerable.rs
- line71 of ERC-721 extension mod.rs
 - line486 of ERC-721 extension mod.rs
 - line665 of ERC-721 extension mod.rs
 
 During development, having well-described TODO or FIXME comments will make the process of tracking and resolving them easier. Without this information, these comments might age and important information for the security of the system might be forgotten by the time it is released to production. As such, consider removing all instances of TODO or FIXME comments and instead tracking them in the issues backlog. Alternatively, consider linking each inline TODO or FIXME to a corresponding backlog issue.

### Contribution Guidelines

ggonzalez94:
Resolves #315 

I'm leaving comments on things that might need some discussion

### DIFF
--- a/contracts/src/token/erc20/extensions/metadata.rs
+++ b/contracts/src/token/erc20/extensions/metadata.rs
@@ -56,9 +56,6 @@ pub trait IErc20Metadata {
     fn decimals(&self) -> u8;
 }
 
-// FIXME: Apply multi-level inheritance to export Metadata's functions.
-// With the current version of SDK it is not possible.
-// See https://github.com/OffchainLabs/stylus-sdk-rs/pull/120
 #[public]
 impl IErc20Metadata for Erc20Metadata {
     fn name(&self) -> String {
@@ -70,9 +67,6 @@ impl IErc20Metadata for Erc20Metadata {
     }
 
     fn decimals(&self) -> u8 {
-        // TODO: Use `U8` an avoid the conversion once
-        // https://github.com/OffchainLabs/stylus-sdk-rs/issues/117
-        // gets resolved.
         DEFAULT_DECIMALS
     }
 }
--- a/contracts/src/token/erc721/extensions/enumerable.rs
+++ b/contracts/src/token/erc721/extensions/enumerable.rs
@@ -8,7 +8,6 @@
 //! [`super::super::Erc721::balance_of`] logic, such as `Erc721Consecutive`,
 //! interfere with enumerability and should not be used together with
 //! [`Erc721Enumerable`].
-// TODO: Add link for `Erc721Consecutive` to module docs.
 
 use alloy_primitives::{uint, Address, U256};
 use alloy_sol_types::sol;
@@ -67,8 +66,6 @@ pub trait IErc721Enumerable {
     /// The error type associated to this ERC-721 enumerable trait
     /// implementation.
     type Error: Into<alloc::vec::Vec<u8>>;
-
-    // TODO: fn supports_interface (#33)
 
     /// Returns a token ID owned by `owner` at a given `index` of its token
     /// list.
--- a/contracts/src/token/erc721/mod.rs
+++ b/contracts/src/token/erc721/mod.rs
@@ -483,8 +483,6 @@ impl IErc721 for Erc721 {
         to: Address,
         token_id: U256,
     ) -> Result<(), Error> {
-        // TODO: Once the SDK supports the conversion,
-        // use alloy_primitives::bytes!("") here.
         self.safe_transfer_from_with_data(from, to, token_id, vec![].into())
     }
 
@@ -662,9 +660,6 @@ impl Erc721 {
     /// * `&mut self` - Write access to the contract's state.
     /// * `account` - Account to increase balance.
     /// * `value` - The number of tokens to increase balance.
-    // TODO: Right now this function is pointless since it is not used.
-    // But once we will be able to override internal functions,
-    // it will make a difference.
     pub fn _increase_balance(&mut self, account: Address, value: U128) {
         self._balances.setter(account).add_assign_unchecked(U256::from(value));
     }


contracts/src/token/erc20/extensions/metadata.rs
```diff
@@ -56,9 +56,6 @@ pub trait IErc20Metadata {
     fn decimals(&self) -> u8;
 }
 
-// FIXME: Apply multi-level inheritance to export Metadata's functions.
-// With the current version of SDK it is not possible.
-// See https://github.com/OffchainLabs/stylus-sdk-rs/pull/120
```
ggonzalez94, line 61: We know multi-level inheritance is not the direction the SDK is going, and we found away around this that I think is acceptable, so I'm ok removing this without creating an issue

contracts/src/token/erc20/extensions/metadata.rs
```diff
@@ -70,9 +67,6 @@ impl IErc20Metadata for Erc20Metadata {
     }
 
     fn decimals(&self) -> u8 {
-        // TODO: Use `U8` an avoid the conversion once
-        // https://github.com/OffchainLabs/stylus-sdk-rs/issues/117
-        // gets resolved.
```
ggonzalez94, line 75: This was fixed, so we could change it, but do we actually want this?
- Since this is a view function and not intended to be put in storage, don't see a compelling reason to do so.
- It would involve removing the constant for `DEFAULT_DECIMALS` or converting in place to U8

contracts/src/token/erc721/extensions/enumerable.rs
```diff
@@ -8,7 +8,6 @@
 //! [`super::super::Erc721::balance_of`] logic, such as `Erc721Consecutive`,
 //! interfere with enumerability and should not be used together with
 //! [`Erc721Enumerable`].
-// TODO: Add link for `Erc721Consecutive` to module docs.
```
ggonzalez94, line 11: I think we already did this right?

contracts/src/token/erc721/mod.rs
```diff
@@ -483,8 +483,6 @@ impl IErc721 for Erc721 {
         to: Address,
         token_id: U256,
     ) -> Result<(), Error> {
-        // TODO: Once the SDK supports the conversion,
-        // use alloy_primitives::bytes!("") here.
```
ggonzalez94, line 487: Is it actually significantly better that it's worth tracking? I think the way it is today is probably ok, but happy to hear thoughts if I'm missing something(e.g. gas saving)

contracts/src/token/erc721/mod.rs
```diff
@@ -662,9 +660,6 @@ impl Erc721 {
     /// * `&mut self` - Write access to the contract's state.
     /// * `account` - Account to increase balance.
     /// * `value` - The number of tokens to increase balance.
-    // TODO: Right now this function is pointless since it is not used.
-    // But once we will be able to override internal functions,
-    // it will make a difference.
```
ggonzalez94, line 667: This is probably the one that needs more discussion, but:
A) Internal function overriding is not very "rusty", so probably won't happen?
B) It is still useful for people writing extensions and being able to call this right? I think that's actually the way it is designed to work in Solidity

contracts/src/token/erc721/extensions/enumerable.rs
```diff
@@ -68,8 +67,6 @@ pub trait IErc721Enumerable {
     /// implementation.
     type Error: Into<alloc::vec::Vec<u8>>;
 
-    // TODO: fn supports_interface (#33)
-
```
ggonzalez94, line 72: Should be solved once we merge ERC-165, so no need to track it AFAIK

contracts/src/token/erc721/mod.rs
bidzyyys: A) 💯 
B) I cannot see big value in this function as it does not provide anything special but an additional "setter". IMHO it is fine to remove this fn.

contracts/src/token/erc20/extensions/metadata.rs
bidzyyys: IMHO yes.

contracts/src/token/erc20/extensions/metadata.rs
bidzyyys: Do not see any valid reason tbh.

contracts/src/token/erc721/mod.rs
bidzyyys: I do not think so.

bidzyyys (approved):

contracts/src/token/erc721/mod.rs
qalisander: Not significantly better. I suspect would be zero difference in gas:) Do we have a zero todo's policy in code?:D

contracts/src/token/erc721/mod.rs
qalisander: Basically it is used at consecutive now. In solidity it is impossible to access private fields from child contracts. At stylus it can work, but I feel it's a bad development practice.
As I see the function is just for convenience.
We can remove misleading `TODO`.

contracts/src/token/erc721/mod.rs
ggonzalez94: No, I don't think we should aim for that. Just maybe before each audit we make sure to clean things and move outstanding TODOs to issues? Since we know otherwise they will become findings in the report

PR #309, opened 2024-10-04
fix(M-01):  Erc165 standard support

qalisander:
Resolves: #319

#### PR Checklist

- [x] Tests
- [x] Documentation

PR #331, opened 2024-10-09
fix(L-02): Inconsistent Implementations in ERC-721 Extension

### LINKED ISSUE #312: [Audit]: L-02 Inconsistent Implementations in ERC-721 Extension

The Stylus library ERC-721 extensions are not consistent with the respective functionality of the Solidity library.
These inconsistencies were identified:
• In the uri_storage extension, the token_uri() function directly returns the value from storage but omits key checks:
◦ It does not verify whether the token exists, which is handled in the Solidity implementation.
◦ It does not handle base URI concatenation like the Solidity version. Instead, it directly retrieves and returns the stored token URI without constructing the full URI.
• There is a mismatch in the implementation of ERC721Metadata :
◦ The Rust implementation exposes the base_uri function. In the Solidity version, URIs are only accessible by token ID when the token exists. Not exposing
baseURI is generally advisable, as revealing it prematurely can leak sensitive NFT data before minting. This can enable advanced users to time their minting to acquire more desirable NFTs.
◦ The Rust version lacks the tokenURI function by default, which may cause integration issues.
Consider aligning these implementations with the Solidity library to ensure consistency and mitigate potential risks.

### Contribution Guidelines

qalisander:
These changes are based on ERC-165 (#309). Should be merged after that pr.
Resolves #312


examples/erc721-consecutive/src/constructor.sol
```diff
@@ -2,6 +2,10 @@
 pragma solidity ^0.8.21;
 
 contract Erc721ConsecutiveExample {
+    string private _name;
```
qalisander, line 5: Changes to solidity constructor bring an error from `koba`. Without them and without additional fields inside `Erc721`, consecutive extension's tests can pass.

bidzyyys:
I believe it has been solved in M1.

PR #333, opened 2024-10-10
fix(L-02): Inconsistent Implementations in ERC-721 Extension #331

### LINKED ISSUE #312: [Audit]: L-02 Inconsistent Implementations in ERC-721 Extension

bidzyyys:
Resolves #312

### DIFF
--- a/contracts/src/token/erc721/extensions/metadata.rs
+++ b/contracts/src/token/erc721/extensions/metadata.rs
@@ -90,7 +90,7 @@ impl Erc721Metadata {
         token_id: U256,
         erc721: &impl IErc721<Error = Error>,
     ) -> Result<String, Error> {
-        let _owner = erc721.owner_of(token_id)?;
+        erc721.owner_of(token_id)?;
 
         let base_uri = self.base_uri();
 
--- a/contracts/src/token/erc721/extensions/metadata.rs
+++ b/contracts/src/token/erc721/extensions/metadata.rs
@@ -1,10 +1,14 @@
 //! Optional Metadata of the ERC-721 standard.
 
-use alloc::string::String;
+use alloc::string::{String, ToString};
 
+use alloy_primitives::U256;
 use stylus_proc::{public, sol_storage};
 
-use crate::utils::Metadata;
+use crate::{
+    token::erc721::{Error, IErc721},
+    utils::Metadata,
+};
 
 sol_storage! {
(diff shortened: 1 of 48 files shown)


contracts/src/token/erc721/extensions/metadata.rs
```diff
@@ -53,8 +49,57 @@ impl IErc721Metadata for Erc721Metadata {
     fn symbol(&self) -> String {
         self._metadata.symbol()
     }
+}
 
-    fn base_uri(&self) -> String {
```
qalisander, line 57: In this case `#[interface_id]` won't be that helpful. Because methods inside `IErc721Metadata` trait are not same as in solidity (`token_uri` is missing). So computed interface id will be different..

ggonzalez94: If this is the only thing preventing this PR I would say we don't expose `#[interface_id]` for the metadata extension(since I don't think it is part of the EIP itself).
One thing we should do is add in the Anotra docs or at the top of the contract that people should expose `token_uri` manually(I know it is on the internal function description, but still)

ggonzalez94 (commented):
Looks good overall, I like using the `impl` as a parameter to the function(keeps things more clean). Let's finalize the discussion around `ERC165` support and if we need to add a note to the docs

ggonzalez94 (approved):
LGTM

qalisander (approved):
LGTM

PR #339, opened 2024-10-15
336 audit n 03 inconsistent error types

### LINKED ISSUE #336: [Audit]: N-02 Inconsistent Error Types

Throughout the codebase, there are some extensions that use associated error types to allow
users the flexibility to assign a generic  Vec  as an error type and encode any error that is
needed (or not return an error at all), while others do not use this error style. For example, in
the [IERC20 trait of the ERC20 implementation](https://github.com/OpenZeppelin/rust-contracts-stylus/blob/42c859f37d7332fbd78a216a283064a5227f9c68/contracts/src/token/erc20/mod.rs#L116).
Consider using a consistent error style to avoid confusion and improve code clarity.

### Contribution Guidelines

bidzyyys:
Resolves #336

### DIFF
--- a/.github/workflows/check-publish.yml
+++ b/.github/workflows/check-publish.yml
@@ -1,0 +1,42 @@
+name: check-publish
+# This workflow checks that the libraries can be published on crates.io.
+permissions:
+  contents: read
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+env:
+  CARGO_TERM_COLOR: always
+jobs:
+  check-publish:
+    name: Check publish on crates.io
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: set up rust
+        uses: dtolnay/rust-toolchain@master
+        id: toolchain
+        with:
+          target: wasm32-unknown-unknown
+          components: rust-src
+          toolchain: nightly-2024-01-01
+
+      - uses: Swatinem/rust-cache@v2
+
+      - name: check motsu-proc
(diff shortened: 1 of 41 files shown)


contracts/src/utils/structs/bitmap.rs
```diff
@@ -27,66 +27,90 @@ sol_storage! {
     }
 }
 
-impl BitMap {
+/// Required interface of a [`BitMap`] compliant contract.
+pub trait IBitMap {
```
qalisander, line 31: Do you think we need that kind of abstraction for `BitMap` and `Trace`?
I considering them mostly as primitives not separate contracts..
cc @ggonzalez94

bidzyyys: I would say that everything under `contracts` should have trait.

qalisander: May every function that is assumed to be publicly exposed should be in the trait? (public in solidity) In specified "contracts" there is no function that is assumed to be exposed publicly.

bidzyyys: Curious about @ggonzalez94 thoughts.

ggonzalez94: I don't have a strong opinion here.... I don't think many people would use the trait and will only be interested in the implementation, but I guess for consistency it doesn't hurt to have them

qalisander: Since OL shared that they're going to stick to trait based approach.
And that in future we are expecting being able to declare `#[public]` macro more then once.
image
So this pattern can be done for every trait we have. That is why I'm considering to put just public methods into it.

bidzyyys:
New attitude should be applied with Stylus SDK `v0.8`

PR #373, opened 2024-10-21
Use mini-alloc by default and remove panic handler

ggonzalez94:
Mi intention is not that we directly merge this PR, but that it serves 2 purposes
1. Discuss how we want to handle global allocator and mainly what to do about the panic handler
2. In case we proceed wether we want to merge it to the release branch(and issue a new release) or just to main

## Background
While testing how the experience will be for someone that creates a new stylus project by running `cargo stylus new` I realized a couple things that I think hurt DX.
1. You have to disable the default feature flags on the stylus sdk(which disables mini_alloc as the default globall allocator). Otherwise you get #global_allocator conflicts as it is defined in two places(the stylus sdk and our library). 
``` toml
stylus-sdk = { version = "=0.6.0", default-features = false }
```
2. Since our library is `#[no_std]` by default it defines a custom panic_handler such that if you use it from a `no_std `environment you don’t have to define it yourself. The problem is that if you want to use the std library in your project(and that seems to be a reasonable assumption according to [this](https://github.com/OffchainLabs/stylus-sdk-rs/tree/main?tab=readme-ov-file#feature-highlights)), you cannot do it since it results in a duplication of the panic_handler function. This is also the default behavior when you create a new project with `cargo stylus new`.

## Proposed Solution
1. Using mini-alloc by default via the sdk seems like the best approach as it was one of the recommendations from the [audit](https://forum.arbitrum.foundation/t/stylus-sdk-audit/26674#p-58099-inefficient-allocator-fallback-in-stylus-contracts-44) and is what is used when creating a new project via `cargo stylus new`.
2. Here I see two options:
    - Add a compilation flag to the panic handler such that it is not included if built with std flag on. This is the approach I took in this PR, but it has 2 downsides:
        1. If you just start a project from scratch using `cargo stylus new` it is harder to realize and you might spend some time trying to figure it out.
        2. We are now encouraging people to use the `std` flag on our library, which might enable other behaviors we haven’t considered? Since we’ve been mainly using it for testing
     - Remove the panic handler entirely from the library, allowing projects building with std to have it included by default and projects explicitly opting in to `no_std` and having zero dependencies on the std library would have to define it themselves.
This approach is more reasonable IMO, since it makes it easier for newcomers, but it requires some extra effort if you are experienced enough to define your contracts with `no_std` .
The problem today is that we set the library type to [cdylib](https://doc.rust-lang.org/reference/linkage.html), which requires the crate to define a panic_handler when using `no_std` (more info [here](https://internals.rust-lang.org/t/panic-handler-free-no-std-targets/14697)). If we find a way to do this, maybe this is the best approach, maybe it is possible using unstable features.

## Edit
- We were able to remove the panic handler entirely by not targeting `cdylib`(thanks @qalisander )! Now the user of the library can just use `std`or define their own panic handler.
- [x] Test on a new stylus project
- [x] Update the Readme  
- [x] Update Changelog

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -63,7 +63,7 @@ all = "warn"
 
 [workspace.dependencies]
 # stylus-related
-stylus-sdk = { version = "=0.6.0", default-features = false }
+stylus-sdk = "0.6.0"
 stylus-proc = { version = "=0.6.0", default-features = false }
 mini-alloc = "0.4.2"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,7 +64,7 @@ all = "warn"
 [workspace.dependencies]
 # stylus-related
 stylus-sdk = "0.6.0"
-stylus-proc = { version = "=0.6.0", default-features = false }
+stylus-proc = "0.6.0"
 
 alloy = { version = "=0.1.4", features = [
   "contract",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,7 +65,6 @@ all = "warn"
 # stylus-related
 stylus-sdk = "0.6.0"
 stylus-proc = { version = "=0.6.0", default-features = false }
-mini-alloc = "0.4.2"
 
 alloy = { version = "=0.1.4", features = [
   "contract",
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,5 +1,5 @@
 [target.wasm32-unknown-unknown]
-rustflags = ["-C", "link-arg=-zstack-size=8192"]
+rustflags = ["-C", "link-arg=-zstack-size=8192", "-C", "target-cpu=mvp"]
 
 [target.aarch64-apple-darwin]
 rustflags = ["-C", "link-arg=-undefined", "-C", "link-arg=dynamic_lookup"]
--- a/.github/workflows/check-publish.yml
+++ b/.github/workflows/check-publish.yml
@@ -18,27 +18,22 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: set up rust
-        uses: dtolnay/rust-toolchain@master
-        id: toolchain
+      - name: Install rust
+        uses: actions-rust-lang/setup-rust-toolchain@v1
         with:
-          target: wasm32-unknown-unknown
-          components: rust-src
-          toolchain: nightly-2024-01-01
+          rustflags: ""
 
-      - uses: Swatinem/rust-cache@v2
-
-      - name: check motsu-proc
+      - name: Check motsu-proc
         run: cargo publish -p motsu-proc --dry-run
 
-      - name: check motsu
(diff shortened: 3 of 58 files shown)


qalisander:
Seems this is similar [pr](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/372). We could discuss it there.
Do we want to merge it with `0.1.0` or `main`?

Cargo.toml
```diff
@@ -63,7 +63,7 @@ all = "warn"
 
 [workspace.dependencies]
 # stylus-related
-stylus-sdk = { version = "=0.6.0", default-features = false }
+stylus-sdk = "0.6.0"
 stylus-proc = { version = "=0.6.0", default-features = false }
```
bidzyyys, line 67: Should we remove `default-features` from `stylus-proc` and `alloy` too?

bidzyyys:
> Seems this is similar to [pr](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/372). We could discuss it there. Do we want to merge it with `0.1.0` or `main`?

In terms of DevEx I think that we should prioritise `#[no_std]` environment because of e.g. reducing binary size (we know it's a big issue). I believe that once someone wants to try out Stylus, he starts with e.g. Counter example and only modifies it (without our library). BUT may want to easily add our unit tests (`motsu`). 

I believe that our library will be used by more "mature" Stylus developers that understand the SDK and need to build more advanced projects. For this I think having well documented steps for including our library to the project is enough.

For the merge I see 2 options:
1. Releasing quickly `v0.1.1` of our library with the fix
2. Merging to `main` and releasing with `v0.2.0-alpha.1`

Cargo.toml
qalisander: In this case I think we should "cherry pick" the commit, that uses `stylus-proc` indirectly through `stylus-sdk` from the `main` branch. I hope we will get less merge conflicts

qalisander:
Just updated dependencies for mini alloc. We don't need to use `0.4.2` with altogether with `0.6.0` referenced by `stylus-sdk`.
I think we can merge this pr. And publish patch altogether with this [one](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/375).

Cargo.toml
ggonzalez94: I don't think for alloy we want to do that, since it includes std by [default](https://docs.rs/crate/alloy/latest/features)

qalisander (approved):
I've checked these changes as a patch for `openzeppelin-stylus` at `stylus-debug-issue` repo. Seems it works currently (when you call `cargo clean` between changes).

I suspect that probably issue was with `0.4.2` version of `mini-alloc`.

bidzyyys (approved):
Should be merged after #375 and use `nigthly-2024-09-05`

bidzyyys (approved):
Looks fine.

PR #376, opened 2024-10-23
feat: finite fields

### LINKED ISSUE #342: [Feature]: Finite fields arithmetics

Add basic field arithmetic operations including multiplicative inverses, that will be used by other crypto primitives.
Pay attention to resulted wasm size of the crypto library.

### Contribution Guidelines

qalisander:
Resolves #342

#### PR Checklist

- [x] Remove dependencies to ark-ff
- [x] Add montgomery operations for Fp
- [x] Tests
- [x] Documentation
- [x] Remove posseidon hash from this pr
- [x] Adapt traits for wasm target with usize of 32 bits
- [x] Close all comment todo's


lib/crypto/src/bigint.rs
```diff
@@ -0,0 +1,277 @@
...
+    b: &Uint<LIMBS>,
+) -> Uint<LIMBS> {
+    let (low, high) = a.mul_wide(b);
+    assert!(high.bits() == 0, "overflow on multiplication");
+    low
+}
+
+/// Add two numbers and panic on overflow.
+#[must_use]
+pub const fn add<const LIMBS: usize>(
```
bidzyyys, line 208: I do not like this attitude; I believe you should just implement traits like here:
https://doc.rust-lang.org/std/primitive.u64.html#trait-implementations

bidzyyys: Like [Add](https://doc.rust-lang.org/std/ops/trait.Add.html) trait.

bidzyyys: And require these traits in `BigInteger`

lib/crypto/src/bigint.rs
```diff
@@ -0,0 +1,277 @@
...
+        self.as_words().iter().rev().flat_map(Word::bit_be_iter)
+    }
+}
+
+/// Parse a number from a string in a given radix.
+///
+/// I.e., convert string encoded integer `s` to base-`radix` number.
+#[must_use]
+pub const fn from_str_radix<const LIMBS: usize>(
+    s: &str,
```
bidzyyys, line 156: Why not `AsRef<str>`?

lib/crypto/src/bigint.rs
```diff
@@ -0,0 +1,277 @@
...
+    let mut uint = Uint::from_u32(0);
+    let mut order = Uint::from_u32(1);
+    let uint_radix = Uint::from_u32(radix);
+
+    loop {
+        // Try to parse a digit from utf-8 byte
+        let ch = parse_utf8_byte(bytes[index]);
+        let digit = match ch.to_digit(radix) {
+            None => {
+                panic!("invalid digit");
```
bidzyyys, line 174: I believe it should not panic but return an error, e.g. [ParseIntError](https://doc.rust-lang.org/std/num/struct.ParseIntError.html)

lib/crypto/src/bigint.rs
```diff
@@ -0,0 +1,277 @@
...
+        $crate::bigint::crypto_bigint::Uint::from_be_hex($num)
+    };
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn convert_from_str_radix() {
```
bidzyyys, line 246: I would say that these are perfect tests for fuzzing 😅

lib/crypto/src/field/fp.rs
```diff
@@ -0,0 +1,999 @@
...
+    /// The modulus of the field.
+    const MODULUS: Uint<LIMBS>;
+
+    /// A multiplicative generator of the field.
+    /// [`Self::GENERATOR`] is an element having multiplicative order
+    /// `MODULUS - 1`.
+    const GENERATOR: Fp<Self, LIMBS>;
+
+    /// Set `a += b`.
+    fn add_assign(a: &mut Fp<Self, LIMBS>, b: &Fp<Self, LIMBS>) {
```
bidzyyys, line 48: I believe you should use regular Rust traits for math operations.

bidzyyys (commented):
Great job @qalisander, left some comments for the first iteration.

lib/crypto/proptest-regressions/field/fp.txt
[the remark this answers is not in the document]
qalisander: This is for fuzz tests, to repeat failures. We have already added it in other places in our repo

lib/crypto/Cargo.toml
```diff
@@ -10,10 +10,16 @@ version.workspace = true
 
 [dependencies]
 tiny-keccak.workspace = true
+num-traits.workspace = true
+zeroize.workspace = true
+crypto-bigint.workspace = true
+educe = "0.6.0"
```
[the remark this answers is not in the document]
qalisander: That's a good question. Should we move dependencies specific even to a single crate on workspace level or not..

lib/crypto/src/bigint.rs
qalisander: The whole point of this function to have a `const` evaluation.
Const trait functions are not yet supported in rust, and here is the workaround.

lib/crypto/src/bigint.rs
qalisander: This function also have a `const` context, and we can not call a constant `AsRef::as_ref` trait method in `const` function context.
And the hole point of this imperative implementation is: the function can be evaluated at compile time.

lib/crypto/src/bigint.rs
qalisander: I think you're right. We can try to add them, but for relatively small numbers probably (up to u128) , to be able test them with convenience

lib/crypto/src/field/fp.rs
qalisander: We already have rust math operations for `Fp` types. These ones in `FpParams` just let you to override an algorithm for addition, same as in arkworks library. 
But I'm leaning towards: we don't need math operations in `FpParams` trait at this early stage. And keep it more simple..

bidzyyys (commented):
- [x] Many unresolved warnings like [here](https://github.com/OpenZeppelin/rust-contracts-stylus/actions/runs/11652176671/job/32443044690?pr=376)
- [ ] Update CHANGELOG.md

lib/crypto/Cargo.toml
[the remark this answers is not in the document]
qalisander: Moved it to workspace level

lib/crypto/src/bigint.rs
qalisander: Wdt to add it within Poseidon hash pr? And live it like that for a while

0xNeshi (approved):

lib/crypto/src/bigint.rs
```diff
@@ -0,0 +1,277 @@
...
+    let mut index = bytes.len() - 1;
+
+    let mut uint = Uint::from_u32(0);
+    let mut order = Uint::from_u32(1);
+    let uint_radix = Uint::from_u32(radix);
+
+    loop {
+        // Try to parse a digit from utf-8 byte
+        let ch = parse_utf8_byte(bytes[index]);
+        let digit = match ch.to_digit(radix) {
```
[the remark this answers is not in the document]
bidzyyys: Unfortunately we cannot :(
```sh
error: `std::option::Option::<T>::expect` is not yet stable as a const fn
   --> lib/crypto/src/bigint.rs:172:36
    |
172 |         let digit = Uint::from_u32(ch.to_digit(radix).expect("invalid digit"));
    |                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: add `#![feature(const_option)]` to the crate attributes to enable

error: could not compile `openzeppelin-crypto` (lib) due to 1 previous error
```

bidzyyys (approved):
LGTM

PR #375, opened 2024-10-21
fix: reference-types error on nightly version

qalisander:
It seems that reference types from WASM are not supported by `wasm-parser` crate (https://github.com/rustwasm/wasm-bindgen/issues/4211).
It makes us pin to the recent nightly toolchain version that works, and contribution experience gets quite convoluted.
Hopefully there is an option to disable `reference-types` with a flag: `-C target-cpu=mvp` for `wasm32-unknown-unknown` target.

Resolves #374

Single source of rust toolchain from `rust-toolchain.toml` added for ci pipelines (actions-rust-lang/setup-rust-toolchain@v1 gh action used for that).

Resolves #294

PR #377, opened 2024-10-23
remove panic handlers

qalisander:


.github/workflows/check-wasm.yml
```diff
@@ -25,7 +25,7 @@ jobs:
         with:
           target: wasm32-unknown-unknown
           components: rust-src
-          toolchain: nightly-2024-01-01
+          toolchain: nightly-2024-07-25
```
qalisander, line 28: cargo-stylus refused to compile on `nightly-2024-01-01` this time. There won't be a problem, when we will merge these changes after [pr with nightly fix](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/375)

examples/basic/token/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-#![cfg_attr(not(test), no_main, no_std)]
```
qalisander, line 1: As an option library user can add `#[no_std]` and have a `#[panic_handler]` in his contract.
For simplicity we can just opt out `#[no_std]` for examples.

qalisander: Also I've noticed zero difference in terms of wasm size of example with `#[no_std]` and without it.

PR #379, opened 2024-10-24
feat(test): parallel unit test execution

### LINKED ISSUE #306: [Feature]: Allow unit tests to run in parallel

Currently tests using motsu run serially because of global access to storage. I don't know how feasible this is with the current status of the sdk and other things, but would be great if we allow tests to run in isolation in parallel. 

### Contribution Guidelines

qalisander:
Refactoring for unit tests, that lets them run in parallel.
Every rust test has an associated thread name. Hence corresponding test mock data are stored in a [DashMap](https://docs.rs/dashmap/latest/dashmap/) (concurrent hashmap). Key is name of the thread (full name of the test). Value is mock onchain storage. In the future Value (`MockStorage`) will contain additionally fields for `msg::sender`, `msg::value` and so on..

Resolves #306


.github/workflows/check-publish.yml
```diff
@@ -24,7 +24,7 @@ jobs:
         with:
           target: wasm32-unknown-unknown
           components: rust-src
-          toolchain: nightly-2024-01-01
+          toolchain: nightly-2024-07-25
```
qalisander, line 27: update from `nightly-2024-01-01` since cargo stylus cannot compile on the previous version. Temporarily

lib/motsu/src/storage.rs
```diff
@@ -1,32 +0,0 @@
-//! Shims for storage operations.
-use std::{collections::HashMap, ptr, sync::Mutex};
```
qalisander, line 2: this one moved to the `context.rs`

lib/motsu/src/context.rs
```diff
@@ -1,34 +1,97 @@
...
+use crate::prelude::{Bytes32, WORD_BYTES};
 
-/// A global static mutex.
-///
-/// We use this for scenarios where concurrent mutation of storage is wanted.
-/// For example, when a test harness is running, this ensures each test
-/// accesses storage in an non-overlapping manner.
+/// Context of stylus unit tests associated with the current test thread.
+#[allow(clippy::module_name_repetitions)]
+pub struct TestContext {
```
qalisander, line 13: In nearest future it is possible to add to this `TestContext` struct fiels as `msg_sender`, `msg_value` and so on.

lib/motsu/src/context.rs
```diff
@@ -1,34 +1,97 @@
...
-/// Decorates a closure by running it with exclusive access to storage.
-#[allow(clippy::module_name_repetitions)]
-pub fn with_context<C: StorageType>(closure: impl FnOnce(&mut C)) {
-    let _lock = acquire_storage();
-    let mut contract = C::default();
-    closure(&mut contract);
-    reset_storage();
+/// Context of the test case.
+#[derive(Default)]
+struct TestStorage {
```
qalisander, line 81: In the nearest future it is possible to add to `TestStorage` struct fields like `msg_sender`, `msg_value` and so on for every mock data.

lib/motsu/src/context.rs
```diff
@@ -1,34 +1,98 @@
...
+use crate::prelude::{Bytes32, WORD_BYTES};
 
-/// A global static mutex.
-///
-/// We use this for scenarios where concurrent mutation of storage is wanted.
-/// For example, when a test harness is running, this ensures each test
-/// accesses storage in an non-overlapping manner.
+/// Context of stylus unit tests associated with the current test thread.
+#[allow(clippy::module_name_repetitions)]
+pub struct TestContext {
```
bidzyyys, line 13: I believe we should it rename it just to `Context`. It is in `motsu` library so it should be obvious.

lib/motsu/src/context.rs
```diff
@@ -1,34 +1,98 @@
...
-
-/// Acquires access to storage.
-pub fn acquire_storage() -> MutexGuard<'static, ()> {
-    STORAGE_MUTEX.lock().unwrap_or_else(|e| {
-        reset_storage();
-        e.into_inner()
-    })
+/// The key is the name of the test thread, and the value is the storage of the
+/// test case.
+static MOCK_STORAGE: Lazy<DashMap<TestThreadName, MockStorage>> =
```
bidzyyys, line 61: The same as https://github.com/OpenZeppelin/rust-contracts-stylus/pull/379/files#r1815499126 in terms of `STORAGE`

lib/motsu/src/context.rs
```diff
@@ -1,34 +1,98 @@
...
+    Lazy::new(DashMap::new);
+
+/// Test thread name metadata.
+#[derive(Clone, Eq, PartialEq, Hash)]
+struct TestThreadName(String);
```
bidzyyys, line 66: Nit: rename to `ThreadName`

lib/motsu-proc/src/test.rs
```diff
@@ -15,37 +15,39 @@ pub(crate) fn test(_attr: &TokenStream, input: TokenStream) -> TokenStream {
...
-            fn #fn_name() #fn_return_type {
-                let _lock = ::motsu::prelude::acquire_storage();
-                let res = #fn_block;
-                ::motsu::prelude::reset_storage();
-                res
-            }
-        }
-        .into();
+    // Currently, more than one contract per unit test is not supported.
+    if fn_args.len() > 1 {
```
bidzyyys, line 19: Nit: I would say we should require always 1 contract for simplicity -> `if fn_args.len() != 1`

bidzyyys: If we do not want to test a contract I do not see any reason for not using regular `#[test]`

lib/motsu/src/context.rs
```diff
@@ -1,34 +1,98 @@
...
-/// Decorates a closure by running it with exclusive access to storage.
-#[allow(clippy::module_name_repetitions)]
-pub fn with_context<C: StorageType>(closure: impl FnOnce(&mut C)) {
-    let _lock = acquire_storage();
-    let mut contract = C::default();
-    closure(&mut contract);
-    reset_storage();
+/// Storage for unit test's mock data.
+#[derive(Default)]
+struct MockStorage {
```
bidzyyys, line 81: Nit: I would rename into `Storage`

bidzyyys (approved):
Left some minor comments regarding naming.

lib/motsu/src/context.rs
qalisander: It is mostly a question whether `Context::current().set_msg_sender("0x15a3...")` or `TestContext::current().set_msg_sender("0x15a3...")` better.
I probably agree with you. Short version still understandable.
Curious what @ggonzalez94 think?

lib/motsu/src/context.rs
qalisander: Yeah, static var `MOCK_STORAGE` colliding with `MockStorage` struct. `STORAGE` name will make sense

lib/motsu-proc/src/test.rs
qalisander: Currently it repeats the previous behavior. If no contract is declared as an argument, it will still clear storage associated with the current test thread. In case user declared contract by himself inside test, like: `Contract::default()`. Before this refactoring it would be a bug, but currently it just drops unnecessary recourse from `STORAGE`. Idk..

lib/motsu/src/context.rs
```diff
@@ -1,34 +1,98 @@
...
+        write_bytes32(value, self.get_bytes(&key));
+    }
+
+    /// Set the value at `key` in storage to `val`.
+    pub(crate) fn set_bytes(self, key: Bytes32, val: Bytes32) {
+        let mut storage = MOCK_STORAGE.entry(self.thread_name).or_default();
+        storage.contract_data.insert(key, val);
+    }
+
+    /// Set the raw value at `key` in storage to raw `val`.
```
qalisander, line 43: rewrite comment

lib/motsu/src/context.rs
```diff
@@ -1,34 +1,98 @@
...
+        Self { thread_name: TestThreadName::current() }
+    }
+
+    /// Get the value at `key` in storage.
+    pub(crate) fn get_bytes(self, key: &Bytes32) -> Bytes32 {
+        let storage = MOCK_STORAGE.entry(self.thread_name).or_default();
+        storage.contract_data.get(key).copied().unwrap_or_default()
+    }
+
+    /// Get the raw value at raw `key` in storage.
```
qalisander, line 30: rewrite comment

lib/motsu-proc/src/test.rs
bidzyyys: Ok, let's leave as it is.

lib/motsu/src/context.rs
qalisander: I feel it would be confusing altogether with an new `STORAGE` variable. Let's keep it straight forward this time

PR #387, opened 2024-10-30
fix: contract caching for benchmarks

qalisander:
Benchmarks require cargo stylus to be installed even though we're using koba for deployment.
That is used for caching a contract.


bidzyyys (approved):
LGTM

PR #396, opened 2024-11-07
build: motsu-proc v0.2.0

bidzyyys:


Cargo.toml
```diff
@@ -113,7 +113,7 @@ openzeppelin-stylus = { path = "contracts" }
 openzeppelin-stylus-proc = { path = "contracts-proc", version = "0.1.0" }
 openzeppelin-crypto = { path = "lib/crypto" }
 motsu = { path = "lib/motsu" }
-motsu-proc = { path = "lib/motsu-proc", version = "0.1.0" }
+motsu-proc = { path = "lib/motsu-proc", version = ">=0.1, <0.3" }
```
ggonzalez94, line 116: Not a huge fan of this. Why don't we just ping it to the latest version? `0.2.0` in this case. What's the benefit of doing `version = ">=0.1, <0.3"`?

ggonzalez94 (commented):
Just left a small question

ggonzalez94 (approved):

Cargo.toml
bidzyyys: Answered on Slack.

PR #399, opened 2024-11-08
ref: apply CR changes for Ownable2Step

bidzyyys:


contracts/src/access/ownable.rs
```diff
@@ -93,7 +93,10 @@ pub trait IOwnable {
     /// # Events
     ///
     /// Emits a [`OwnershipTransferred`] event.
-    fn transfer_ownership(&mut self, new_owner: Address) -> Result<(), Error>;
+    fn transfer_ownership(
+        &mut self,
+        new_owner: Address,
+    ) -> Result<(), Self::Error>;
```
ggonzalez94, line 99: What is the need/benefit from doing `Self::Error` here vs just `Error`?

examples/ownable-two-step/tests/ownable_two_step.rs
```diff
@@ -151,6 +153,113 @@ async fn accept_ownership(alice: Account, bob: Account) -> Result<()> {
     Ok(())
 }
 
+#[e2e::test]
+async fn cancel_transfer_ownership(alice: Account, bob: Account) -> Result<()> {
```
ggonzalez94, line 157: I think this name is better. I don't think we have a written rule for how to name tests yet, but personally I prefer to prefix them with the name of the function. And this is the convention I followed on this file
```suggestion
async fn transfer_ownership_cancel_transfer(alice: Account, bob: Account) -> Result<()> {
```

examples/ownable-two-step/tests/ownable_two_step.rs
```diff
@@ -151,6 +153,113 @@ async fn accept_ownership(alice: Account, bob: Account) -> Result<()> {
...
+    });
+
+    let Ownable2Step::ownerReturn { owner } = contract.owner().call().await?;
+    assert_eq!(owner, alice_addr);
+
+    let Ownable2Step::pendingOwnerReturn { pendingOwner } =
+        contract.pendingOwner().call().await?;
+    assert_eq!(pendingOwner, Address::ZERO);
+
+    Ok(())
```
ggonzalez94, line 198: I think this test is more concise and gets to the point. WDYT?

```suggestion
    let alice_addr = alice.address();
    let bob_addr = bob.address();

    let contract_addr = alice
        .as_deployer()
        .with_constructor(ctr(alice_addr))
        .deploy()
        .await?
        .address()?;
    let contract = Ownable2Step::new(contract_addr, &alice.wallet);
    receipt!(contract.transferOwnership(bob_addr))?;

    let receipt = receipt!(contract.transferOwnership(Address::ZERO))?;
    assert!(receipt.emits(OwnershipTransferStarted {
        previousOwner: alice_addr,
        newOwner: Address::ZERO,
    }));

    let Ownable2Step::pendingOwnerReturn { pendingOwner } =
        contract.pendingOwner().call().await?;
    assert_eq!(pendingOwner, Address::ZERO);

    Ok(())
```

ggonzalez94 (commented):
Thanks for tackling the comments. Left a couple suggestions and a question

contracts/src/access/ownable.rs
bidzyyys: We should use error type that is oriented around a trait.

ggonzalez94 (approved):
LGTM!

PR #402, opened 2024-11-11
feat: Add VestingWallet

### LINKED ISSUE #348: [Feature]: Add VestingWallet

Solidity reference: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/finance/VestingWallet.sol

We should still decide if we want to include extensions for cliff or other curves.

### Contribution Guidelines

0xNeshi:
Resolves #348 

#### PR Checklist

- [x] Unit Tests
- [x] E2E Tests
- [x] Documentation


0xNeshi:
Missing a couple of edge case e2e tests, but otherwise the PR is ready for review

contracts/src/finance/mod.rs
```diff
@@ -0,0 +1,2 @@
+//! Primitives for financial systems
```
bidzyyys, line 1:
```suggestion
//! Primitives for financial systems.
```

contracts/src/finance/vesting_wallet.rs
```diff
@@ -0,0 +1,617 @@
...
+        );
+    }
+
+    #[motsu::test]
+    fn gets_vesting_schedule_zero_duration(contract: VestingWallet) {
+        let start = start();
+        contract._start.set(start);
+        contract._duration.set(U64::ZERO);
+
+        let one = uint!(1_U256);
```
bidzyyys, line 607: Unused variable `one`.

contracts/src/finance/vesting_wallet.rs
```diff
@@ -0,0 +1,617 @@
...
+};
+
+sol! {
+    /// Emitted when `amount` of ether has been released.
+    ///
+    /// * `amount` - Total ether released.
+    #[allow(missing_docs)]
+    event EtherReleased(uint256 amount);
+
+    /// Emitted when `amount` of ERC20 `token` has been released.
```
bidzyyys, line 49:
```suggestion
    /// Emitted when `amount` of ERC-20 `token` has been released.
```

contracts/src/finance/vesting_wallet.rs
```diff
@@ -0,0 +1,617 @@
...
+    event ERC20Released(address indexed token, uint256 amount);
+}
+
+sol! {
+    /// Indicates an error related to the underlying Ether transfer.
+    #[derive(Debug)]
+    #[allow(missing_docs)]
+    error ReleaseEtherFailed();
+
+    /// Indicates an error related to the underlying ERC20 transfer.
```
bidzyyys, line 63:
```suggestion
    /// Indicates an error related to the underlying ERC-20 transfer.
```

contracts/src/finance/vesting_wallet.rs
```diff
@@ -0,0 +1,617 @@
...
+
+    /// The token address is not valid. (eg. `Address::ZERO`)
+    ///
+    /// * `token` - Address of the token being released.
+    #[derive(Debug)]
+    #[allow(missing_docs)]
+    error InvalidToken(address token);
+}
+
+// TODO: use existing IErc20
```
bidzyyys, line 78: Unresolved? TODO

contracts/src/finance/vesting_wallet.rs
```diff
@@ -0,0 +1,617 @@
...
+    /// The token address is not valid. (eg. `Address::ZERO`)
+    InvalidToken(InvalidToken),
+}
+
+sol_storage! {
+    /// State of the [`VestingWallet`] Contract.
+    pub struct VestingWallet {
+        /// Ownable contract
+        Ownable ownable;
+        /// Amount of eth already released.
```
bidzyyys, line 111: Make sure to have consistent naming -- you have `ether`/`Ether`/`eth`

contracts/src/finance/vesting_wallet.rs
```diff
@@ -0,0 +1,617 @@
...
+        uint256 _released;
+        /// Amount of ERC20 tokens already released.
```
bidzyyys, line 113:
```suggestion
        /// Amount of ERC-20 tokens already released.
```

contracts/src/finance/vesting_wallet.rs
```diff
@@ -0,0 +1,617 @@
...
+        mapping(address => uint256) _erc20_released;
+        /// Start timestamp.
+        uint64 _start;
+        /// Vesting duration.
+        uint64 _duration;
+        /// SafeErc20 contract
```
bidzyyys, line 119:
```suggestion
        /// [`SafeErc20`] contract.
```

contracts/src/finance/vesting_wallet.rs
```diff
@@ -0,0 +1,617 @@
...
+    /// Indicates an error related to the underlying [`IErc20`] transfer.
+    ReleaseTokenFailed(ReleaseTokenFailed),
+    /// The token address is not valid. (eg. `Address::ZERO`)
+    InvalidToken(InvalidToken),
+}
+
+sol_storage! {
+    /// State of the [`VestingWallet`] Contract.
+    pub struct VestingWallet {
+        /// Ownable contract
```
bidzyyys, line 109:
```suggestion
        /// [`Ownable`] contract.
```

contracts/src/finance/vesting_wallet.rs
```diff
@@ -0,0 +1,617 @@
...
+pub enum Error {
+    /// Error type from [`Ownable`] contract [`ownable::Error`].
+    Ownable(ownable::Error),
+    /// Error type from [`call::Call`] contract [`call::Error`].
+    StylusError(call::Error),
+    /// Indicates an error related to the underlying Ether transfer.
+    ReleaseEtherFailed(ReleaseEtherFailed),
+    /// Indicates an error related to the underlying [`IErc20`] transfer.
+    ReleaseTokenFailed(ReleaseTokenFailed),
+    /// The token address is not valid. (eg. `Address::ZERO`)
```
bidzyyys, line 102:
```suggestion
    /// The token address is not valid. (eg. `Address::ZERO`).
```

contracts/src/finance/vesting_wallet.rs
```diff
@@ -0,0 +1,617 @@
...
+    error ReleaseEtherFailed();
+
+    /// Indicates an error related to the underlying ERC20 transfer.
+    ///
+    /// * `token` - Address of the token being released.
+    #[derive(Debug)]
+    #[allow(missing_docs)]
+    error ReleaseTokenFailed(address token);
+
+    /// The token address is not valid. (eg. `Address::ZERO`)
```
bidzyyys, line 70:
```suggestion
    /// The token address is not valid. (eg. `Address::ZERO`).
```

contracts/src/finance/vesting_wallet.rs
```diff
@@ -0,0 +1,617 @@
...
+/// BorrowMut<Self>)`. Should be fixed in the future by the Stylus team.
+unsafe impl TopLevelStorage for VestingWallet {}
+
+/// Required interface of a [`VestingWallet`] compliant contract.
+#[interface_id]
+pub trait IVestingWallet {
+    /// The error type associated to this trait implementation.
+    type Error: Into<alloc::vec::Vec<u8>>;
+
+    /// The contract should be able to receive Eth.
```
bidzyyys, line 135:
```suggestion
    /// The contract should be able to receive Eth.
    ///
    /// # Arguments
    ///
    /// * `&self` - Read access to the contract's state.
```

contracts/src/finance/vesting_wallet.rs
```diff
@@ -0,0 +1,617 @@
...
+    ///
+    /// # Errors
+    ///
+    /// If the `token` address is not a contract, then the error
+    /// [`Error::InvalidToken`] is returned.
+    ///
+    /// # Panics
+    ///
+    /// If `total_allocation * (timestamp - self.start())` exceeds `U256::MAX`.
+    #[selector(name = "vestedAmount")]
```
bidzyyys, line 343: I believe `#[selector(name = "...")]` is not needed in trait definition.

contracts/src/finance/vesting_wallet.rs
```diff
@@ -0,0 +1,617 @@
...
+
+// TODO: use existing IErc20
+pub use token::IErc20;
+#[allow(missing_docs)]
+mod token {
+    alloy_sol_types::sol! {
+        /// Interface of the ERC-20 token.
+        interface IErc20 {
+            function balanceOf(address account) external view returns (uint256);
+            function transfer(address to, uint256 value) external returns (bool);
```
bidzyyys, line 86: I believe `transfer` is not used in the contract anymore.

contracts/src/finance/vesting_wallet.rs
```diff
@@ -0,0 +1,617 @@
...
+        &self,
+        token: Address,
+        timestamp: u64,
+    ) -> Result<U256, Self::Error> {
+        if !Address::has_code(&token) {
+            return Err(InvalidToken { token }.into());
+        }
+
+        let call = IErc20::balanceOfCall { account: contract::address() };
+        let balance = RawCall::new()
```
bidzyyys, line 465: I believe [`Call`](https://docs.rs/stylus-sdk/latest/stylus_sdk/call/struct.Call.html) would be more convenient to use here. But you would need to `sol_interface!` macro instead of `sol!`.

bidzyyys (commented):
Left some comments.

contracts/src/finance/vesting_wallet.rs
0xNeshi: oops, leftover from some temp. attempt

contracts/src/finance/vesting_wallet.rs
0xNeshi: Can it be removed from [IErc721::safe_transfer_from_with_data](https://github.com/OpenZeppelin/rust-contracts-stylus/blob/main/contracts/src/token/erc721/mod.rs#L329) too?

contracts/src/finance/vesting_wallet.rs
bidzyyys: That's a great question!
I used `RawCall` in `SafeErc20` because most of the functions that need to be called may return bool value or nothing -- `RawCall` allows more advanced configurations for other contracts calls (like [setting return data length](https://github.com/OpenZeppelin/rust-contracts-stylus/blob/main/contracts/src/token/erc20/utils/safe_erc20.rs#L369)). For `RawCall` contract's interface needs to be specified in `alloy_sol_types::sol!` macro.
On the other hand `Call` is more convenient to be used by a specific interface (the interface needs to be specified in `stylus_sdk::prelude::solidity_interface!` macro) - it calls a contract and casts the output to the specified one as you can see in this example:
```
use stylus_sdk::call::{Call, Error};
use stylus_sdk::{prelude::*, evm, msg, alloy_primitives::Address};
extern crate alloc;

sol_interface! {
    interface IService {
        function makePayment(address user) external payable returns (string);
    }
}

pub fn do_call(account: IService, user: Address) -> Result<String, Error> {
    let config = Call::new()
        .gas(evm::gas_left() / 2)       // limit to half the gas left
        .value(msg::value());           // set the callvalue

    account.make_payment(config, user)  // note the snake case
}
```

To summarise, I wanted to have a consistent calling within a smart contract implementation.

contracts/src/finance/vesting_wallet.rs
```diff
@@ -0,0 +1,621 @@
...
+        let amount = self.releasable_erc20(token)?;
+        let owner = self.ownable.owner();
+
+        self.safe_erc20.safe_transfer(token, owner, amount).map_err(|_| {
+            Error::ReleaseTokenFailed(ReleaseTokenFailed { token })
+        })?;
+
+        // SAFETY: total supply of a [`crate::token::erc20::Erc20`] cannot
+        // exceed `U256::MAX`.
+        self._erc20_released.setter(token).add_assign_unchecked(amount);
```
bidzyyys, line 444: I believe someone may use a token that is not built on top of our contracts (overflow possibly may occur). I would still use safe math and panic.

contracts/src/finance/vesting_wallet.rs
```diff
@@ -0,0 +1,621 @@
...
+    ) -> Result<U256, Self::Error> {
+        if !Address::has_code(&token) {
+            return Err(InvalidToken { token }.into());
+        }
+
+        let erc20 = IErc20::new(token);
+        let call = Call::new_in(self);
+        let balance = erc20
+            .balance_of(call, contract::address())
+            .expect("should return the balance");
```
bidzyyys, line 474: I believe you do not need to check `Address::has_code(&token)` but just map an error into `InvalidToken` here. `Erc20::balance_of` should always return `U256`. Additionally, any other contract may exist on this address but it may not be `Erc20` - it will panic and this path is not covered in rust docs.

contracts/src/finance/vesting_wallet.rs
```diff
@@ -0,0 +1,621 @@
...
+
+        // SAFETY: total supply of a [`crate::token::erc20::Erc20`] cannot
+        // exceed `U256::MAX`.
+        let total_allocation = balance + self.released_erc20(token);
```
bidzyyys, line 478: I believe someone may use a token that is not built on top of our contracts (overflow possibly may occur). I would still use safe math and panic.

contracts/src/finance/vesting_wallet.rs
```diff
@@ -0,0 +1,621 @@
...
+    fn vesting_schedule(&self, total_allocation: U256, timestamp: U64) -> U256 {
+        let timestamp = U256::from(timestamp);
+
+        if timestamp < self.start() {
+            U256::ZERO
+        } else if timestamp >= self.end() {
+            total_allocation
+        } else {
+            // SAFETY: `timestamp` is guaranteed to be greater than
+            // `self.start()`, as checked by earlier bounds.
```
bidzyyys, line 507:
```suggestion
            // `self.start()` as checked by earlier bounds.
```

examples/vesting-wallet/tests/mock/erc20_return_false.rs
bidzyyys, line 1: I would say assume that `SafeErc20` works properly in tests for `VestingWallet`.

examples/vesting-wallet/src/constructor.sol
```diff
@@ -0,0 +1,33 @@
...
+    event OwnershipTransferred(
+        address indexed previousOwner,
+        address indexed newOwner
+    );
+
+    constructor(address beneficiary, uint64 startTimestamp, uint64 durationSeconds) payable {
+         if (beneficiary == address(0)) {
+            revert OwnableInvalidOwner(address(0));
+        }
+        _transferOwnership(beneficiary);
```
bidzyyys, line 22: Nit: it would be fine to just set `beneficiary` like `_owner = beneficiary`, even without emitting an event. The reason is that we do not care here about `Ownable` but about `VestingWallet`.

examples/vesting-wallet/tests/vesting-wallet.rs
```diff
@@ -0,0 +1,438 @@
...
+mod erc20_vesting {
+    use super::*;
+
+    async fn deploy(
+        account: &Account,
+        start: u64,
+        duration: u64,
+        allocation: u64,
+    ) -> eyre::Result<(Address, Address)> {
+        let contract_addr = account
```
bidzyyys, line 230: Nit: I would suggest having a dedicated fn for both eth & erc20 for deploying vesting wallet and additionally `deploy_erc20` for deploying just `erc20` token. For me it is not clear in this fn name.

examples/vesting-wallet/tests/vesting-wallet.rs
bidzyyys, line 1: Are you able to cause any panic in E2E tests?

bidzyyys (commented):
Great job @0xNeshi, left few comments.

contracts/src/finance/vesting_wallet.rs
```diff
@@ -0,0 +1,621 @@
...
+    ///
+    /// If the `token` address is not a contract, then the error
+    /// [`Error::InvalidToken`] is returned.
+    ///
+    /// # Panics
+    ///
+    /// If `total_allocation * (timestamp - self.start())` exceeds `U256::MAX`.
+    #[selector(name = "vestedAmount")]
+    fn vested_amount_erc20(
+        &mut self,
```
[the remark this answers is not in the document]
0xNeshi: This is required to be able to pass it to [Call::new_in](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/402/files/3339d41ac9e4eefefccd4d2d461f370831bd756b#diff-4083d07e30a6c3e16c834ce38c8c2236c57a943e7a89f8e4212209eb4a1bcd96R471). 
Now that I think about it, is reentrancy a problem in this function? :thinking: Maybe using `Call::new()` would be enough here, so that `mut` can be removed.

contracts/src/finance/vesting_wallet.rs
```diff
@@ -0,0 +1,621 @@
...
+    #[selector(name = "release")]
+    fn release_eth(&mut self) -> Result<(), Self::Error> {
+        let amount = self.releasable_eth();
+        let released = self
+            .released_eth()
+            .checked_add(amount)
+            .expect("should not exceed `U256::MAX` for `_released`");
+        self._released.set(released);
+
+        evm::log(EtherReleased { amount });
```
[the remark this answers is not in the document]
0xNeshi: Originally that's how I wrote this, but opted for emitting before the transfer to avoid having to manually return `Ok(())`.

Current (the result is whatever `transfer_eth` returns):

```rust
        evm::log(EtherReleased { amount });

        transfer_eth(self.ownable.owner(), amount)
            .map_err(|_| Error::ReleaseEtherFailed(ReleaseEtherFailed {}))
```

After reorder:

```rust
        transfer_eth(self.ownable.owner(), amount)
            .map_err(|_| Error::ReleaseEtherFailed(ReleaseEtherFailed {}))?;

        evm::log(EtherReleased { amount });

        Ok(())
```

I don't mind having the latter, will change

contracts/src/finance/vesting_wallet.rs
0xNeshi: Nice catch

contracts/src/finance/vesting_wallet.rs
0xNeshi: Nice catch

examples/vesting-wallet/tests/mock/erc20_return_false.rs
0xNeshi: This mock is here to force `SafeErc20` transfer to fail, so that we have a test case verifying that in [this case `VestingWallet` reverts transactions](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/402/files/3339d41ac9e4eefefccd4d2d461f370831bd756b#diff-f5693d7266bf932d5637b73b39158615807090da40f9254f5bdae49c5b54fe5dR389).

examples/vesting-wallet/tests/vesting-wallet.rs
0xNeshi: Will add additional test cases for various overflow checks

0xNeshi: Would appreciate your help here.

I added a test for [erc20 to force a panic in `vesting_schedule`](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/402/commits/a7abc28b764e859bd8680b470e1c050f8c92f822), but stumbled upon a problem when trying to do the same when adding one for Ether - I needed to be able to fund the contract with `U256::MAX` Ether, but was unable to do so with `e2e:fund_account` (not enough funds to cover tx costs).

Also, I couldn't find an example of how to create a mock Stylus ERC-20 contract for tests that uses `U320` or larger for token supply? This was I could force a panic [in these cases](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/402#discussion_r1846958972).

examples/vesting-wallet/tests/mock/erc20_return_false.rs
bidzyyys: Ok.

examples/vesting-wallet/tests/vesting-wallet.rs
bidzyyys: If you are unable to write this test, it is fine to skip it.

0xNeshi:
@bidzyyys @qalisander would appreciate your help to understand why [gas-bench workflow fails](https://github.com/OpenZeppelin/rust-contracts-stylus/actions/runs/11930050607/job/33250077909) when vesting-wallet benches are added to _benches/src/main.rs_ :pray:
I remove it until I figure this out

contracts/src/finance/vesting_wallet.rs
```diff
@@ -0,0 +1,621 @@
...
+    }
+
+    fn duration(&self) -> U256 {
+        U256::from(self._duration.get())
+    }
+
+    fn end(&self) -> U256 {
+        // SAFETY: both `start` and `duration` are stored as u64,
+        // so they cannot exceed `U256::MAX`
+        self.start() + self.duration()
```
bidzyyys, line 389:
```suggestion
        self.start().checked_add(self.duration()).expect("should not panic")
```

contracts/src/finance/vesting_wallet.rs
```diff
@@ -0,0 +1,621 @@
...
+        Ok(vested - self.released_erc20(token))
+    }
+
+    #[selector(name = "release")]
+    fn release_eth(&mut self) -> Result<(), Self::Error> {
+        let amount = self.releasable_eth();
+        // SAFETY: total vested amount is by definition smaller than or equal to
+        // the total allocation, which is already verified to be smaller than
+        // `U256::MAX` in [`Self::releasable_eth`].
+        self._released.set(self._released.get() + amount);
```
bidzyyys, line 426:
```suggestion
        let released = self._released.get().checked_add(amount).expect("should not fail");
        self._released.set(released);
```

contracts/src/finance/vesting_wallet.rs
```diff
@@ -0,0 +1,621 @@
...
+        Ok(())
+    }
+
+    #[selector(name = "release")]
+    fn release_erc20(&mut self, token: Address) -> Result<(), Self::Error> {
+        let amount = self.releasable_erc20(token)?;
+        let owner = self.ownable.owner();
+
+        self.safe_erc20.safe_transfer(token, owner, amount).map_err(|_| {
+            Error::ReleaseTokenFailed(ReleaseTokenFailed { token })
```
bidzyyys, line 442: I believe it should return wrapped error from `SafeErc20`.

contracts/src/finance/vesting_wallet.rs
bidzyyys: Ok. I believe we should have panicking implementation similar to `add_assign_unchecked`.

contracts/src/finance/vesting_wallet.rs
```diff
@@ -0,0 +1,621 @@
...
+    #[selector(name = "release")]
+    fn release_eth(&mut self) -> Result<(), Self::Error> {
+        let amount = self.releasable_eth();
+        // SAFETY: total vested amount is by definition smaller than or equal to
+        // the total allocation, which is already verified to be smaller than
+        // `U256::MAX` in [`Self::releasable_eth`].
+        self._released.set(self._released.get() + amount);
+
+        transfer_eth(self.ownable.owner(), amount)
+            .map_err(|_| Error::ReleaseEtherFailed(ReleaseEtherFailed {}))?;
```
bidzyyys, line 429: I believe it should returned wrapped error from `tranfer_eth`.

examples/vesting-wallet/tests/vesting-wallet.rs
```diff
@@ -0,0 +1,477 @@
...
+/// (to calculate the `start` timestamp) and the final release of vested funds
+/// in the test, it is best we assert that the released amount is within
+/// some predefined range.
+/// The reason why the timestamp can change is that we perform many mutations
+/// on-chain, from deploying and activating contracts, sending initial ETH/ERC20
+/// to the contract and then finally releasing the funds.
+fn assert_in_delta(expected: U256, actual: U256) {
+    let diff = expected.abs_diff(actual);
+    let delta = U256::from(1);
+    assert!(diff <= delta,"Your result of {actual} should be within {delta} of the expected result {expected}");
```
bidzyyys, line 63:
```suggestion
    assert!(diff <= delta, "Your result of {actual} should be within {delta} of the expected result {expected}");
```

examples/vesting-wallet/tests/vesting-wallet.rs
```diff
@@ -0,0 +1,477 @@
...
+        durationSeconds: duration_seconds,
+    }
+}
+
+async fn block_timestamp(account: &Account) -> eyre::Result<u64> {
+    let timestamp = account
+        .wallet
+        .get_block(
+            BlockId::latest(),
+            alloy::rpc::types::BlockTransactionsKind::Full,
```
bidzyyys, line 43: Why not import `use alloy::rpc::types::BlockTransactionKind;`?

bidzyyys (commented):
Almost ready to merge!

contracts/src/finance/vesting_wallet.rs
qalisander: But seems panic shouldn't be possible. Make sense to omit the check, isn't it?

bidzyyys: Right now it is not possible, but what if the contract would use `U256` for `start` and `duration` in the future?

qalisander: It will look like a mistake then. U64 timestamp should be more then enough for the current century :)

qalisander: I see that at solidity it was done, to remove unnecessary checks. And improve gas usage. Idk really how is it related to our case. But yeah, I don't mind to copy their approach for now.
Besides @0xNeshi added edge case to the `reads_max_end` test case

contracts/src/finance/vesting_wallet.rs
```diff
@@ -0,0 +1,621 @@
...
+            two,
+            contract.vesting_schedule(two, start + duration + U64::from(1))
+        );
+    }
+
+    #[motsu::test]
+    fn gets_vesting_schedule_zero_duration(contract: VestingWallet) {
+        let start = start();
+        contract._start.set(start);
+        contract._duration.set(U64::ZERO);
```
qalisander, line 610: Nit: If you're using this initialization frequently, make sense to introduce some kind of `fn init(start: U64, duration: u64)` function just for unit tests

qalisander (approved):
LGTM!
Awesome job @0xNeshi
I think we're ready to ship, may be some very minor things left..

contracts/src/finance/vesting_wallet.rs
0xNeshi: This should never panic, as `amount` and `_released` are equal to vested amount at this point in time, so if this were to ever panic, it would have happened in the [`releasable_eth` call](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/402/files/ba278154b86aba20ae81f076f0741a7514f6de3e#diff-4083d07e30a6c3e16c834ce38c8c2236c57a943e7a89f8e4212209eb4a1bcd96R406).

In general, I'd be more in favor of a tiny perf. impact of using `checked_X` instead of relying on `SAFETY` comments (the code can change, and the comment becomes wrong, even harmful), but interested in your thoughts?

examples/vesting-wallet/tests/vesting-wallet.rs
0xNeshi: It is only used here and nowhere else, but will import it as suggested

contracts/src/finance/vesting_wallet.rs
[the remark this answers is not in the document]
bidzyyys: I believe [`Call::new`](https://docs.rs/stylus-sdk/latest/stylus_sdk/call/struct.Call.html#method.new) wdyt @qalisander? Should we also use just `Call::new` in our other contracts?

contracts/src/finance/vesting_wallet.rs
bidzyyys: I am more into `checked_X` on situations like here. Code may change in the future and someone may forget to apply checks here.

bidzyyys (commented):
Almost ready to merge 🚀

contracts/src/finance/vesting_wallet.rs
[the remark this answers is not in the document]
qalisander: I suspect so. We should try to change

bidzyyys (approved):

PR #352, opened 2024-10-16
feat: Ownable2Step

ggonzalez94:
Resolves #290 

- Introduce a new Ownable2Step contract
- Make `onlyOwner()` internal in Ownable and implement `MethodError` to be able to use the error in our new contract(we've already seen we will need this for others)
- Propose we use nextest instead of vainilla `cargo test` in the CI. It is faster(not as relevant right now) and the output is so much nicer

#### PR Checklist
- [x] Unit Tests
- [x] e2e Tests
- [x] Documentation

PR #404, opened 2024-11-12
ref: improvements to Ownable2Step

bidzyyys:


contracts/src/access/ownable_two_step.rs
```diff
@@ -196,6 +188,22 @@ impl IOwnable2Step for Ownable2Step {
...
+    ///
+    /// # Arguments
+    ///
+    /// * `&self` - Read access to the contract's state.
+    ///
+    /// # Errors
+    ///
+    /// If called by any account other than the owner, then the error
+    /// [`OwnableError::UnauthorizedAccount`] is returned.
+    pub fn only_owner(&self) -> Result<(), OwnableError> {
```
ggonzalez94, line 203: Why are you re-exporting only owner back? It was an explicit [design decision](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/352#discussion_r1832280260). I think we should at least discuss it.
And if we do decide to add it, we should change all usages of `only_owner` to use this?

ggonzalez94 (commented):
Left a question about `only_owner`

contracts/src/access/ownable_two_step.rs
bidzyyys: After a second thought, I strongly believe that the interface of `Ownable2Step` should be same as in Solidity, so `only_owner()` should be possible to use on `Ownable2Step` contract without the need of accessing it via contract's attributes (better DevEx). But I am open to discussion on this.

contracts/src/access/ownable_two_step.rs
```diff
@@ -38,14 +38,6 @@ sol! {
         address indexed new_owner
     );
 
-    /// Emitted when ownership gets transferred between accounts.
-    ///
-    /// * `previous_owner` - Address of the previous owner.
-    /// * `new_owner` - Address of the new owner.
-    event OwnershipTransferred(
```
0xNeshi, line 45: Don't forget to update the Events section for [_transfer_ownership](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/404/files#diff-a792a5d83a14094d0fbcbde7b4e41620f927a81aab7154c2e30b2f3e71f4c28eR221) to state that it emits `crate::access::ownable::OwnershipTransferred`

0xNeshi (approved):

contracts/src/access/ownable_two_step.rs
ggonzalez94: Wel.... this isn't part of the public interface, it is just internal. I do agree that it is better DX, but my concern is that this will result in a bunch of code repetition if we follow this pattern(and we should be consistent)

0xNeshi: I generally lean towards better DevEx (in this case reexporting `only_owner`), but not at the price of having to update (almost) all existing contracts that do not follow the proposed convention.

Maybe if we start getting complaints from devs we can rethink?

contracts/src/access/ownable_two_step.rs
bidzyyys: Solved.

contracts/src/access/ownable_two_step.rs
bidzyyys: Reverted

PR #414, opened 2024-11-20
docs: Refactor pausable for erc721 example

### LINKED ISSUE #212: Refactor pausable for erc721

Try to apply an approach similar to consecutive implementation for erc721. But for pausable extension.
Implement PausableErc721 contract that duplicates functions from erc721 that are necessary to be overridden. (just `_update` in the context of pausable extension)
And duplicate functions that depend on `_update` also.

p.s. potentially can help if stylus sdk team will force to have unique selectors.

### Contribution Guidelines

0xNeshi:
Resolves #212


bidzyyys (changes requested):
I am against those changes as they are NOT consistent with our examples.
If we want to merge this PR, also all the rust docs should be updated where needed.

@qalisander @ggonzalez94 wdyt?

qalisander:
I see what I missed. Technically you're calling `when_not_paused` every time we need it. I think we need to refactor our pausable implementation before updating examples

0xNeshi:
@qalisander @bidzyyys ahh, life without multi-inheritance is such a pain :pensive: 

The more I look at this code, the more I like the original design. If a dev wants to make a pausable ERC721, they should inherit both and reimplement all the functions they want to make pausable to just call `when_not_paused` before calling the relevant erc721 function after it. Although it's a bit boring to have to do that, I think it's much worse to have to actually reimplement all the underlying erc721 function because it all depends on `fn _update`.
(Technically speaking, it would have been more correct if I had reimplemented all the erc721 functions I wanted to make pausable and public instead of going around them and calling `_update` directly. It looks a bit hacky imho)

Maybe we just close this PR and leave things as-is until multi-inheritance is properly supported, what do you guys think?

0xNeshi:
We [agreed](https://openzeppelin.slack.com/archives/C06R5AGFE9G/p1733469339343659) to close this.

PR #423, opened 2024-11-28
refactor(test): migrate unit tests to the new motsu layout

### LINKED ISSUE #326: [Feature]: Mock another contract in `#[motsu::test]`

Some contracts need to call another contracts, e.g. `Erc721::_check_on_erc721_received`.
We want to be able to mock some behaviour, or at least mock return data.

```rust
pub fn _check_on_erc721_received(
        &mut self,
        operator: Address,
        from: Address,
        to: Address,
        token_id: U256,
        data: Bytes,
    ) -> Result<(), Error> {
        const RECEIVER_FN_SELECTOR: FixedBytes<4> = fixed_bytes!("150b7a02");

        if !to.has_code() {
            return Ok(());
        }

        let receiver = IERC721Receiver::new(to);
        let call = Call::new_in(self);
        let result = receiver.on_erc_721_received(
            call,
            operator,
            from,
            token_id,
            data.to_vec().into(),
        );

        let id = match result {
            Ok(id) => id,
            Err(e) => {
                if let call::Error::Revert(ref reason) = e {
                    if reason.len() > 0 {
                        // Non-IERC721Receiver implementer.
                        return Err(Error::InvalidReceiverWithReason(e));
                    }
                }

                return Err(ERC721InvalidReceiver { receiver: to }.into());
            }
        };

        // Token rejected.
        if id != RECEIVER_FN_SELECTOR {
            return Err(ERC721InvalidReceiver { receiver: to }.into());
        }

        Ok(())
    }
}
```

### Contribution Guidelines

### LINKED ISSUE #470: [Feature]: Un-ignore and Fix Erc20FlashMint Unit Tests

See https://github.com/OpenZeppelin/rust-contracts-stylus/pull/407/files/987d3a3496bcdce46fa11a27c2ec54dbec113a0c#r1886275933

Blocked until https://github.com/OpenZeppelin/stylus-test-helpers/issues/5 is implemented in `motsu`

### Contribution Guidelines

qalisander:
This pr contains migration of unit test to the new layout of motsu unit tests, that will be introduced in OpenZeppelin/stylus-test-helpers#8 and OpenZeppelin/stylus-test-helpers#12

Resolves #470 

#### PR Checklist

- [x] Tests
- [ ] Documentation
- [ ] Changelog


contracts/src/utils/context.rs
```diff
@@ -0,0 +1,17 @@
...
+    motsu::prelude::Context::current()
+        .get_msg_sender()
+        .expect("msg_sender should be set")
+}
+
+/// Returns the address of the message sender.
+#[cfg(not(test))]
+pub fn msg_sender() -> Address {
+    msg::sender()
+}
```
qalisander, line 17: Here is an example how `msg_sender` can be called just for test cases. This design is not perfect since motsu users should implement this mock for their own library manually or through macro.
Calling stylus sdk api `msg::sender()` won't work due to [this limitation](https://github.com/OpenZeppelin/rust-contracts-stylus/issues/30#issuecomment-2037231720).

contracts/src/token/erc721/mod.rs
```diff
@@ -2505,4 +2515,49 @@ mod tests {
...
+        let token_id = random_token_id();
+        erc721
+            .sender(alice)
+            ._safe_mint(receiver.address(), token_id, vec![0, 1, 2, 3].into())
+            .unwrap();
+
+        let received_token_id = receiver.sender(alice).received_token_id();
+
+        assert_eq!(received_token_id, token_id);
+    }
```
qalisander, line 2562: Overall the test case for Erc721Receiver can look like this

lib/motsu/src/lib.rs
```diff
@@ -48,3 +48,118 @@ pub mod prelude;
...
+        alloy_primitives::{Address, U256},
+        call::Call,
+        prelude::{public, sol_storage, StorageType, TopLevelStorage},
+    };
+
+    use crate::context::{Account, Contract, TestRouter};
+
+    /// Message sender that mocks `msg::sender()` in tests.
+    #[must_use]
+    pub fn msg_sender() -> Address {
```
0xNeshi, line 68: Why is the helper redefined here when it's already defined with the same code in [context.rs](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/423/files#diff-c9edad7c85ffd5b6ce47be67e85ec0460548f9c7b988b60a6ae7874c93f87567R7)? Can't we just import it?

lib/motsu/src/context.rs
```diff
@@ -46,11 +54,196 @@ impl Context {
...
+        self.get_storage().msg_sender
+    }
+
+    /// Set the address of the contract, that should be called.
+    fn set_msg_receiver(&self, msg_receiver: Address) -> Option<Address> {
+        self.get_storage().msg_receiver.replace(msg_receiver)
+    }
+
+    /// Get the address of the contract, that should be called.
+    fn get_msg_receiver(&self) -> Option<Address> {
```
0xNeshi, line 91: Warning: ```method `get_msg_receiver` is never used```

contracts/src/utils/context.rs
```diff
@@ -0,0 +1,17 @@
+//! Context functions used for mocking unit tests.
+
+use stylus_sdk::{alloy_primitives::Address, msg};
+
+/// Returns the address of the message sender.
+#[cfg(test)]
+pub fn msg_sender() -> Address {
+    motsu::prelude::Context::current()
+        .get_msg_sender()
+        .expect("msg_sender should be set")
```
0xNeshi, line 10: Devs would then need to copy/paste these `msg_sender` definitions in order to take advantage of motsu's mocking. 
Since these are so tightly related to motsu, would it make sense to make these helpers _a part_ of motsu?

motsu would then be a testing library that provides wrappers for `msg::sender`, `msg::value`, `block::timestamp` etc. that devs can use to be able to then mock these in their motsu (unit) tests.

Pros: devs can just use these helpers; easier to setup and use motsu; clearer documentation.
Cons: once the Stylus team adds ability to mock these from the SDK, it will make our wrappers redundant - devs using motsu will have a breaking change at some point.

 @bidzyyys @qalisander wdyt?

NOTE: given that the Stylus team will enable mocking these values, thus making our wrappers redundant, it makes our efforts to mock these temporary, whether we make these helpers part of motsu or not.

qalisander: >would it make sense to make these helpers a part of motsu?
Motsu is a part of `[dev-dependencies]`. In this case we need make motsu available for non-dev `[depencencies]`, or introduce the other helper module.
p.s. `msg_sender()` abstraction should be available as a non development dependency. It is not just for tests.

> NOTE: given that the Stylus team will enable mocking these values, thus making our wrappers redundant, it makes our efforts to mock these temporary, whether we make these helpers part of motsu or not.
Correct @0xNeshi, this was a temporary solution, since I didn't know how happy OL team about stylus sdk cache reset. We will remove it once stylus sdk be ready.

lib/motsu/src/lib.rs
qalisander: This one for test cases. It shows the main disadvantage of an approach, that user should declare that mock by himself. 
Will be removed once stylus sdk will be ready not to cache the `msg::sender()` value

contracts/src/utils/context.rs
0xNeshi: Yeah, it would make sense to make these helpers a non-dev dependency then. 

We can keep things as-is for now, as it's possible OL will update their SDK to allow mocking these values soon.
We shouldn't invest too much time into this imo.

bidzyyys:
This change should be common with #441

0xNeshi (commented):
Imo the API is very good, especially as it conforms to existing conventions for writing code that mocks senders, values etc. 

I think the builder pattern is the right choice here, as we might need to extend the functionality at a later point (e.g. [to add `.with_block_timestamp`](https://github.com/OpenZeppelin/rust-contracts-stylus/issues/337), but there may be others, who knows).

Good job! :fire:

Once clippy warnings are addressed I'll give this a final review, as the PR is too cluttered with warnings atm

contracts/Cargo.toml
```diff
@@ -14,14 +14,15 @@ alloy-sol-types.workspace = true
...
-stylus-sdk.workspace = true
+stylus-sdk = { workspace = true, default-features = true }
 keccak-const.workspace = true
 openzeppelin-stylus-proc.workspace = true
 
 [dev-dependencies]
 alloy-primitives = { workspace = true, features = ["arbitrary"] }
 motsu.workspace = true
 rand.workspace = true
+stylus-sdk.workspace = true
```
0xNeshi, line 25: stylus-sdk is listed both in `dependencies` and in `dev-dependencies`

contracts/src/access/control.rs
```diff
@@ -41,6 +41,8 @@
 //! accounts that have been granted it. We recommend using
 //! `AccessControlDefaultAdminRules` to enforce additional security measures for
 //! this role.
+use alloc::vec::Vec;
```
0xNeshi, line 44: Why are we adding this import?

contracts/src/finance/vesting_wallet.rs
```diff
@@ -89,15 +91,11 @@ pub enum Error {
...
-    stylus_sdk::stylus_proc::sol_interface! {
-        /// Interface of the ERC-20 token.
-        interface IErc20 {
-            function balanceOf(address account) external view returns (uint256);
-        }
+sol_interface! {
+    /// Interface of the ERC-20 token.
+    #[allow(missing_docs)]
+    interface IErc20 {
+        function balanceOf(address account) external view returns (uint256);
```
0xNeshi, line 98: Won't coverage be reduced by including this?

qalisander: Yeap. It will be reduced, currently the simplest way to compile it.

0xNeshi: It doesn't compile if you do:
```rust
pub use token::IErc20;
mod token {
    #![allow(missing_docs)]
    #![cfg_attr(coverage_nightly, coverage(off))]
    stylus_sdk::prelude::sol_interface! {
        /// Interface of the ERC-20 token.
        interface IErc20 {
            function balanceOf(address account) external view returns (uint256);
        }
    }
}
```
? :open_mouth:

qalisander: `IErc20` now made private. So it will de-sugar to `struct IErc20`, instead of `pub struct IErc20`. And visibility level will just mod wise

contracts/Cargo.toml
qalisander: Since OL agreed to have a default feature for ENABLING hostio caching, now we should write `stylus-sdk = { version = "0.7.0", default-features = false }` on the workspace level, and `stylus-sdk = { workspace = true, default-features = true }` for `[dependencies]`.
And use without default features for unit tests `[dev-dependencies]`

qalisander: I would better add a feature for disabling caching (not enabling). It would more simple

contracts/src/finance/vesting_wallet.rs
qalisander: May be try to change it again in stylus sdk..

qalisander:
@0xNeshi I'll keep migration to the new unit-test layout within this pr

examples/access-control/src/lib.rs
```diff
@@ -48,3 +48,7 @@ impl AccessControlExample {
         self.access._set_role_admin(role, new_admin_role)
     }
 }
+
+impl AccessControlExample {
```
bidzyyys, line 52: I am not sure if we need this impl block.

bidzyyys (commented):
This new motsu layout is great 🚀

contracts/src/token/erc20/extensions/flash_mint.rs
bidzyyys, line 1: @0xNeshi I believe #470 is solved with this PR, correct me if I am wrong please 🙏

benches/src/erc721.rs
```diff
@@ -57,9 +57,9 @@ pub async fn run(cache_opt: Opt) -> eyre::Result<Vec<FunctionReport>> {
     let token_3 = uint!(3_U256);
     let token_4 = uint!(4_U256);
 
-    let _ = receipt!(contract.mint(alice_addr, token_2))?;
-    let _ = receipt!(contract.mint(alice_addr, token_3))?;
-    let _ = receipt!(contract.mint(alice_addr, token_4))?;
+    _ = receipt!(contract.mint(alice_addr, token_2))?;
+    _ = receipt!(contract.mint(alice_addr, token_3))?;
+    _ = receipt!(contract.mint(alice_addr, token_4))?;
```
0xNeshi, line 62:
```suggestion
    receipt!(contract.mint(alice_addr, token_2))?;
    receipt!(contract.mint(alice_addr, token_3))?;
    receipt!(contract.mint(alice_addr, token_4))?;
```
nit: not sure why ever made these redundant assignments haha

benches/src/lib.rs
```diff
@@ -125,7 +125,7 @@ fn cache_contract(
 ) -> eyre::Result<()> {
     // We don't need a status code.
     // Since it is not zero when the contract is already cached.
-    let _ = Command::new("cargo")
+    _ = Command::new("cargo")
```
0xNeshi, line 128:
```suggestion
    Command::new("cargo")
```

contracts/src/token/erc1155/extensions/supply.rs
```diff
@@ -390,41 +391,61 @@ mod tests {
...
-        assert_eq!(values[0], contract.total_supply(token_ids[0]));
-        assert_eq!(values[0], contract.total_supply_all());
-        assert!(contract.exists(token_ids[0]));
+    fn after_mint_single(
+        contract: Contract<Erc1155Supply>,
+        alice: Address,
+        bob: Address,
+    ) {
+        let (token_ids, values) =
+            contract.init(alice, |contract| init(contract, bob, 1));
```
0xNeshi, line 408: Tbh this looks weird compared to doing the same manually:

```rust
        let mut callable = contract.sender(alice);
        let (token_ids, values) = init(&mut callable, bob, 1);
```

0xNeshi (approved):
This PR looks like it took quite a bit of time, kudos to you both!

contracts/src/token/erc20/extensions/erc4626.rs
```diff
@@ -1143,11 +1143,11 @@ impl Erc4626 {
     }
 }
 
-// TODO: Add missing tests once `motsu` supports calling external contracts.
 #[cfg(all(test, feature = "std"))]
```
0xNeshi, line 1146:
```suggestion
// TODO: Add missing tests once `motsu` supports calling external contracts.
#[cfg(all(test, feature = "std"))]
```
No new tests were added, so this comment should be kept

contracts/src/token/erc20/extensions/flash_mint.rs
0xNeshi: [Unfortunately no](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/423/files#r1944860604) :sad-pepe:

bidzyyys (approved):

0xNeshi:
I think it's safe to merge this PR, as it doesn't rely on [e2e tests](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/531), only on unit tests (which pass)

PR #441, opened 2024-12-10
feat: add event checks to motsu

bidzyyys:
Resolves OpenZeppelin/stylus-test-helpers#9 

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog


lib/motsu/src/context/environment.rs
```diff
@@ -0,0 +1,83 @@
...
+
+/// Dummy msg sender set for tests.
+const MSG_SENDER: &[u8; 42] = b"0xDeaDbeefdEAdbeefdEadbEEFdeadbeEFdEaDbeeF";
+
+pub(crate) struct Environment {
+    account_codehash: [u8; 66],
+    block_timestamp: u64,
+    chain_id: u64,
+    contract_address: [u8; 42],
+    events: Vec<Vec<u8>>,
```
qalisander, line 24: I would better create wrapper type for event, to store them like `Vec<Event>`

lib/motsu/src/context/environment.rs
```diff
@@ -0,0 +1,83 @@
...
+    b"0xdCE82b5f92C98F27F116F70491a487EFFDb6a2a9";
+
+/// Externally Owned Account (EOA) code hash.
+const EOA_CODEHASH: &[u8; 66] =
+    b"0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470";
+
+/// Dummy msg sender set for tests.
+const MSG_SENDER: &[u8; 42] = b"0xDeaDbeefdEAdbeefdEadbEEFdeadbeEFdEaDbeeF";
+
+pub(crate) struct Environment {
```
qalisander, line 19: I remember, that the whole reason of `MockStorage` was to store not just storage bytes (`contract_storage`), but also fields that you've put into `Environment` struct

qalisander: Because these data are also "mocked" values for our test cases

lib/motsu/src/context/environment.rs
```diff
@@ -0,0 +1,83 @@
+//! Module with unit test EVM environment for Stylus contracts.
+
+/// Block Timestamp - Epoch timestamp: 1st January 2025 `00::00::00`.
+const BLOCK_TIMESTAMP: u64 = 1_735_689_600;
```
qalisander, line 4: Good! Seems rational for me to move these constants to the other module

lib/motsu/src/context/mod.rs
```diff
@@ -0,0 +1,188 @@
...
+    pub(crate) unsafe fn emit_log(
+        self,
+        data: *const u8,
+        len: usize,
+        topics: usize,
+    ) {
+        // https://github.com/OffchainLabs/stylus-sdk-rs/blob/v0.6.0/stylus-sdk/src/evm.rs#L38-L52
+        let buffer = read_bytes(data, len);
+        let encoded_event: Vec<u8> =
+            buffer.clone().into_iter().skip(topics * WORD_BYTES).collect();
```
qalisander, line 98: If you would have `Event` struct (or `EncodedEvent`), here you could also add to it conversion from `buffer`

lib/motsu/src/context/mod.rs
```diff
@@ -0,0 +1,188 @@
...
+    storage: MockStorage,
+    environment: Environment,
+}
+
+/// A global mutable key-value store mockig EVM behaviour.
+/// Allows concurrent access.
+///
+/// The key is the name of the test thread,
+/// and the value is the context of the test case.
+static EVM: Lazy<DashMap<ThreadName, TestCase>> = Lazy::new(DashMap::new);
```
qalisander, line 134: We can call it `MOCK_EVM`. But yeah, mentioning `EVM` in the name of the test storage seems quite reasonable to me!

qalisander (commented):
Awesome @bidzyyys!
Few improvements I have:

lib/motsu/src/context/mod.rs
qalisander: Since we are trying to mimic its behavior for test cases

contracts/src/utils/pausable.rs
```diff
@@ -204,6 +206,10 @@ mod tests {
         let res = contract.pause();
         assert!(res.is_ok());
         assert_eq!(contract.paused(), true);
+        let expected_event = Paused { account: msg::sender() };
+        assert!(motsu::emits_event(expected_event.clone()));
```
qalisander, line 210: How about changing logic for `motsu::emits_event`, that it would take an event from collection in case of match. No need in calling `motsu::clear_events()` after.

bidzyyys:
Moved to https://github.com/OpenZeppelin/stylus-test-helpers/pull/13.

PR #446, opened 2024-12-10
ref: drop math for storage guard

qalisander:
Drop `add_assign_unchecked` and `sub_assign_unchecked` functions for `StorageGuardMut`, since guard already implements `Deref` and `DerefMut` to wrapped types.


0xNeshi (approved):
Nice catch!

bidzyyys (approved):
Great catch @qalisander but CHANGELOG first 🔔

qalisander:
That is internal refactoring. No change log update needed

Issue #264, opened 2024-08-26
[Feature]: Poseidon Hash


ggonzalez94:
### What is the feature you would like to see?

zk friendly hash functions tend to be computational expensive, which make them good candidates for Stylus.
Reference implemenations:
- Renegade: [here](https://github.com/renegade-fi/renegade/blob/main/renegade-crypto/src/hash/poseidon2.rs) 

### Contribution Guidelines

- [X] I agree to follow this project's Contribution Guidelines

PR #463, opened 2024-12-18
feat: stylus poseidon hash benchmark against solidity

qalisander:
- Optimize rise in `pow` for Poseidon hash.
- Add Poseidon hash comparison benchmarks between stylus and plain solidity implementations.
- Extend benchmark's report with additional column `WASM Opt & Cached`, that applies wasm-opt tool to the cached contract.
- Small refactoring of bench-marking logic.

Follows #264
Addresses #468

#### PR Checklist

- [x] Tests
- [x] Documentation
- [ ] Changelog


benches/src/PoseidonT3.sol
```diff
@@ -0,0 +1,53 @@
...
+    }
+    return swap0;
+  }
+
+  function pow5mod(uint i) public pure returns (uint) {
+    uint a = mulmod(i, i, F);
+    uint c = mulmod(a, a, F);
+    return mulmod(i, c, F);
+  }
+}
```
0xNeshi, line 53:
```suggestion
}

```

0xNeshi, line 53:
```suggestion
}

```
missing new line

benches/src/lib.rs
```diff
@@ -36,9 +38,10 @@ struct ArbOtherFields {
 /// Cache options for the contract.
 /// `Bid(0)` will likely cache the contract on the nitro test node.
 #[derive(Clone)]
-pub enum CacheOpt {
+pub enum Opt {
```
0xNeshi, line 41: Docs above need to be updated

benches/src/lib.rs
```diff
@@ -90,8 +102,11 @@ async fn deploy(
         .expect("should deploy contract")
         .address()?;
 
-    if let CacheOpt::Bid(bid) = cache_opt {
-        cache_contract(account, address, bid)?;
+    match opt {
+        Opt::Cache | Opt::WasmOpt => {
+            cache_contract(account, address, 0)?;
+        }
+        Opt::None => {}
```
0xNeshi, line 109: Would it make sense to have a benchmark for non-cached wasm-opted contracts?

benches/src/report.rs
```diff
@@ -24,14 +24,40 @@ pub struct ContractReport {
     contract: String,
     functions: Vec<FunctionReport>,
     functions_cached: Vec<FunctionReport>,
+    functions_wasm_opt_cached: Vec<FunctionReport>,
 }
 
 impl ContractReport {
+    pub async fn generate<
```
0xNeshi, line 31: Nice refactor :+1:

benches/src/poseidon_sol.rs
```diff
@@ -0,0 +1,112 @@
...
+}
+
+pub async fn deploy(account: &Account) -> eyre::Result<Address> {
+    let contract = PoseidonT3::deploy(&account.wallet).await?;
+    Ok(*contract.address())
+}
+
+sol! {
+    #[allow(missing_docs)]
+    // Built with Remix IDE; solc v0.8.21+commit.d9974bed
```
0xNeshi, line 60: nit: why not compile with 0.8.24 (our CI solc version)?

benches/src/poseidon_sol.rs
```diff
@@ -0,0 +1,112 @@
...
+    contract PoseidonT3 {
```
0xNeshi, line 62: Do we need the [_PoseidonT3.sol_ file](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/463/files#diff-23c85c11823d4385a7149be4c3e7ab6cc86ee95174c6d3ae8caf23c8ecfa87cd) as well?

benches/src/poseidon_sol.rs
```diff
@@ -0,0 +1,112 @@
...
+    }
+);
+
+pub async fn bench() -> eyre::Result<ContractReport> {
+    let reports = run().await?;
+    let report = reports
+        .into_iter()
+        .try_fold(ContractReport::new("Sol:Poseidon"), ContractReport::add)?;
+
+    Ok(report)
```
0xNeshi, line 27: Is there no way to cache Solidity contracts?

scripts/bench.sh
```diff
@@ -5,12 +5,37 @@ MYDIR=$(realpath "$(dirname "$0")")
...
 
+# Optimize contract's wasm binary by crate name.
+opt_wasm () {
+  local CONTRACT_CRATE_NAME=$1
+  local CONTRACT_BIN_NAME="${CONTRACT_CRATE_NAME//-/_}.wasm"
+  local CONTRACT_OPT_BIN_NAME="${CONTRACT_CRATE_NAME//-/_}_opt.wasm"
+
+  echo
+  echo "Optimising $CONTRACT_CRATE_NAME WASM binary"
+  # https://rustwasm.github.io/book/reference/code-size.html
```
0xNeshi, line 16:
```suggestion
  # https://rustwasm.github.io/book/reference/code-size.html#use-the-wasm-opt-tool
```
nit: link directly to wasm-opt section

scripts/bench.sh
```diff
@@ -5,12 +5,37 @@ MYDIR=$(realpath "$(dirname "$0")")
...
+  wasm-opt -O3 -o ./target/wasm32-unknown-unknown/release/"$CONTRACT_OPT_BIN_NAME" ./target/wasm32-unknown-unknown/release/"$CONTRACT_BIN_NAME"
```
0xNeshi, line 17: Did you test the gas costs with different optimization targets to see if any of them produced greater gas savings?

It's theoretically possible that reducing wasm size, or even opting for a less aggressive speed optimization (`-O`) would bring more gas savings, we can't know unless we test

Interesting that savings are only ~4%, but hey:
image

0xNeshi (commented):
Nice work!

benches/src/poseidon_sol.rs
qalisander: When I tried to cache sol contact last time, cache command failed. I suppose this is an exclusive feature of WASM compiled contracts

scripts/bench.sh
qalisander: I tested with `-O` flag (instead -`O3`). It was a bit worse. Also I've noticed, that wasm binary size got a bit smaller with `-O3` opt flag

benches/src/lib.rs
qalisander: Yeah, we could have `non-cached wasm-opt` option. But by the end, it will worse then only cached contract for most of our tokens/nfts. And for poseidon it will give the same result as only caching.
But ofcourse caching with wasm-opt is better

qalisander: Write now we can show the progression of gas savings with incremental optimizations

scripts/bench.sh
```diff
@@ -5,12 +5,37 @@ MYDIR=$(realpath "$(dirname "$0")")
...
+# Retrieve all alphanumeric contract's crate names in `./examples` directory.
+get_example_crate_names () {
+  # shellcheck disable=SC2038
+  # NOTE: optimistically relying on the 'name = ' string at Cargo.toml file
+  find ./examples -maxdepth 2 -type f -name "Cargo.toml" | xargs grep 'name = ' | grep -oE '".*"' | tr -d "'\""
+}
+
 cargo build --release --target wasm32-unknown-unknown -Z build-std=std,panic_abort -Z build-std-features=panic_immediate_abort
 
+# Optimize contract's wasm for gas usage.
```
bidzyyys, line 29: Previously you have `Optimising`, here `Optimize`. Use either AmE or BrE version please.

bidzyyys (approved):
Looks good to me.

.github/workflows/gas-bench.yml
```diff
@@ -28,6 +28,9 @@ jobs:
       - name: Install cargo-stylus
         run: cargo install cargo-stylus@0.5.3
 
+      - name: Install wasm-opt
+        run: cargo install wasm-opt@0.116.1
```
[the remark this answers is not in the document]
qalisander: Once we will start using in other GH actions, then yes.
But I see, that there is a point to have `wasm-opt` just for `gas-bench.yml` pipeline. Since gas optimizations are not crucial for tests.

benches/src/poseidon_sol.rs
0xNeshi: I understand. It would be ideal if we didn't have to duplicate the Solidity code there and here in `sol!` macro

benches/src/poseidon_sol.rs
0xNeshi: That seems peculiar, especially since Solidity _is_ deployable on Stylus. Ok then

scripts/bench.sh
0xNeshi: Did you test with `-Oz` (aggressively optimize for _size_)?

benches/src/PoseidonT3.sol
```diff
@@ -2,52 +2,52 @@
...
-    uint a = mulmod(i, i, F);
-    uint c = mulmod(a, a, F);
-    return mulmod(i, c, F);
-  }
+    function pow5mod(uint i) public pure returns (uint) {
+        uint a = mulmod(i, i, F);
+        uint c = mulmod(a, a, F);
+        return mulmod(i, c, F);
+    }
 }
```
0xNeshi, line 53:
```suggestion
}

```

0xNeshi (approved):
LGTM

scripts/bench.sh
0xNeshi: Tested it out, turned out worse for Poseidon, which makes sense

bidzyyys (approved):

Issue #386, opened 2024-10-30
Rethink current naming convention in contracts


bidzyyys:
### What is the feature you would like to see?

Right now we use in some cases Solidity like naming convention (`_do_sth()`). It is not recommended attitude in Rust and causes `cargo clippy` issues (right now disabled). We should rethink this topic and make sure our library is consistent in terms of naming.

Related https://github.com/OpenZeppelin/rust-contracts-stylus/issues/491
Related https://github.com/OpenZeppelin/rust-contracts-stylus/issues/499

TODO: Remove `_` prefix from state variables (for functions see https://github.com/OpenZeppelin/rust-contracts-stylus/issues/386#issuecomment-2595763777)

### Contribution Guidelines

- [X] I agree to follow this project's Contribution Guidelines

bidzyyys:
The same applies for struct attributes, like `_pub _owner: StorageAddress,`

0xNeshi:
Do we include [test name standardization](https://openzeppelin.slack.com/archives/C06R5AGFE9G/p1732709957377129?thread_ts=1732618832.030739&cid=C06R5AGFE9G) in this issue as well?

- `function_name_description` for standard tests.
- `function_name_revert_when_condition` for tests expecting a revert

bidzyyys:
> Do we include [test name standardization](https://openzeppelin.slack.com/archives/C06R5AGFE9G/p1732709957377129?thread_ts=1732618832.030739&cid=C06R5AGFE9G) in this issue as well?
> 
> * `function_name_description` for standard tests.
> * `function_name_revert_when_condition` for tests expecting a revert

@0xNeshi not in this issue, for test name standardisation I suggest having another one. Easier to solve & merge.

0xNeshi:
Makes sense, created a separate issue https://github.com/OpenZeppelin/rust-contracts-stylus/issues/491

0xNeshi:
@bidzyyys @qalisander 

In Solidity, internal functions use the `_fn_name` prefix pattern, making their visibility immediately clear. In our Rust contracts, we have a few scenarios to consider:

1. When a public function has an internal implementation (like `transfer_ownership`), we can use the `fn_name_internal` suffix pattern. This maintains clarity while following Rust conventions.

[...]

PR #467, opened 2024-12-26
ref: Add/Sub Assignment Operations + e2e-tests.sh

### LINKED ISSUE #443: [Feature]: Make E2E Target Names Easier to Use (i.e. Make _e2e-tests.sh_ Script Easier to Use)

E.g. currently there's no way to run only `safe-erc20-example` e2e tests without modifying the _e2e-tests.sh_ script. 
This is because we run our e2e tests from the workspace, which looks at _file names_ to filter out specific tests to run. 

Let's say you wanted to run them - you'd probably write something like:

```bash
./scripts/e2e-tests.sh safe-erc20*
```

This will error out with the following log:

```bash
error: no test target matches pattern `safe-erc20*`.
Available test targets:
    access_control
    ecdsa
    erc1155
    erc1155-metadata-uri
    erc20
    erc20permit
    erc721-consecutive
    erc721
    erc721
    ownable
    ownable_two_step
    address_with_no_code
    erc20
    erc20_that_always_returns_false
    erc20_that_does_not_return
    usdt_approval_behavior
    vesting-wallet
```

A couple of problems are immediately noticeable:
1. No target name indicates a relation to `SafeErc20`.
1. There are duplicate target names.
1. There seem to be missing certain e2e tests.

This may not be necessarily a problem for us who are intimately acquainted with our code base, but it will be a problem for any new contributor, both external, and potentially internal ones.

_I propose we make it even easier and clearer how to run e2e tests from the command line using one of the below approaches:_

1. **Rename each e2e test file to start with its directory name.**
    Pros:
    - No change necessary to _e2e-tests.sh_ script.
    - Filtering tests (e.g.  `safe-erc20*` - note the asterisk) becomes easier.
    Cons:
    - Long and repetitive file names.
2. **~~Update _e2e-tests.sh_ to run tests withing specific _packages_.~~**
    ~~Pros:~~
    ~~- File names remain unchanged.~~
    ~~Cons:~~
    ~~- Writing a wrong package name does not list valid packages.~~
    ```bash
    ~$ ./scripts/e2e-tests.sh non-existent-package
        Finished `release` profile [optimized] target(s) in 0.12s
    error: cannot specify features for packages outside of workspace
    ```
    ~~- Need to declare `std` feature in all examples.~~
    
I'm in favor of option 1) just because when a wrong target name is typed, the error log contains all valid targets.

### Contribution Guidelines

### LINKED ISSUE #444: [Feature]: Update _e2e-tests.sh_ Script to Allow Targeting Specific Test Cases Within Test Files

It is possible to update _e2e-tests.sh_ so that we're able to use it in the following ways:

```bash
# run every test
./scripts/e2e-tests.sh
# run only tests in files named 'erc721'
./scripts/e2e-tests.sh erc721
# run every test file starting with 'erc721' (erc721)
./scripts/e2e-tests.sh erc721*
# run only test cases that contain 'support_interface' in their name from within erc721 test files
./scripts/e2e-tests.sh erc721 -- support_interface
# run all tests, and pass custom flags/params to `cargo test`
./scripts/e2e-tests.sh * --nocapture
```

### Contribution Guidelines

### LINKED ISSUE #445: [Feature]: Shorten All Relevant Add/Sub Assignment Operations in Every Contract

The `AddAssignUnchecked` and `SubAssignUnchecked` traits are implemented for `StorageUint` (see [source](https://github.com/OpenZeppelin/rust-contracts-stylus/blob/9192c1849680f0c0f60822ac288b172a8b276c89/contracts/src/utils/math/storage.rs)).

This makes it possible to shorten integer state assignments from:

```rust
let total_supply_all = self._total_supply_all.get() - total_burn_value;
self._total_supply_all.set(total_supply_all);
```

To a one-liner:

```rust
self._total_supply_all.sub_assign_unchecked(total_burn_value);
```

We should update all similar state updates in all contracts.

### Contribution Guidelines

0xNeshi:
Resolves #445, Resolves #444 

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed (Breaking)
 
--
+- `Nonce::use_nonce` now panics on exceeding `U256::MAX`. #467
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Use `AddAssignUnchecked` and `SubAssignUnchecked` in `erc20::_update` calculations. #
+- Refactor how `AddAssignUnchecked` and `SubAssignUnchecked` are used in `Erc20`, `Erc721`, and `Erc721Consecutive`. #
 - Refactor `Nonces`. #
 
 ### Changed (Breaking)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,24 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- 
+
+### Changed
+
+- Use `AddAssignUnchecked` and `SubAssignUnchecked` in `erc20::_update` calculations. #
+
+### Changed (Breaking)
+
+-
+
+### Fixed
+
+-
+
 ## [v0.2.0-alpha.2] - 2024-12-18
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Refactor how `AddAssignUnchecked` and `SubAssignUnchecked` are used in `Erc20`, `Erc721`, and `Erc721Consecutive`. #467
-- Refactor `Nonces`. #467
+- Refactor how `AddAssignUnchecked` and `SubAssignUnchecked` traits are used in `Erc20`, `Erc721`, and `Erc721Consecutive`. #467
 
 ### Changed (Breaking)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Use `AddAssignUnchecked` and `SubAssignUnchecked` in `erc20::_update` calculations. #
+- Refactor `Nonces`. #
 
 ### Changed (Breaking)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Refactor how `AddAssignUnchecked` and `SubAssignUnchecked` traits are used in `Erc20`, `Erc721`, and `Erc721Consecutive`. #467
+- Use `AddAssignUnchecked` and `SubAssignUnchecked` in `erc20::_update` calculations. #467
 
 ### Changed (Breaking)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Refactor how `AddAssignUnchecked` and `SubAssignUnchecked` are used in `Erc20`, `Erc721`, and `Erc721Consecutive`. #
-- Refactor `Nonces`. #
+- Refactor how `AddAssignUnchecked` and `SubAssignUnchecked` are used in `Erc20`, `Erc721`, and `Erc721Consecutive`. #467
+- Refactor `Nonces`. #467
 
 ### Changed (Breaking)
 
--- a/contracts/src/access/control.rs
+++ b/contracts/src/access/control.rs
@@ -119,6 +119,7 @@ pub struct RoleData {
 #[storage]
 pub struct AccessControl {
     /// Role identifier -> Role information.
+    #[allow(clippy::used_underscore_binding)]
     pub _roles: StorageMap<FixedBytes<32>, RoleData>,
 }
 
--- a/contracts/src/finance/vesting_wallet.rs
+++ b/contracts/src/finance/vesting_wallet.rs
@@ -107,12 +107,16 @@ pub struct VestingWallet {
     /// [`Ownable`] contract.
     pub ownable: Ownable,
     /// Amount of Ether already released.
+    #[allow(clippy::used_underscore_binding)]
     pub _released: StorageU256,
     /// Amount of ERC-20 tokens already released.
+    #[allow(clippy::used_underscore_binding)]
     pub _erc20_released: StorageMap<Address, StorageU256>,
     /// Start timestamp.
+    #[allow(clippy::used_underscore_binding)]
     pub _start: StorageU64,
     /// Vesting duration.
+    #[allow(clippy::used_underscore_binding)]
     pub _duration: StorageU64,
     /// [`SafeErc20`] contract.
     pub safe_erc20: SafeErc20,
--- a/contracts/src/utils/math/storage.rs
+++ b/contracts/src/utils/math/storage.rs
@@ -1,5 +1,5 @@
 //! Simple math operations missing in `stylus_sdk::storage`.
-use alloy_primitives::{ruint::UintTryFrom, Uint};
+use alloy_primitives::Uint;
 use stylus_sdk::storage::StorageUint;
 
 /// Adds value and assign the result to `self`, ignoring overflow.
@@ -8,13 +8,11 @@ pub(crate) trait AddAssignUnchecked<T> {
     fn add_assign_unchecked(&mut self, rhs: T);
 }
 
-impl<T, const B: usize, const L: usize> AddAssignUnchecked<T>
+impl<const B: usize, const L: usize> AddAssignUnchecked<Uint<B, L>>
     for StorageUint<B, L>
-where
-    Uint<B, L>: UintTryFrom<T>,
 {
-    fn add_assign_unchecked(&mut self, rhs: T) {
-        let new_balance = self.get() + Uint::<B, L>::from(rhs);
+    fn add_assign_unchecked(&mut self, rhs: Uint<B, L>) {
+        let new_balance = self.get() + rhs;
         self.set(new_balance);
     }
 }
@@ -25,13 +23,11 @@ pub(crate) trait SubAssignUnchecked<T> {
     fn sub_assign_unchecked(&mut self, rhs: T);
 }
 
-impl<T, const B: usize, const L: usize> SubAssignUnchecked<T>
+impl<const B: usize, const L: usize> SubAssignUnchecked<Uint<B, L>>
     for StorageUint<B, L>
-where
-    Uint<B, L>: UintTryFrom<T>,
 {
-    fn sub_assign_unchecked(&mut self, rhs: T) {
-        let new_balance = self.get() - Uint::<B, L>::from(rhs);
+    fn sub_assign_unchecked(&mut self, rhs: Uint<B, L>) {
+        let new_balance = self.get() - rhs;
         self.set(new_balance);
     }
 }
--- a/contracts/src/utils/math/storage.rs
+++ b/contracts/src/utils/math/storage.rs
@@ -1,5 +1,5 @@
 //! Simple math operations missing in `stylus_sdk::storage`.
-use alloy_primitives::Uint;
+use alloy_primitives::{ruint::UintTryFrom, Uint};
 use stylus_sdk::storage::StorageUint;
 
 /// Adds value and assign the result to `self`, ignoring overflow.
@@ -8,11 +8,13 @@ pub(crate) trait AddAssignUnchecked<T> {
     fn add_assign_unchecked(&mut self, rhs: T);
 }
 
-impl<const B: usize, const L: usize> AddAssignUnchecked<Uint<B, L>>
+impl<T, const B: usize, const L: usize> AddAssignUnchecked<T>
     for StorageUint<B, L>
+where
+    Uint<B, L>: UintTryFrom<T>,
 {
-    fn add_assign_unchecked(&mut self, rhs: Uint<B, L>) {
-        let new_balance = self.get() + rhs;
+    fn add_assign_unchecked(&mut self, rhs: T) {
+        let new_balance = self.get() + Uint::<B, L>::from(rhs);
         self.set(new_balance);
     }
 }
@@ -23,11 +25,13 @@ pub(crate) trait SubAssignUnchecked<T> {
     fn sub_assign_unchecked(&mut self, rhs: T);
 }
 
-impl<const B: usize, const L: usize> SubAssignUnchecked<Uint<B, L>>
+impl<T, const B: usize, const L: usize> SubAssignUnchecked<T>
     for StorageUint<B, L>
+where
+    Uint<B, L>: UintTryFrom<T>,
 {
-    fn sub_assign_unchecked(&mut self, rhs: Uint<B, L>) {
-        let new_balance = self.get() - rhs;
+    fn sub_assign_unchecked(&mut self, rhs: T) {
+        let new_balance = self.get() - Uint::<B, L>::from(rhs);
         self.set(new_balance);
     }
 }
--- a/contracts/src/utils/nonces.rs
+++ b/contracts/src/utils/nonces.rs
@@ -35,6 +35,7 @@ pub enum Error {
 #[storage]
 pub struct Nonces {
     /// Mapping from address to its nonce.
+    #[allow(clippy::used_underscore_binding)]
     pub _nonces: StorageMap<Address, StorageU256>,
 }
 
--- a/contracts/src/utils/nonces.rs
+++ b/contracts/src/utils/nonces.rs
@@ -63,7 +63,7 @@ impl Nonces {
     ///
     /// # Safety
     ///
-    /// the function will experience [undefined behavior], if the nonce for
+    /// This function will experience [undefined behavior], if the nonce for
     /// the given `owner` has reached the maximum value representable by `U256`.
     /// Extreme caution should be taken to ensure overflow cannot occur.
     ///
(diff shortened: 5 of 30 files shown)


contracts/src/utils/nonces.rs
```diff
@@ -60,13 +60,17 @@ impl Nonces {
...
-    /// method to return `None`.
+    /// the function will experience **undefined behavior**, if the nonce for
+    /// the given `owner` has reached the maximum value representable by `U256`.
+    /// Extreme caution should be taken to ensure overflow cannot occur.
     pub fn use_nonce(&mut self, owner: Address) -> U256 {
         let nonce = self._nonces.get(owner);
+
+        // For each account, the nonce has an initial value of 0, can only be
+        // incremented by one, and cannot be decremented or reset. This
+        // guarantees that the nonce never overflows.
```
0xNeshi, line 73: Add comment from [Solidity version](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/c3b3ae7e82c6694c1c5c12e6eb832db55ef1e7fd/contracts/utils/Nonces.sol#L29)

scripts/e2e-tests.sh
```diff
@@ -1,14 +1,18 @@
...
+# Set RPC_URL environment variable
 export RPC_URL=http://localhost:8547
 
-cargo test --features std,e2e --test "$TEST_ARG"
+# Run tests based on arguments
+if [[ $# -eq 0 ]]; then
+    cargo test --features std,e2e --test "*"
+else
+    cargo test --features std,e2e "$@"
+fi
```
0xNeshi, line 16: The script can now be called in the following ways:

- if no arguments are passed, the behavior is the same as before
- otherwise, pass whatever arguments are present to `cargo test` directly

We can now easily filter for even unit tests with this script. 

This should cover most of our testing needs.

NOTE: feel free to suggest or add any additional feature you think might be useful.

0xNeshi: An other option is to omit the if-statement, and just call `cargo test --features std,e2e "$@"`, passing whatever arguments to the command. In our e2e workflow, we'd then call the script like `./scripts/e2e-tests.sh --test "*"`.

At this point it's really a matter of taste

bidzyyys (commented):
I believe we can apply `#[allow(clippy::used_underscore_binding)]` for the whole contracts library

CHANGELOG.md
```diff
@@ -5,6 +5,25 @@ All notable changes to this project will be documented in this file.
...
 
+## [Unreleased]
+
+### Added
+
+- 
+
+### Changed
+
+- Refactor how `AddAssignUnchecked` and `SubAssignUnchecked` are used in `Erc20`, `Erc721`, and `Erc721Consecutive`. #467
```
bidzyyys, line 16:
```suggestion
- Refactor how `AddAssignUnchecked` and `SubAssignUnchecked` traits are used in `Erc20`, `Erc721`, and `Erc721Consecutive`. #467
```

CHANGELOG.md
```diff
@@ -5,6 +5,25 @@ All notable changes to this project will be documented in this file.
...
+- Refactor `Nonces`. #467
```
bidzyyys, line 17: IMHO if anything in API has changed than we should mention this change. If not, no need.

contracts/src/utils/nonces.rs
```diff
@@ -60,13 +61,19 @@ impl Nonces {
     /// * `&mut self` - Write access to the contract's state.
     /// * `owner` - The address for which to consume the nonce.
     ///
-    /// # Panics
+    /// # Safety
     ///
-    /// This function will panic if the nonce for the given `owner` has reached
-    /// the maximum value representable by `U256`, causing the `checked_add`
-    /// method to return `None`.
+    /// the function will experience [undefined behavior], if the nonce for
```
bidzyyys, line 66:
```suggestion
    /// This function will experience [undefined behavior], if the nonce for
```

contracts/src/utils/nonces.rs
```diff
@@ -60,13 +61,19 @@ impl Nonces {
...
+    /// the given `owner` has reached the maximum value representable by `U256`.
+    /// Extreme caution should be taken to ensure overflow cannot occur.
+    ///
+    /// [undefined behavior]: https://doc.rust-lang.org/reference/behavior-considered-undefined.html
     pub fn use_nonce(&mut self, owner: Address) -> U256 {
         let nonce = self._nonces.get(owner);
+
+        // For each account, the nonce has an initial value of 0, can only be
```
bidzyyys, line 74:
```suggestion
        // For each account, the nonce has an initial value of `U256::ZERO`, can only be
```

contracts/src/utils/nonces.rs
```diff
@@ -83,34 +90,32 @@ impl Nonces {
     /// * `owner` - The address for which to consume the nonce.
     /// * `nonce` - The nonce to consume.
     ///
-    /// # Panics
+    /// # Safety
     ///
-    /// This function will panic if the nonce for the given `owner` has reached
-    /// the maximum value representable by `U256`, causing the `checked_add`
-    /// method to return `None`.
+    /// the function will experience [undefined behavior], if the nonce for
```
bidzyyys, line 95:
```suggestion
    /// This function will experience [undefined behavior], if the nonce for
```

contracts/src/utils/nonces.rs
```diff
@@ -60,13 +61,19 @@ impl Nonces {
...
+    /// [undefined behavior]: https://doc.rust-lang.org/reference/behavior-considered-undefined.html
     pub fn use_nonce(&mut self, owner: Address) -> U256 {
         let nonce = self._nonces.get(owner);
+
+        // For each account, the nonce has an initial value of 0, can only be
+        // incremented by one, and cannot be decremented or reset. This
+        // guarantees that the nonce never overflows.
         self._nonces
             .setter(owner)
             .set(unsafe { nonce.checked_add(ONE).unwrap_unchecked() });
```
bidzyyys, line 79: Tbh I do not like code where we use `unsafe` in our contracts. Is it possible to apply panicking operation before the setter?

bidzyyys (commented):
Left some comments, good job @0xNeshi

CHANGELOG.md
0xNeshi, line 17:
```suggestion
```

contracts/src/access/control.rs
```diff
@@ -119,6 +119,7 @@ pub struct RoleData {
 #[storage]
 pub struct AccessControl {
     /// Role identifier -> Role information.
+    #[allow(clippy::used_underscore_binding)]
```
0xNeshi, line 122: In response to :
> I believe we can apply `#[allow(clippy::used_underscore_binding)]` for the whole contracts library

We would then not get warnings for:

```rust
let _x = 1u32;
let y = _x + 1; // should be a warning
```

contracts/src/utils/nonces.rs
0xNeshi: I don't really understand how this "guarantees that the nonce never overflows", but that's what's in the Solidity version.

How bad performance-wise can it be if we _did_ check for overflows here?

bidzyyys: I believe we should have panic here to make it obvious that it may fail, like:
```rust
let updated_nonce = nonce.checked_add(ONE).expect("...");
 self._nonces
            .setter(owner)
            .set(updated_nonce);
```

bidzyyys: And then we need to have `# Panics` sections in Rust docs for proper fns.

contracts/src/access/control.rs
bidzyyys: Fair enough. I am ok with leaving as it is.

bidzyyys (commented):
Almost ready to merge!

contracts/src/access/control.rs
0xNeshi: It would be useful to have an attribute similar to [`[serde(rename = ...)]`](https://serde.rs/field-attrs.html#rename), so we could use Rust conventions for naming fields, and just rename those fields when building the contract :thinking:

0xNeshi: Actually, are we really required to use the same names as in Solidity?

It is a Solidity convention to prefix private state fields with an underscore `_`, but this is not the case in Rust. 

Why can't we just omit the underscore?

0xNeshi: @bidzyyys @qalisander

contracts/src/utils/nonces.rs
0xNeshi: @bidzyyys @qalisander I think this was mentioned before, but it would be useful to have "checked" add/sub assign versions in [math/storage.rs](https://github.com/OpenZeppelin/rust-contracts-stylus/blob/59c1ead477e969e0d85b27ac8784c760acbb8ca0/contracts/src/utils/math/storage.rs):

```rust
trait AddAssignChecked
trait SubAssignChecked
```

contracts/src/access/control.rs
bidzyyys: @0xNeshi this is the corresponding issue #386 that we have. Tbh I do not like either as it is against Rust attitude. Right now I would merge as it is and then focus on renaming based on our common agreement.

contracts/src/finance/vesting_wallet.rs
```diff
@@ -107,12 +107,16 @@ pub struct VestingWallet {
     /// [`Ownable`] contract.
     pub ownable: Ownable,
     /// Amount of Ether already released.
+    #[allow(clippy::used_underscore_binding)]
```
bidzyyys, line 110: I can see some inconsistency in attributes' naming, e.g. `ownable` & `_released`. We should fix it with #386.

scripts/e2e-tests.sh
bidzyyys: Let's leave right now the current behaviour.

bidzyyys (approved):
LGTM!

contracts/src/utils/nonces.rs
bidzyyys: @0xNeshi yeah, having "safe" (`Checked`) operation would be great. If you want to include those traits in this PR go ahead; If not feel free to add an issue to this milestone.

scripts/e2e-tests.sh
```diff
@@ -1,14 +1,16 @@
 #!/bin/bash
 set -e
 
-TEST_ARG="${1:-*}"
-
-MYDIR=$(realpath "$(dirname "$0")")
-cd "$MYDIR"
-cd ..
+# Navigate to project root
+cd "$(dirname "$(realpath "$0")")/.."
```
qalisander, line 5: Consider that reduction of code lines, doesn't reduce code complexity.
After all we could feed all code as a single line to compiler and he will be fine.

p.s. If you think that new changes are better lets keep them

examples/erc20/src/lib.rs
```diff
@@ -110,6 +110,10 @@ impl Erc20Example {
             || Erc20Metadata::supports_interface(interface_id)
     }
 
+    /// WARNING: These functions are intended for **testing purposes** only. In
```
qalisander, line 113: ++

contracts/src/utils/math/storage.rs
```diff
@@ -8,11 +8,13 @@ pub(crate) trait AddAssignUnchecked<T> {
...
-impl<const B: usize, const L: usize> AddAssignUnchecked<Uint<B, L>>
+impl<T, const B: usize, const L: usize> AddAssignUnchecked<T>
     for StorageUint<B, L>
+where
+    Uint<B, L>: UintTryFrom<T>,
 {
-    fn add_assign_unchecked(&mut self, rhs: Uint<B, L>) {
-        let new_balance = self.get() + rhs;
+    fn add_assign_unchecked(&mut self, rhs: T) {
+        let new_balance = self.get() + Uint::<B, L>::from(rhs);
```
qalisander, line 17: This refactoring looks smart.
But initial reason why we are rarely using `Uint::<_,_>::from(..)` method, that it performs conversion in runtime.

In the following example a conversion is done at compile time, with the help of `uint!` macro.

```
self._balances.setter(from).sub_assign_unchecked(uint!(1_U256))
```

Also `Uint::<B, L>::from` has actual conversion code beneath with size checks, and we have `add_unsigned_unchecked` function name

contracts/src/access/control.rs
qalisander: Don't like `_*` either

qalisander (commented):
Good job!

contracts/src/access/control.rs
0xNeshi: Seems the agreement is there :smile: 
Nevertheless, I'd prefer this be done as part of a separate PR linked to the [issue you mentioned](https://github.com/OpenZeppelin/rust-contracts-stylus/issues/386)

contracts/src/utils/math/storage.rs
0xNeshi: This is actually a very good catch!
I'll revert this, and try to think of a way to keep this generic, but won't spend too much time on it.

scripts/e2e-tests.sh
0xNeshi: To me it's the same, as the dirname+realpath combo is (un)clear enough.

Though now that I think about it, wouldn't it be enough to just do `cd ..`? :thinking:

contracts/src/utils/nonces.rs
0xNeshi: Removed

scripts/e2e-tests.sh
0xNeshi: Basically, it lets you call this script as if you were to call `cargo test`, meaning it enables you to do all the funky filtering stuff

contracts/src/utils/nonces.rs
0xNeshi: I'll add a separate issue, as there are many contracts where the change could be applied

contracts/src/utils/math/storage.rs
0xNeshi: Wait, does this mean that `uint!` is more performant than `U256::from`?

contracts/src/utils/nonces.rs
0xNeshi: Here's the issue https://github.com/OpenZeppelin/rust-contracts-stylus/issues/469

scripts/e2e-tests.sh
0xNeshi: All of these are now supported

```bash
# run all tests (appends `--test "*"` as before)
./scripts/e2e-tests.sh
# pass arguments to `cargo test`
./scripts/e2e-tests.sh --test erc721
./scripts/e2e-tests.sh --no-run
# pass arguments both to `cargo test` and to test binary
./scripts/e2e-tests.sh --test erc721 -- support_interface
./scripts/e2e-tests.sh erc721 --no-run -- --nocapture
```

qalisander: I see.
But in this case it is impossible to invoke the command like `./scripts/e2e-tests.sh erc721 error_when_minting_token_id_twice`

If you would change it like this:
```
if [[ $# -eq 0 ]]; then
    cargo test --features std,e2e --test "*"
else
    cargo test --features std,e2e --test "$@"
fi
```
It can work.

qalisander: And we should spell it like `./scripts/e2e-tests.sh --test erc721 error_when_minting_token_id_twice`

qalisander: But actually what I like very much in your approach, is that if we spell `./scripts/e2e-tests.sh --help` it will print docs for cargo sub commands. So no need to bother ourselves with script's documentation

scripts/e2e-tests.sh
qalisander: I see it was necessary before:D or just a good tone of writing repo wide scripts. Step to root directory first

contracts/src/utils/math/storage.rs
qalisander: Instead of adding `ONE`, we first "create `ONE` from integer checking that integer is not bigger. If it is bigger then panic, if not" add ONE to the value

qalisander: Ofc it won't affect performance dramatically and we wont notice it probably. But I don't see much point of that kind of refactoring. We don't win much in code simplicity.

scripts/e2e-tests.sh
0xNeshi: The current implementation is the simplest one, with the most straightforward way to use.

Updating the script to support invocations like:
```bash
./scripts/e2e-tests.sh erc721 error_when_minting_token_id_twice
```
Would make it very complex if we also wanted to support passing other valid args/flags to `cargo test`, or to the test binary.

> But actually what I like very much in your approach, is that if we spell ./scripts/e2e-tests.sh --help it will print docs for cargo sub commands. So no need to bother ourselves with script's documentation

Can you clarify the last sentence?

scripts/e2e-tests.sh
qalisander: Lets use then this construct in other scripts
```
# Move to project root
cd "$(dirname "$(realpath "$0")")/.."
```

scripts/e2e-tests.sh
qalisander: Let's keep you changes

scripts/e2e-tests.sh
0xNeshi: My bad, it is actually better to ensure we're in the project root :sweat_smile: 

I don't mind reverting this to the way it was before, to me it's the same. Will do so.

contracts/src/utils/math/storage.rs
0xNeshi: You're right, reverted the change https://github.com/OpenZeppelin/rust-contracts-stylus/pull/467/commits/c3a1599b2397eda8aa71bf29f2f671594c66e7d0

scripts/e2e-tests.sh
qalisander: > Can you clarify the last sentence?
I mean, that we can print valid documentation for `e2e-test.sh` script, since we're just forwarding arguments to `cargo test` like:
```./scripts/e2e-tests.sh --help```

0xNeshi: Does the script then have your blessing? :smile:

qalisander (approved):
LGTM

bidzyyys (approved):
LGTM! 🚀

PR #474, opened 2025-01-03
feat: implement trait AddAssignChecked for StorageUint

### LINKED ISSUE #469: [Feature]: Add "Checked" Versions of [Add/Sub]Assign Traits

Add traits similar to existing [[Add/Sub]AssignUnchecked](https://github.com/OpenZeppelin/rust-contracts-stylus/blob/3205fc7a9bff59af753d4fef9253343ae2c631b8/contracts/src/utils/math/storage.rs), but "checked" versions:

```rust
trait AddAssignChecked
trait SubAssignChecked
```

This would enable us to change code like:

```rust
let updated_nonce = nonce.checked_add(uint!(1_U256)).expect("error message");
self._nonces.setter(owner).set(updated_nonce);
```

Into:

```rust
self._nonces.setter(owner).add_assign_checked(uint!(1_U256), "error message");
```

### Contribution Guidelines

bidzyyys:
Resolves #469

#### PR Checklist

- [x] Tests
- [ ] Documentation
- [x] Changelog


bidzyyys:
@qalisander @0xNeshi I decided to remove `SubAssignChecked` trait in https://github.com/OpenZeppelin/rust-contracts-stylus/pull/474/commits/1d7def31bd78be779204ae0ad26be35e132915f0 as it was not needed in our library and I am not a fan of maintaining "dead code"

contracts/src/finance/vesting_wallet.rs
```diff
@@ -421,12 +422,10 @@ impl IVestingWallet for VestingWallet {
...
 
-        let released = self
-            ._released
-            .get()
-            .checked_add(amount)
-            .expect("total released should not exceed `U256::MAX`");
-        self._released.set(released);
+        self._released.add_assign_checked(
+            amount,
+            "total released should not exceed `U256::MAX`",
```
qalisander, line 427: Idk.. Usually that kind api in rust`*_checked` returns `Option`

qalisander: But returning `Option` seems pointless, since we are not assumed to return a value

0xNeshi: > But returning Option seems pointless, since we are not assumed to return a value

:point_up: 

1. **We don't have a use-case where the new add-assigned value is used afterwards.** If this were to change, I'd be in favor of returning a value.
2. **AFAIK add/sub-assign is always a non-returning operation in most popular languages.** It would make sense to keep this consistent with the helper too.
3. **Returning a value would add an additional if-check.**
    1. Call `checked_add`.
    2. _Check if returned value is `Some`, and then set the new value; otherwise return `None`._
    3. Call `expect` on the returned `Option`.

0xNeshi (approved):
LGTM

contracts/src/finance/vesting_wallet.rs
bidzyyys: Okay, so merging as it is.

bidzyyys: We may change it if needed.

qalisander (approved):
LGTM!

PR #477, opened 2025-01-09
test: Investigate Fuzzing

### LINKED ISSUE #458: [Feature]: Investigate Fuzzing

- [x] find which, if any, contracts/functions are good candidates for fuzzing
- [x] determine whether we can use any existing library for fuzzing, or if we need to create this logic ourselves
- [ ] implement fuzzing for a single function (to be used as template for others)

### Contribution Guidelines

0xNeshi:
As [per our agreement](https://openzeppelin.slack.com/archives/C06R5AGFE9G/p1737021738986409), this PR includes a basic fuzz tests setup and some missing unit/property tests for `openzeppelin-crypto` crate.

I created a couple of related issues to address missing property tests (see #501 #502 #503).
Adding property tests for contract will be easily possible after `motsu` is refactored (see https://github.com/OpenZeppelin/stylus-test-helpers/pull/14/files#r1919897820)

Resolves #458 

#### PR Checklist

- [x] Tests
- [x] Documentation
- [ ] Changelog


fuzz/.gitignore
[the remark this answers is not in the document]
0xNeshi: No need. The fuzzer generates these dirs in the root of its project, and uses target bin names as subdir names:
```
[[bin]]
name = "keccak_doesnt_panic" # this is the subdir name
path = "fuzz_targets/keccak/no_panics.rs"
```

fuzz/Cargo.toml
```diff
@@ -0,0 +1,53 @@
+[package]
+name = "test-fuzz-fuzz"
+version = "0.0.0"
```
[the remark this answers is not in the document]
0xNeshi: No can do. The fuzz project must be standalone.

fuzz/Cargo.toml
```diff
@@ -0,0 +1,53 @@
...
+[dependencies.openzeppelin-crypto]
+path = "../lib/crypto"
+
+[dependencies.openzeppelin-stylus]
+path = "../contracts"
+
+[[bin]]
+name = "merkle_verify"
+path = "fuzz_targets/merkle_verify.rs"
+test = false
```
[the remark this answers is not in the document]
0xNeshi: I don't think _Cargo.toml_ supports this. 
Do you know of a way this can be done?

fuzz/fuzz_targets/openzeppelin_stylus_utils.rs
```diff
@@ -0,0 +1,27 @@
...
+    _ = openzeppelin_stylus::utils::cryptography::ecdsa::recover(
+        &mut storage,
+        hash,
+        v,
+        r,
+        s,
+    );
+});
+
+// TODO: add missing shims
```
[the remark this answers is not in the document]
0xNeshi: there are dozens of missing shims, and the solution will be extremely fragile. I think we should wait on Host refactors on the Stylus SDK side before we implement fuzzing for contracts.

bidzyyys:
@0xNeshi we should consider fuzzing for this folder -> https://github.com/OpenZeppelin/rust-contracts-stylus/tree/main/contracts/src/utils/math

0xNeshi:
> @0xNeshi we should consider fuzzing for this folder -> https://github.com/OpenZeppelin/rust-contracts-stylus/tree/main/contracts/src/utils/math

Definitely :100: We will [need shims](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/477#discussion_r1923659866) for this to work though, I'd suggest we hold off on this for now

fuzz/fuzz_targets/openzeppelin_stylus_utils.rs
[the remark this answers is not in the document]
bidzyyys: No need to implement them now, let's wait for `MockHost` from Stylus SDK.

fuzz/Cargo.toml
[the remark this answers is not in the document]
bidzyyys: Unfortunately not :(

bidzyyys (approved):
LGTM!

lib/crypto/src/keccak.rs
```diff
@@ -45,3 +45,225 @@ impl Hash for [u8; 32] {
...
     }
 }
+
+#[cfg(all(test, feature = "std"))]
+mod tests {
+    use proptest::prelude::*;
+
+    use super::*;
+
+    proptest! {
```
qalisander, line 55: I see, that we are mostly testing here how keccak is implemented in the other crate (tiny-keccak). That it can produce the same hash, meets consistency and distribution requirement. Shouldn't these tests be implemented [there](https://github.com/debris/tiny-keccak/tree/master/tests) already?

lib/crypto/src/bits.rs
```diff
@@ -32,14 +32,115 @@ impl_bit_iter_be!(usize);
 
 #[cfg(all(test, feature = "std"))]
 mod tests {
+    use proptest::prelude::*;
+
     use super::*;
 
+    proptest! {
```
qalisander, line 39: The function you testing here is a one-liner: `(0..<$int>::BITS).rev().map(move |i| self & (1 << i) != 0)`
And we have 100 lines of test cases. 
I'm scared of progression. If we would add another simple one-liner `BitIteratorLE` for little-endian byte iteration. Will we write the same number of new test cases?
And these test cases are still checking partial scenarios.

Can't we have a single proptest, that calls std api's bit counters: `count_zeros`, `count_ones`, `leading_zeros` and `leading_ones`. And compare with booleans counted from `bit_be_trimmed_iter` and `bit_be_iter`?

qalisander: And belief, that `BitIteratorBE` will be also implicitly checked by finite fields and posseidon tests:D

lib/crypto/src/keccak.rs
0xNeshi: [You'd think](https://github.com/debris/tiny-keccak/blob/master/tests/keccak.rs)

0xNeshi: It's remarkably low on tests for such a popular crate

lib/crypto/src/bits.rs
0xNeshi: I may have gone overboard with tests, but I don't think you can say any of them are redundant.

Rarely is it the case that you can 100% test one layer of implementation with tests for an upper layer of implementation (e.g. test `BitIteratorBE` with tests for poseidon). There is great value in testing at least a known flow or two, and edge cases that are unlikely to be invoked by the upper layers.

0xNeshi: You gain confidence that no one in the future can accidentally break your apparently simple code

0xNeshi: Also this one-liner:
```rust
(0..<$int>::BITS).rev().map(move |i| self & (1 << i) != 0)
```
looks like a spell invocation. I think it warrants a couple of tests :D

qalisander: Okay:D We can discuss our approach on testing for `openzeppelin-crypto`. 
I would prefer simple tests (non proptests) to be not extensive, but be like a smoke tests. And rely more on proptests as error proofs.

qalisander: Besides you forgot here to test U64, U128, U192, U256, U384 and so on... With this approach we would have a LOT of tests.

qalisander: Or whether consider macros for testing

lib/crypto/src/bits.rs
```diff
@@ -32,14 +32,116 @@ impl_bit_iter_be!(usize);
...
+
+    #[test]
+    fn trimmed_is_subset_of_full() {
+        proptest!(|(value: u64)| {
+            let full: Vec<bool> = value.bit_be_iter().collect();
+            let trimmed: Vec<bool> = value.bit_be_trimmed_iter().collect();
+            let start_idx = value.leading_zeros() as usize;
+            prop_assert_eq!(&full[start_idx..], trimmed);
+        })
+    }
```
0xNeshi, line 56: @qalisander is this better for your IDE than when all tests are grouped in one `proptest!`? You mentioned that this way you're able to run individual tests from within IDE

qalisander (approved):
LGTM!

lib/crypto/src/keccak.rs
0xNeshi: Opened an issue asking the maintainers of tiny-keccak to let us add more tests, see https://github.com/debris/tiny-keccak/issues/59

PR #525, opened 2025-02-05
ref: update bit tests

qalisander:


qalisander:
I would leave as few tests as it is here. @0xNeshi wdt?

0xNeshi (approved):
Excellent!

PR #526, opened 2025-02-06
ref: make all proptests stand-alone + replace prop_assume calls with helper fn

0xNeshi:
Making all prop tests stand-alone is helpful as it enables devs using certain IDEs to run each test individually by clicking an appropriate button that appears next to individual tests


qalisander (approved):
LGTM!

bidzyyys (approved):
LGTM!

PR #449, opened 2024-12-11
refactor permit

0xNeshi:
Resolves #???

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog

PR #552, opened 2025-02-18
feat: implement Deref for Erc20Permit and Ownable2Step

0xNeshi:
This will allow devs to implement a `permit + burnable` combination without accessing the nested `Erc20Permit.erc20` instance.

Implementing `Deref`:
- makes the implementation more intuitive.
- makes implementing Wizard's function printing logic more straightforward ([current](https://github.com/OpenZeppelin/contracts-wizard/blob/76dee5295d843c227b0d24c4f78cf62d1933d427/packages/core/stylus/src/erc20.ts#L220-L225) VS [if this PR is merged](https://github.com/OpenZeppelin/contracts-wizard/blob/b536968c1a3b3ba91c5f2217b22ffa946cc7d789/packages/core/stylus/src/erc20.ts#L218-L223))

```rust
// other imports...
use openzeppelin_stylus::{
    token::erc20::extensions::{Erc20Permit, IErc20Burnable},
    utils::cryptography::eip712::IEip712,
};

#[entrypoint]
#[storage]
struct Erc20PermitExample {
    #[borrow]
    pub erc20_permit: Erc20Permit<Eip712>,
}

// define Eip712...

#[public]
#[inherit(Erc20Permit<Eip712>)]
impl Erc20PermitExample {
    pub fn burn(&mut self, value: U256) -> Result<(), Vec<u8>> {
        // before:
        // self.erc20_permit.erc20.burn(value).map_err(|e| e.into())
        // now:
        self.erc20_permit.burn(value).map_err(|e| e.into())
    }

    pub fn burn_from(
        &mut self,
        account: Address,
        value: U256,
    ) -> Result<(), Vec<u8>> {
        // before:
        // self.erc20_permit.erc20.burn_from(account, value).map_err(|e| e.into())
        // now:
        self.erc20_permit.burn_from(account, value).map_err(|e| e.into())
    }
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
--
+- Implement `Deref<Target = Ownable>` for `Ownable2Step` and `Deref<Target = Erc20>` for `Erc20Permit`. #552
 
 ### Changed (Breaking)
 
--- a/contracts/src/access/ownable_two_step.rs
+++ b/contracts/src/access/ownable_two_step.rs
@@ -17,6 +17,7 @@
 //! available.
 
 use alloc::vec::Vec;
+use core::ops::{Deref, DerefMut};
 
 use alloy_primitives::Address;
 pub use sol::*;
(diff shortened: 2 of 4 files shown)


bidzyyys:
@0xNeshi I believe we should take another look on `Erc20Permit` and other extensions, and make them consistent...

0xNeshi:
> @0xNeshi I believe we should take another look on `Erc20Permit` and other extensions, and make them consistent...

You mean like [this](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/449/files)?

I think that would be much better (and make Wizard logic less complex), but I remember we agreed not to "touch" Permit contract since we'll have to update it anyway once multiple `#[public]`s are supported. Though I wouldn't be against making the change even now, as it can be done rather quickly, but do note that it would be a breaking change.

bidzyyys:
> > @0xNeshi I believe we should take another look on `Erc20Permit` and other extensions, and make them consistent...
> 
> You mean like [this](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/449/files)?
> 
> I think that would be much better (and make Wizard logic less complex), but I remember we agreed not to "touch" Permit contract since we'll have to update it anyway once multiple `#[public]`s are supported. Though I wouldn't be against making the change even now, as it can be done rather quickly, but do note that it would be a breaking change.

As the SDK should support soon better contract composition I think we should wait for it first.

qalisander (approved):
LGTM

PR #553, opened 2025-02-18
test: Add Differential Fuzz Test for _merkle.rs_

### LINKED ISSUE #502: [Feature]: Add Differential Tests for Functions in _merkle.rs_

Towards #520 

See how Solidity's Foundry defines [differential testing](https://book.getfoundry.sh/forge/differential-ffi-testing).

Differential tests would compare the results of our merkle verification implementation against some third-party implementation. This would help pinpoint any latent issues, bugs etc.

One possible candidate for this is [rs-merkle](https://github.com/antouhou/rs-merkle), but we can search for other implementations if they're more robust/well-known

### Contribution Guidelines

Haxry:
Resolves #502 

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog


fuzz/Cargo.toml
```diff
@@ -13,6 +13,7 @@ motsu = "0.3.0"
 stylus-sdk = { version = "0.7.0", default-features = false, features = [
     "mini-alloc",
 ] }
+rs_merkle = { git = "https://github.com/Haxry/rs-merkle.git", branch = "sha3" }
```
0xNeshi, line 16: Saw you [opened a PR](https://github.com/antouhou/rs-merkle/pull/42), kudos to you sir! :rocket: Doing open-sourcing the way it's supposed to :fire: 

Until these changes are part of the main crate, we'd prefer to have the Sha3 implementation within our repository if possible. Can't we implement the [same algorithm](https://github.com/antouhou/rs-merkle/pull/42/files#diff-a0b0e495bc25c7683ee3b9293744068dbba678401228ed36642210e4af6ef845R5) within our _fuzz_ project?

fuzz/fuzz_targets/diff_merkle_verify.rs
```diff
@@ -0,0 +1,49 @@
...
+use libfuzzer_sys::fuzz_target;
+use openzeppelin_crypto::merkle::Verifier;
+use rs_merkle::{MerkleTree, algorithms::Keccak256, Hasher};
+
+fuzz_target!(|data: &[u8]| {
+    if data.len() < 33 {
+        return;
+    }
+
+
```
0xNeshi, line 11: Format this file with `cargo +nightly fmt`

fuzz/fuzz_targets/diff_merkle_verify.rs
```diff
@@ -0,0 +1,49 @@
...
+fuzz_target!(|data: &[u8]| {
+    if data.len() < 33 {
+        return;
+    }
+
+    
+    let mut leaves = Vec::new();
+    for chunk in data[1..].chunks(32) {
+        if chunk.len() == 32 {
```
0xNeshi, line 14: Didn't you already ensure that chunks are 32 bytes in length?

fuzz/fuzz_targets/diff_merkle_verify.rs
```diff
@@ -0,0 +1,49 @@
...
+            leaf.copy_from_slice(chunk);
+            leaves.push(leaf);
+        }
+    }
+
+    if leaves.len() < 2 {
+        return;
+    }
+
+    println!("Leaves: {:?}", leaves);
```
0xNeshi, line 25: Remove `println!`s before merging

fuzz/fuzz_targets/diff_merkle_verify.rs
```diff
@@ -0,0 +1,49 @@
...
+    let index_to_prove = data[0] as usize % leaves.len();
+    println!("index to prove {}", index_to_prove);
+    let rs_proof = rs_merkle_tree.proof(&[index_to_prove]);
+
+    
+    let oz_proof: Vec<[u8; 32]> = rs_proof.proof_hashes().iter().map(|h| h.to_owned()).collect();
+    println!("Proof: {:?}", oz_proof);
+
+    
+    let oz_verification = Verifier::verify(&oz_proof, rs_root, leaves[index_to_prove]);
```
0xNeshi, line 39: Could you try to include a `Verifier::verify_multi_proof` call with the same data?

It would be awesome if we could check both functions, but if only one is possible per fuzz test, then this is fine and we can leave a fuzz test for `Verifier::verify_multi_proof` as part of a separate issue.

fuzz/fuzz_targets/diff_merkle_verify.rs
```diff
@@ -0,0 +1,49 @@
...
+
+    
+    assert_eq!(
+        oz_verification,
+        rs_proof.verify(rs_root, &[index_to_prove], &[leaves[index_to_prove]], leaves.len()),
+        "Verification mismatch between rs-merkle and OpenZeppelin"
+    );
```
0xNeshi, line 46: Hmm considering you used `rs_merkle` to construct a valid root and a valid proof, shouldn't it be safe to just check that `Verifier::verify` returns `true`?

0xNeshi (commented):
Very good work! Left some minor comments

fuzz/fuzz_targets/diff_merkle_verify.rs
```diff
@@ -0,0 +1,49 @@
...
+        if chunk.len() == 32 {
+            let mut leaf = [0u8; 32];
+            leaf.copy_from_slice(chunk);
+            leaves.push(leaf);
+        }
+    }
+
+    if leaves.len() < 2 {
+        return;
+    }
```
0xNeshi, line 23:
```suggestion
    // Need at least 65 bytes: 1 for index + 2 * 32 for minimum leaves
    if data.len() < 65 {
        return;
    }

    // Pre-allocate vector with exact capacity needed
    let mut leaves = Vec::with_capacity(data[1..].len() / 32);
    
    // Collect full chunks only
    for chunk in data[1..].chunks_exact(32) {
        let mut leaf = [0u8; 32];
        leaf.copy_from_slice(chunk);
        leaves.push(leaf);
    }
```
more efficient

fuzz/fuzz_targets/diff_merkle_verify.rs
```diff
@@ -0,0 +1,49 @@
...
+
+    println!("Leaves: {:?}", leaves);
+
+    
+    let rs_merkle_tree = MerkleTree::<Keccak256>::from_leaves(&leaves);
+    let rs_root:[u8; 32] = rs_merkle_tree.root().unwrap();
```
0xNeshi, line 29: What if this panics? The fuzz test might register a false failure.
Better to return if `None` is the result.

fuzz/fuzz_targets/diff_merkle_verify.rs
```diff
@@ -0,0 +1,49 @@
+#![no_main]
+use libfuzzer_sys::fuzz_target;
+use openzeppelin_crypto::merkle::Verifier;
+use rs_merkle::{MerkleTree, algorithms::Keccak256, Hasher};
+
+fuzz_target!(|data: &[u8]| {
+    if data.len() < 33 {
+        return;
+    }
```
0xNeshi, line 9: Instead of using the first data byte to derive a leaf index to check, why not just let the fuzzer generate a random index?

Using an explicit struct (clearer):
```rust
#[derive(Debug, arbitrary::Arbitrary)]
struct FuzzInput {
    leaves_data: Vec<u8>,
    index_byte: u8,
}

fuzz_target!(|input: FuzzInput| {
```

Or just adding an additional input field (might be sufficient, no need to derive `Arbitrary`):
```rust
fuzz_target!(|data: (&[u8], u8)| {
    let (data, index) = data;
    
    // no longer checking `< 33`...
    if data.len() < 32 {
        return;
    }
```

0xNeshi:
@Haxry hey checking in, are you still interested in working on this?

lib/crypto/src/merkle.rs
```diff
@@ -779,4 +783,26 @@ mod tests {
         );
         assert!(verification.is_err());
     }
+
+    #[test]
+    fn verify_empty_proof_should_mean_leaf_equal_to_root() {
```
0xNeshi, line 788: missing edge case test

lib/crypto/src/merkle.rs
```diff
@@ -749,23 +751,25 @@ mod tests {
...
+        // const keccak256 = require('keccak256'); // v1.0.6
         //
-        // // Now we can pass any **malicious** fake leaves as valid!
-        // const maliciousLeaves = ['malicious', 'leaves'].map(ethers.id)
-        //                          .map(ethers.toBeArray).sort(Buffer.compare);
-        // const maliciousProof = [leave, leave];
-        // const maliciousProofFlags = [true, true, false];
+        // const leaves = [keccak256('real leaf'), Buffer.alloc(32, 0)];
+        // const merkleTree = new MerkleTree(leaves, keccak256, { sortPairs: true });
+        // const root = merkleTree.getRoot();
```
0xNeshi, line 764: I copy/pasted the exact example used in [Solidity](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/21c8312b022f495ebe3621d5daeed20552b43ff9/test/utils/cryptography/MerkleProof.test.js#L184) to ensure validity, so I updated the code comment block to the one I was able to use to reproduce it.

0xNeshi (approved):

0xNeshi (approved):
- removed redundant fuzz test scaffolds
- added merkle corpus generator
- updated merkle unit tests
- add fuzz.yml workflow

bidzyyys (commented):
@0xNeshi you should add CI job to run fuzz tests triggered on either release branch creation or clicking a button.

bidzyyys (approved):

qalisander (approved):
LGTM

PR #611, opened 2025-03-17
ref: use nitro-devnode instead of nitro-testnode

### LINKED ISSUE #609: [Refactor]: Optimize E2E tests and benchmarks with nitro-devnode
- https://github.com/OffchainLabs/nitro-devnode
- https://docs.arbitrum.io/run-arbitrum-node/run-nitro-dev-node

bidzyyys:
Resolves #609 

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [x] Changelog


0xNeshi:
I concluded that the devnode is unsuitable for our e2e workflow - no matter how I reconfigured the e2e setup, a subset of tests start failing, with the subset being different each run, and the reason for this boils down to: nitro devnode was created for quick local prototyping, so it doesn't behave like a regular local blockchain node, which invalidates some of our current assumptions in the context of our e2e.
- if the number of test threads is larger than 1, the chance of a test failing increases (usually with `ProgramNotActivated` and occasionally with `ContractAlreadyDeployed` (in another test) errors).
- gas costs are extremely unpredictable (in all my test runs I tried relying on `wallet.estimate_gas` but it seemed to underestimate the gas amount by 2-3x).
- master signer seems to run out of funds pretty quickly (we have ~1.000 test accounts to fund across all test runs); returning the funds after the test is executed is problematic due to previous point.

I think the most prudent thing to do is to drop switching to devnode, especially since we'll be updating most of our e2e tests to motsu tests.

PR #627, opened 2025-04-10
feat: add Pedersen hash

### LINKED ISSUE #269: [Feature]: Pedersen Hash

Pedersen Hash for Stylus

### Contribution Guidelines

bidzyyys:
Resolves #269

#### PR Checklist

- [x] Tests
- [x] Documentation
- [ ] Changelog


lib/crypto/src/pedersen/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+                U256::ZERO <= element && element < F::FIELD_PRIME,
+                "Pedersen hash failed -- invalid input"
+            );
+
+            let start_idx = 2 + idx * F::N_ELEMENT_BITS_HASH;
+            let end_idx = 2 + (idx + 1) * F::N_ELEMENT_BITS_HASH;
+
+            if end_idx > F::CONSTANT_POINTS.len() {
+                panic!("Pedersen hash failed -- too many elements");
+            }
```
qalisander, line 95: I like clippy recomendation:
```suggestion
            assert!(end_idx <= F::CONSTANT_POINTS.len(), "Pedersen hash failed -- too many elements");
```

lib/crypto/src/pedersen/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+        let hash = self.finalize();
+        hash.x().expect("Pedersen hash failed")
+    }
+}
+
+impl<F: PedersenParams<P>, P: SWCurveConfig> Hasher for Pedersen<F, P> {
+    type Output = Affine<P>;
+
+    /// Update the hash state with a new element.
+    fn update(&mut self, input: impl AsRef<[u8]>) {
```
qalisander, line 66: I don't think it is a good API for this kind of hash. Since after update is called you recreate `U256` from variable size of bytes.
And e.g. calling `update()` two times with inputs `[42, 42]` and `[43,43]`. Or calling it with inputs `[42, 42, 0, 0]` and `[43, 43]` should produce the same hash, since bytes are not padded. And `[42, 42]` and `[42, 42, 0, 0]` will be parsed to the same integer. So we can easily peek different set of bytes, that will produce the same hash. It is an actual vulnerability to preimage attack. I don't think we should implement `Hasher` in this case (more simple) or redesign the trait and make it more generic.

I see that starknet-rs has only [api](https://github.com/xJonathanLEI/starknet-rs/blob/c1e636bfab2d46a2ce484c8313547950a27058e6/starknet-crypto/src/pedersen_hash/mod.rs#L27) where you can pass `256-bit` integer.

lib/crypto/src/pedersen/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+impl<F: PedersenParams<P>, P: SWCurveConfig> Default for Pedersen<F, P> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl<F: PedersenParams<P>, P: SWCurveConfig> Pedersen<F, P> {
+    #[must_use]
+    #[inline]
+
```
qalisander, line 34: Empty line
```suggestion
    #[inline]
```

lib/crypto/src/pedersen/instance/starknet.rs
```diff
@@ -0,0 +1,2216 @@
...
+        Affine::new_unchecked(G_GENERATOR_X, G_GENERATOR_Y);
+}
+#[derive(Clone, Default, PartialEq, Eq)]
+/// Pedersen Hash parameters for Starknet.
+pub struct StarknetPedersenParams;
+
+// TODO: confirm this
+impl PedersenParams<StarknetCurveConfig> for StarknetPedersenParams {
+    #[allow(long_running_const_eval)]
+    const CONSTANT_POINTS: &'static [Affine<StarknetCurveConfig>] = &[
```
qalisander, line 69: Clippy is angry on us not using `vec!` for that large local array. May be disable this lint?

qalisander (commented):
Good job!
First pass

lib/crypto/src/pedersen/mod.rs
bidzyyys: Fair point, I removed `Hasher` trait implementation.

lib/crypto/src/pedersen/mod.rs
```diff
@@ -0,0 +1,115 @@
...
+            assert!(point_list.len() == F::N_ELEMENT_BITS_HASH);
+
+            for pt in point_list {
+                assert!(pt.x != point.x, "Unhashable input.");
+                if element.ct_is_odd() {
+                    point += pt;
+                }
+                element.div2_assign();
+            }
+            assert!(element.is_zero());
```
0xNeshi, line 109: Missing error message.

Also, I'm asking as a layperson, is this so unlikely to fail to warrant a panic? If we expect this to fail occasionally, is there no way to assert such a thing before the loop?

0xNeshi, line 109: Some comments explaining in short what's going on in each part of the function and why certain things are done would be good for maintainability.

lib/crypto/src/pedersen/params.rs
```diff
@@ -0,0 +1,17 @@
+//! Pedersen hash parameters.
+
+use crate::{
+    arithmetic::uint::U256,
+    curve::sw::{Affine, SWCurveConfig},
+};
+/// Pedersen hash parameters.
```
0xNeshi, line 7:
```suggestion
};

/// Pedersen hash parameters.
```

lib/crypto/src/pedersen/mod.rs
```diff
@@ -0,0 +1,115 @@
...
+        AffineRepr,
+    },
+    pedersen::params::PedersenParams,
+};
+
+/// Pedersen hash.
+#[derive(Clone, Debug)]
+pub struct Pedersen<F: PedersenParams<P>, P: SWCurveConfig> {
+    params: core::marker::PhantomData<F>,
+    curve: core::marker::PhantomData<P>,
```
0xNeshi, line 20:
```suggestion
    params: PhantomData<F>,
    curve: PhantomData<P>,
```
nit: import `PhantomData` directly instead of using absolute path repeatedly

lib/crypto/src/pedersen/mod.rs
```diff
@@ -0,0 +1,115 @@
+//! This module contains Pedersen Hash Function implementation.
+
+pub mod instance;
+pub mod params;
+use alloc::{vec, vec::Vec};
```
0xNeshi, line 5: nit: place `pub mod`s below imports for clarity, I almost didn't realize these modules were declared here (confused them with additional imports)

lib/crypto/src/pedersen/instance/starknet.rs
```diff
@@ -0,0 +1,2231 @@
...
+			fp_from_num!("1713931329540660377023406109199410414810705867260802078187082345529207694986")
+		);
+
+    fn constant_points() -> &'static [Affine<StarknetCurveConfig>] {
+        CONSTANT_POINTS.as_ref()
+    }
+}
+
+/// Constant points for Starknet Curve.
+static CONSTANT_POINTS: Lazy<Vec<Affine<StarknetCurveConfig>>> = Lazy::new(
```
0xNeshi, line 84: A link to where these constant points were taken from would be useful, helps maintainability and review

lib/crypto/src/pedersen/instance/starknet.rs
```diff
@@ -0,0 +1,2231 @@
...
+/// Base Field for [`StarknetCurveConfig`].
+pub type Fq = Fp256<FqParam>;
+/// Base Field parameters for [`StarknetCurveConfig`].
+pub struct FqParam;
+
+impl FpParams<LIMBS_256> for FqParam {
+    // The multiplicative generator of Fp.
+    const GENERATOR: Fp256<Self> = fp_from_num!("3");
+    // Starknet's base field modulus.
+    const MODULUS: U256 = from_num!("3618502788666131213697322783095070105623107215331596699973092056135872020481");
```
0xNeshi, line 30: Link to source of this const, if not here then at the module-level docs `//!`

lib/crypto/src/pedersen/mod.rs
```diff
@@ -0,0 +1,115 @@
...
+                end_idx <= constant_points_len,
+                "Pedersen hash failed -- too many elements"
+            );
+
+            let point_list = &constant_points[start_idx..end_idx];
+
+            assert!(point_list.len() == F::N_ELEMENT_BITS_HASH);
+
+            for pt in point_list {
+                assert!(pt.x != point.x, "Unhashable input.");
```
0xNeshi, line 103: We should `assert!` things that must always hold for our code to be valid, the "invariants" of our code. 
Otherwise returning a `Result::Err` might be clearer and more in line with Rust's way of doing things.

Not saying this or any other assert is either of the two as I lack in-depth knowledge, just pointing it out so that you can determine the right approach and that we're on the same page.

0xNeshi (approved):

0xNeshi (commented):
Wrong approval

lib/crypto/src/pedersen/instance/starknet.rs
0xNeshi: Would it be more efficient to apply the clippy recommendation?

bidzyyys: @0xNeshi it is not the case, right now it uses `Lazy<>`.

bidzyyys:
Solved in #644.

PR #633, opened 2025-04-22
feat: twisted edwards curves

qalisander:
Follows #589

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog

### DIFF
--- a/lib/crypto/src/curve/te/affine.rs
+++ b/lib/crypto/src/curve/te/affine.rs
@@ -1,0 +1,248 @@
+//! Affine coordinates for a point on a Twisted Edwards curve ([Affine Space]).
+//!
+//! [Affine Space]: https://en.wikipedia.org/wiki/Affine_space
+use core::{
+    borrow::Borrow,
+    fmt::{Debug, Display, Formatter},
+    ops::{Add, Mul, Neg, Sub},
+};
+
+use educe::Educe;
+use num_traits::{One, Zero};
+use zeroize::Zeroize;
+
+use super::{Projective, TECurveConfig};
+use crate::{
+    bits::BitIteratorBE,
+    curve::AffineRepr,
+    field::{group::AdditiveGroup, prime::PrimeField, Field},
+};
+
+/// Affine coordinates for a point on a twisted Edwards curve, over the
+/// base field `P::BaseField`.
+#[derive(Educe)]
+#[educe(Copy, Clone, PartialEq, Eq, Hash)]
+#[must_use]
+pub struct Affine<P: TECurveConfig> {
+    /// X coordinate of the point represented as a field element
+    pub x: P::BaseField,
+    /// Y coordinate of the point represented as a field element
+    pub y: P::BaseField,
+}
+
+impl<P: TECurveConfig> Display for Affine<P> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
+        if self.is_zero() {
+            write!(f, "infinity")
+        } else {
+            write!(f, "({}, {})", self.x, self.y)
+        }
+    }
+}
+
+impl<P: TECurveConfig> Debug for Affine<P> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
+        if self.is_zero() {
+            write!(f, "infinity")
+        } else {
+            write!(f, "({}, {})", self.x, self.y)
+        }
+    }
+}
+
+impl<P: TECurveConfig> PartialEq<Projective<P>> for Affine<P> {
+    fn eq(&self, other: &Projective<P>) -> bool {
+        self.into_group() == *other
+    }
+}
+
+impl<P: TECurveConfig> Affine<P> {
+    /// Construct a new group element without checking whether the coordinates
+    /// specify a point in the subgroup.
+    pub const fn new_unchecked(x: P::BaseField, y: P::BaseField) -> Self {
+        Self { x, y }
+    }
+
+    /// Construct a new group element in a way while enforcing that points are
+    /// in the prime-order subgroup.
+    ///
+    /// # Panics
+    ///
+    /// * If point is not on curve.
+    /// * If point is not in the prime-order subgroup.
+    pub fn new(x: P::BaseField, y: P::BaseField) -> Self {
+        let p = Self::new_unchecked(x, y);
+        assert!(p.is_on_curve());
+        assert!(p.is_in_correct_subgroup_assuming_on_curve());
+        p
+    }
+
+    /// Construct the identity of the group
+    pub const fn zero() -> Self {
+        Self::new_unchecked(P::BaseField::ZERO, P::BaseField::ONE)
+    }
+
+    /// Is this point the identity?
+    pub fn is_zero(&self) -> bool {
+        self.x.is_zero() && self.y.is_one()
+    }
+
+    /// Checks that the current point is on the elliptic curve.
+    pub fn is_on_curve(&self) -> bool {
+        let x2 = self.x.square();
+        let y2 = self.y.square();
+
+        let lhs = y2 + P::mul_by_a(x2);
+        let rhs = P::BaseField::one() + (P::COEFF_D * (x2 * y2));
+
+        lhs == rhs
+    }
+}
+
+impl<P: TECurveConfig> Affine<P> {
+    /// Checks if `self` is in the subgroup having order equaling that of
+    /// `P::ScalarField` given it is on the curve.
+    pub fn is_in_correct_subgroup_assuming_on_curve(&self) -> bool {
+        P::is_in_correct_subgroup_assuming_on_curve(self)
+    }
+}
+
+impl<P: TECurveConfig> AffineRepr for Affine<P> {
+    type BaseField = P::BaseField;
+    type Config = P;
+    type Group = Projective<P>;
+    type ScalarField = P::ScalarField;
+
+    fn xy(&self) -> Option<(Self::BaseField, Self::BaseField)> {
+        (!self.is_zero()).then_some((self.x, self.y))
+    }
+
+    fn generator() -> Self {
+        P::GENERATOR
+    }
+
+    fn zero() -> Self {
+        Self::new_unchecked(P::BaseField::ZERO, P::BaseField::ONE)
+    }
+
+    fn mul_bigint(&self, by: impl BitIteratorBE) -> Self::Group {
+        P::mul_affine(self, by)
+    }
+
+    /// Multiplies this element by the cofactor and output the
+    /// resulting projective element.
+    fn mul_by_cofactor_to_group(&self) -> Self::Group {
+        P::mul_affine(self, Self::Config::COFACTOR)
+    }
+
+    /// Performs cofactor clearing.
+    /// The default method is simply to multiply by the cofactor.
+    /// Some curves can implement a more efficient algorithm.
+    fn clear_cofactor(&self) -> Self {
+        P::clear_cofactor(self)
+    }
+}
+
+impl<P: TECurveConfig> Zeroize for Affine<P> {
+    // The phantom data does not contain element-specific data
+    // and thus does not need to be zeroized.
+    fn zeroize(&mut self) {
+        self.x.zeroize();
+        self.y.zeroize();
+    }
+}
+
+impl<P: TECurveConfig> Neg for Affine<P> {
+    type Output = Self;
+
+    fn neg(self) -> Self {
+        Self::new_unchecked(-self.x, self.y)
+    }
+}
+
+impl<P: TECurveConfig, T: Borrow<Self>> Add<T> for Affine<P> {
+    type Output = Projective<P>;
+
+    fn add(self, other: T) -> Self::Output {
+        let mut copy = self.into_group();
+        copy += other.borrow();
+        copy
+    }
+}
+
+impl<P: TECurveConfig> Add<Projective<P>> for Affine<P> {
+    type Output = Projective<P>;
+
+    fn add(self, other: Projective<P>) -> Projective<P> {
+        other + self
+    }
+}
+
+impl<'a, P: TECurveConfig> Add<&'a Projective<P>> for Affine<P> {
+    type Output = Projective<P>;
+
+    fn add(self, other: &'a Projective<P>) -> Projective<P> {
+        *other + self
+    }
+}
+
+impl<P: TECurveConfig, T: Borrow<Self>> Sub<T> for Affine<P> {
+    type Output = Projective<P>;
+
+    fn sub(self, other: T) -> Self::Output {
+        let mut copy = self.into_group();
+        copy -= other.borrow();
+        copy
+    }
+}
+
+impl<P: TECurveConfig> Sub<Projective<P>> for Affine<P> {
+    type Output = Projective<P>;
+
+    fn sub(self, other: Projective<P>) -> Projective<P> {
+        self + (-other)
+    }
+}
+
+impl<'a, P: TECurveConfig> Sub<&'a Projective<P>> for Affine<P> {
+    type Output = Projective<P>;
+
+    fn sub(self, other: &'a Projective<P>) -> Projective<P> {
+        self + (-*other)
+    }
+}
+
+impl<P: TECurveConfig> Default for Affine<P> {
+    #[inline]
+    fn default() -> Self {
+        Self::zero()
+    }
+}
+
+impl<P: TECurveConfig, T: Borrow<P::ScalarField>> Mul<T> for Affine<P> {
+    type Output = Projective<P>;
+
+    #[inline]
+    fn mul(self, other: T) -> Self::Output {
+        self.mul_bigint(other.borrow().into_bigint())
+    }
+}
+
+// The projective point X, Y, T, Z is represented in the affine
+// coordinates as X/Z, Y/Z.
+impl<P: TECurveConfig> From<Projective<P>> for Affine<P> {
+    fn from(p: Projective<P>) -> Affine<P> {
+        if p.is_zero() {
+            Affine::zero()
+        } else if p.z.is_one() {
+            // If Z is one, the point is already normalized.
+            Affine::new_unchecked(p.x, p.y)
+        } else {
+            // Z is nonzero, so it must have inverse in a field.
+            let z_inv = p.z.inverse().unwrap();
+            let x = p.x * z_inv;
+            let y = p.y * z_inv;
+            Affine::new_unchecked(x, y)
+        }
+    }
+}
--- a/lib/crypto/src/curve/te/mod.rs
+++ b/lib/crypto/src/curve/te/mod.rs
@@ -1,0 +1,111 @@
+//! This module contains definitions for the [Twisted Edwards model] of the
+//! curve.
+//!
+//! [Twisted Edwards model]: https://www.hyperelliptic.org/EFD/g1p/auto-twisted.html
+use num_traits::Zero;
+
+mod affine;
+pub use affine::*;
+
+mod projective;
+pub use projective::*;
+
+use crate::{
+    bits::BitIteratorBE,
+    curve::AffineRepr,
+    field::{group::AdditiveGroup, prime::PrimeField},
+};
+
+/// Constants and convenience functions
+/// that define the [Twisted Edwards model] of the curve.
+///
+/// In this model, the curve equation is `a * x² + y² = 1 + d * x² * y²`, for
+/// constants `a` and `d`.
+///
+/// [Twisted Edwards model]: https://www.hyperelliptic.org/EFD/g1p/auto-twisted.html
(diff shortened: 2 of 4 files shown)


lib/crypto/src/curve/te/affine.rs
```diff
@@ -0,0 +1,248 @@
...
+            write!(f, "infinity")
+        } else {
+            write!(f, "({}, {})", self.x, self.y)
+        }
+    }
+}
+
+impl<P: TECurveConfig> PartialEq<Projective<P>> for Affine<P> {
+    fn eq(&self, other: &Projective<P>) -> bool {
+        self.into_group() == *other
```
0xNeshi:
```suggestion
        &self.into_group() == other
```
avoids dereferencing (temp. value creation)

lib/crypto/src/curve/te/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+    #[inline(always)]
+    fn mul_by_a(elem: Self::BaseField) -> Self::BaseField {
+        elem * Self::COEFF_A
+    }
+
+    /// Checks that the current point is in the prime order subgroup given
+    /// the point on the curve.
+    fn is_in_correct_subgroup_assuming_on_curve(item: &Affine<Self>) -> bool {
+        Self::mul_affine(item, Self::ScalarField::characteristic()).is_zero()
+    }
```
0xNeshi:
```suggestion
    /// Checks that the current point is in the prime order subgroup given
    /// the point on the curve.
    fn is_in_prime_order_subgroup(item: &Affine<Self>) -> bool {
        Self::mul_affine(item, Self::ScalarField::characteristic()).is_zero()
    }
```
Judging by docs, it seems that this is what it's checking

lib/crypto/src/curve/te/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+    /// the product can be computed faster than standard field multiplication
+    /// (eg: via doubling if `COEFF_A == 2`, or if `COEFF_A.is_zero()`).
+    #[inline(always)]
+    fn mul_by_a(elem: Self::BaseField) -> Self::BaseField {
+        elem * Self::COEFF_A
+    }
+
+    /// Checks that the current point is in the prime order subgroup given
+    /// the point on the curve.
+    fn is_in_correct_subgroup_assuming_on_curve(item: &Affine<Self>) -> bool {
```
0xNeshi:
```suggestion
    /// Checks that the current point is in the prime order subgroup, assuming
    /// the point is already on the curve.
    fn is_in_correct_subgroup_assuming_on_curve(item: &Affine<Self>) -> bool {
```

lib/crypto/src/curve/te/affine.rs
```diff
@@ -0,0 +1,248 @@
...
+        lhs == rhs
+    }
+}
+
+impl<P: TECurveConfig> Affine<P> {
+    /// Checks if `self` is in the subgroup having order equaling that of
+    /// `P::ScalarField` given it is on the curve.
+    pub fn is_in_correct_subgroup_assuming_on_curve(&self) -> bool {
+        P::is_in_correct_subgroup_assuming_on_curve(self)
+    }
```
0xNeshi:
```suggestion
    /// Checks if this point is in the prime-order subgroup.
    /// 
    /// This assumes the point is already on the curve and verifies it belongs
    /// to the subgroup with order equal to `P::ScalarField`.
    pub fn is_in_prime_order_subgroup(&self) -> bool {
        P::is_in_correct_subgroup_assuming_on_curve(self)
    }
```
Is this a bit clearer?

lib/crypto/src/curve/te/affine.rs
```diff
@@ -0,0 +1,248 @@
...
+
+impl<P: TECurveConfig> AffineRepr for Affine<P> {
+    type BaseField = P::BaseField;
+    type Config = P;
+    type Group = Projective<P>;
+    type ScalarField = P::ScalarField;
+
+    fn xy(&self) -> Option<(Self::BaseField, Self::BaseField)> {
+        (!self.is_zero()).then_some((self.x, self.y))
+    }
```
0xNeshi, line 120: Why do you think it's better to return `None` for zero than to return actual zero tuple (`(P::BaseField::ZERO, P::BaseField::ONE)`)?

lib/crypto/src/curve/te/affine.rs
```diff
@@ -0,0 +1,248 @@
...
+    /// Performs cofactor clearing.
+    /// The default method is simply to multiply by the cofactor.
+    /// Some curves can implement a more efficient algorithm.
+    fn clear_cofactor(&self) -> Self {
+        P::clear_cofactor(self)
+    }
+}
+
+impl<P: TECurveConfig> Zeroize for Affine<P> {
+    // The phantom data does not contain element-specific data
```
0xNeshi: Which "phantom data" is this referring to?

lib/crypto/src/curve/te/affine.rs
```diff
@@ -0,0 +1,248 @@
...
+    // and thus does not need to be zeroized.
+    fn zeroize(&mut self) {
+        self.x.zeroize();
+        self.y.zeroize();
+    }
```
0xNeshi, line 152: If [`Affine::<P>::zero()`](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/633/files#diff-54bb3fa19516067745d9122a6c161cd60718e8a01672a2bd8b1830dfd51c9f62R124) returns:
```rust
        Self::new_unchecked(P::BaseField::ZERO, P::BaseField::ONE) // (x: 0, y: 1)
```
Shouldn't this be something like:
```rust
fn zeroize(&mut self) {
    self.x.zeroize();
    self.y = P::BaseField::ONE;
}
```
?

lib/crypto/src/curve/te/affine.rs
```diff
@@ -0,0 +1,248 @@
...
+
+impl<P: TECurveConfig> Add<Projective<P>> for Affine<P> {
+    type Output = Projective<P>;
+
+    fn add(self, other: Projective<P>) -> Projective<P> {
+        other + self
+    }
+}
+
+impl<'a, P: TECurveConfig> Add<&'a Projective<P>> for Affine<P> {
```
0xNeshi, line 181: Current `Add` and `Sub` impls allow:
1. owned <op> owned
2. owned <op> reference

Would be useful to have impls for:
1. reference <op> reference

lib/crypto/src/curve/te/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+
+        for (result, (expected_x, expected_y)) in result.iter().zip(expected) {
+            assert!(result.is_on_curve());
+            assert_eq!(result.x, expected_x);
+            assert_eq!(result.y, expected_y);
+        }
+    }
+
+    #[test]
+    fn point_add() {
```
0xNeshi, line 230: missing tests for:
- owned + ref
- ref + ref

lib/crypto/src/curve/te/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+        assert_eq!(g2_affine, expected_g2);
+
+        // Test G + (-G) = 0
+        let neg_g = -g_proj;
+        let zero = g_proj + neg_g;
+        assert!(zero.is_zero());
+    }
+
+    #[test]
+    fn point_sub() {
```
0xNeshi, line 252: missing tests for:
- owned - ref
- ref - ref

lib/crypto/src/curve/te/projective.rs
```diff
@@ -0,0 +1,412 @@
...
+    pub y: P::BaseField,
+    /// The t-coordinate of the point in projective coordinates.
+    pub t: P::BaseField,
+    /// The z-coordinate of the point in projective coordinates.
+    pub z: P::BaseField,
+}
+
+impl<P: TECurveConfig> PartialEq<Affine<P>> for Projective<P> {
+    fn eq(&self, other: &Affine<P>) -> bool {
+        *self == other.into_group()
```
0xNeshi:
```suggestion
        self == &other.into_group()
```
avoids deref

lib/crypto/src/curve/te/projective.rs
```diff
@@ -0,0 +1,412 @@
...
+    }
+}
+
+impl<P: TECurveConfig> Display for Projective<P> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
+        write!(f, "{}", Affine::from(*self))
```
0xNeshi, line 52: Could be useful to have :
```rust
impl<'a, P: TECurveConfig> From<&'a Projective<P>> for Affine<P> {
```

so this becomes 
```rust
        write!(f, "{}", Affine::from(self))
```

lib/crypto/src/curve/te/projective.rs
```diff
@@ -0,0 +1,412 @@
...
+            return other.is_zero();
+        }
+
+        if other.is_zero() {
+            return false;
+        }
+
+        // x1/z1 == x2/z2  <==> x1 * z2 == x2 * z1
+        (self.x * other.z) == (other.x * self.z)
+            && (self.y * other.z) == (other.y * self.z)
```
0xNeshi: in comment: missing Y-part of the equation

is the comment even necessary?

lib/crypto/src/curve/te/projective.rs
```diff
@@ -0,0 +1,412 @@
...
+}
+impl<P: TECurveConfig> Zeroize for Projective<P> {
+    // The phantom data does not contain element-specific data
+    // and thus does not need to be zeroized.
+    fn zeroize(&mut self) {
+        self.x.zeroize();
+        self.y.zeroize();
+        self.t.zeroize();
+        self.z.zeroize();
+    }
```
0xNeshi, line 123: same as before, shouldn't `y` and `z` be `P::BaseField::one()`?

lib/crypto/src/curve/te/projective.rs
```diff
@@ -0,0 +1,412 @@
...
+            P::BaseField::one(),
+        )
+    }
+
+    fn is_zero(&self) -> bool {
+        self.x.is_zero()
+            && self.y == self.z
+            && !self.y.is_zero()
+            && self.t.is_zero()
+    }
```
0xNeshi, line 136: so even if `self.y` is not 1, but e.g. 2, this will still hold as long as `y == z`?

lib/crypto/src/curve/te/projective.rs
```diff
@@ -0,0 +1,412 @@
...
+
+impl<P: TECurveConfig> Zero for Projective<P> {
+    fn zero() -> Self {
+        Self::new_unchecked(
+            P::BaseField::zero(),
+            P::BaseField::one(),
+            P::BaseField::zero(),
+            P::BaseField::one(),
+        )
+    }
```
0xNeshi: Can't we just return [`Projective::<P>::ZERO`](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/633/files#diff-eedd6cf1ffa6385ac5599d98d82b5af798889ea6480576e10ba50d15a8589040R148) here?

lib/crypto/src/curve/te/projective.rs
```diff
@@ -0,0 +1,412 @@
...
+        // H = D - B
+        let h = d - b;
+        // X3 = E * F
+        self.x = e * f;
+        // Y3 = G * H
+        self.y = g * h;
+        // T3 = E * H
+        self.t = e * h;
+        // Z3 = F * G
+        self.z = f * g;
```
0xNeshi, line 177: nit: are the comments necessary?

lib/crypto/src/curve/te/projective.rs
```diff
@@ -0,0 +1,412 @@
...
+        P::BaseField::ONE,
+        P::BaseField::ZERO,
+        P::BaseField::ONE,
+    );
+
+    fn double_in_place(&mut self) -> &mut Self {
+        // See "Twisted Edwards Curves Revisited"
+        // Huseyin Hisil, Kenneth Koon-Ho Wong, Gary Carter, and Ed Dawson
+        // 3.3 Doubling in E^e
+        // Source: https://www.hyperelliptic.org/EFD/g1p/data/twisted/extended/doubling/dbl-2008-hwcd
```
0xNeshi: Better to put this comment above as fn docs

lib/crypto/src/curve/te/projective.rs
```diff
@@ -0,0 +1,412 @@
...
+        // H = B-a*A
+        let h = b - P::mul_by_a(a);
+        // X3 = E*F
+        self.x = e * f;
+        // Y3 = G*H
+        self.y = g * h;
+        // T3 = E*H
+        self.t = e * h;
+        // Z3 = F*G
+        self.z = f * g;
```
0xNeshi: Same question as before for comments

lib/crypto/src/curve/te/projective.rs
```diff
@@ -0,0 +1,412 @@
...
+        // See "Twisted Edwards Curves Revisited" (https://eprint.iacr.org/2008/522.pdf)
+        // by Huseyin Hisil, Kenneth Koon-Ho Wong, Gary Carter, and Ed Dawson
+        // 3.1 Unified Addition in E^e
+
+        // A = x1 * x2
+        let a = self.x * other.x;
+
+        // B = y1 * y2
+        let b = self.y * other.y;
+
```
0xNeshi: Same question as before for comments

lib/crypto/src/curve/te/projective.rs
```diff
@@ -0,0 +1,412 @@
...
+    fn sub(mut self, other: T) -> Self {
+        self -= other.borrow();
+        self
+    }
+}
+
+impl_additive_ops_from_ref!(Projective, TECurveConfig);
+
+impl<'a, P: TECurveConfig> Add<&'a Self> for Projective<P> {
```
0xNeshi, line 301: Would it make sense to implement `ref + ref`?

lib/crypto/src/curve/te/affine.rs
0xNeshi: Could these be implemented with `impl_additive_ops_from_ref!`?

lib/crypto/src/curve/te/affine.rs
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    /// Construct a new group element in a way while enforcing that points are
+    /// in the prime-order subgroup.
+    ///
+    /// # Panics
+    ///
+    /// * If point is not on curve.
+    /// * If point is not in the prime-order subgroup.
+    pub fn new(x: P::BaseField, y: P::BaseField) -> Self {
```
0xNeshi, line 73: Some general comments:
- could any of the new functions be marked with `inline(always)` or `must_use`?
- some flows are not covered with unit tests
- could we implement any proptests for this?
- missing CHANGELOG

bidzyyys: Agree with @0xNeshi 💯

lib/crypto/src/curve/te/affine.rs
```diff
@@ -0,0 +1,248 @@
...
+        let p = Self::new_unchecked(x, y);
+        assert!(p.is_on_curve());
+        assert!(p.is_in_correct_subgroup_assuming_on_curve());
+        p
+    }
+
+    /// Construct the identity of the group
```
bidzyyys:
```suggestion
    /// Construct the identity of the group.
```

lib/crypto/src/curve/te/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+    /// Performs cofactor clearing.
+    /// The default method is simply to multiply by the cofactor.
+    /// For some curve families though, it is sufficient to multiply
+    /// by a smaller scalar.
+    fn clear_cofactor(item: &Affine<Self>) -> Affine<Self> {
+        item.mul_by_cofactor()
+    }
+
+    /// Default implementation of group multiplication for projective
+    /// coordinates
```
bidzyyys:
```suggestion
    /// coordinates.
```

lib/crypto/src/curve/te/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+            if b {
+                res += base;
+            }
+        }
+
+        res
+    }
+
+    /// Default implementation of group multiplication for affine
+    /// coordinates
```
bidzyyys:
```suggestion
    /// coordinates.
```

lib/crypto/src/curve/te/projective.rs
```diff
@@ -0,0 +1,412 @@
...
+        // to its affine representation, by the conversion
+        // (x, y, t, z) -> (x/z, y/z, t/z, 1)
+        // Batch normalizing N twisted edwards curve elements costs:
+        //     1 inversion + 6N field multiplications
+        // (batch inversion requires 3N multiplications + 1 inversion)
+        let mut z_s = v.iter().map(|g| g.z).collect::<Vec<_>>();
+
+        batch_inversion(&mut z_s);
+
+        // Perform affine transformations
```
bidzyyys:
```suggestion
        // Perform affine transformations.
```

lib/crypto/src/curve/te/projective.rs
```diff
@@ -0,0 +1,412 @@
...
+impl<P: TECurveConfig> CurveGroup for Projective<P> {
+    type Affine = Affine<P>;
+    type BaseField = P::BaseField;
+    type Config = P;
+    type FullGroup = Affine<P>;
+
+    fn normalize_batch(v: &[Self]) -> Vec<Self::Affine> {
+        // A projective curve element (x, y, t, z) is normalized
+        // to its affine representation, by the conversion
+        // (x, y, t, z) -> (x/z, y/z, t/z, 1)
```
bidzyyys:
```suggestion
        // (x, y, t, z) -> (x/z, y/z, t/z, 1).
```

lib/crypto/src/curve/te/projective.rs
```diff
@@ -0,0 +1,412 @@
...
+        // Batch normalizing N twisted edwards curve elements costs:
+        //     1 inversion + 6N field multiplications
+        // (batch inversion requires 3N multiplications + 1 inversion)
```
bidzyyys:
```suggestion
        // (batch inversion requires 3N multiplications + 1 inversion).
```

bidzyyys (commented):
Almost looks good, take care of @0xNeshi comments please.

lib/crypto/src/curve/te/affine.rs
qalisander: Yeah. Removed it. Consequences of refactor

lib/crypto/src/curve/te/affine.rs
qalisander: This trait is mostly about erasing object from memory, than setting the correct "identity element" (zero). In this case it is (0, 1))
image

lib/crypto/src/curve/te/projective.rs
qalisander: It serves as an explanation of comparison. Added another clarification comment

lib/crypto/src/curve/te/projective.rs
qalisander: I don't mind of leaving them

lib/crypto/src/curve/te/projective.rs
qalisander: Same explained [here](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/633#discussion_r2207815843)

lib/crypto/src/curve/te/projective.rs
0xNeshi: To me the code is clear enough without the comment, but we can leave it just in case

lib/crypto/src/curve/te/projective.rs
0xNeshi: It shouldn't, see https://github.com/OpenZeppelin/rust-contracts-stylus/pull/633/files#r2207815843

lib/crypto/src/curve/te/affine.rs
qalisander: I think the problem here is in the similar api with sw curves.. We implement the same trait. There it kinda make sense to return None

lib/crypto/src/curve/te/projective.rs
qalisander: Yeah, since those are protective coordinates. You can check also equality is a bit different from what are we expecting

lib/crypto/src/curve/te/affine.rs
qalisander: Actually sw also missing some flows and proptests. Currently tests are enough just to validate implementation. But we will add much more proptests in future.
Added notion in the CHANGELOG

0xNeshi (approved):

bidzyyys (approved):

PR #647, opened 2025-05-05
ref: remove associated error types

bidzyyys:


contracts/src/lib.rs
```diff
@@ -38,11 +38,7 @@ impl MyContract { }
 ```
 */
 
-#![allow(
-    clippy::module_name_repetitions,
-    clippy::used_underscore_items,
-    deprecated
-)]
+#![allow(clippy::module_name_repetitions, deprecated)]
```
0xNeshi, line 41: Fix clippy warnings https://github.com/OpenZeppelin/rust-contracts-stylus/pull/647/checks?check_run_id=41635499044

0xNeshi (commented):
I'm looking into the errors https://github.com/OpenZeppelin/rust-contracts-stylus/actions/runs/14832140078/job/41635428689?pr=647

contracts/src/lib.rs
bidzyyys: Reverted.

0xNeshi (commented):
Removing associated error types will introduce new problems when combined with the [new inheritance model](https://github.com/OffchainLabs/stylus-sdk-rs/pull/223/files#diff-27de61b1f56cbee715e5bb14c6b92f3854701a7a3d7b324fb1e109cec6bc762e) - for each contract our example(s) inherit, we will need to implement a corresponding trait (interface) - this means we won't be able to return custom errors in those implementations. 

To understand what I mean, imagine we merge this PR and are then creating an `Erc20 + Pausable` contract example. We would "override" `transfer` and `transfer_from` to be callable only when not paused, which means they should be returning both `erc20::Error` and `pausable::Error`.
Below is the would-be full code (imagine we have `IPausable` implemented:
```rust
#![cfg_attr(not(test), no_main)]
extern crate alloc;

use alloc::vec::Vec;

use alloy_primitives::{Address, U256};
use openzeppelin_stylus::{
    token::erc20::{self, Erc20, IErc20},
    utils::{pausable, IPausable, Pausable},
};
use stylus_sdk::prelude::*;

#[derive(SolidityError, Debug)]
enum Error {
    InsufficientBalance(erc20::ERC20InsufficientBalance),
    InvalidSender(erc20::ERC20InvalidSender),
    InvalidReceiver(erc20::ERC20InvalidReceiver),
    InsufficientAllowance(erc20::ERC20InsufficientAllowance),
    InvalidSpender(erc20::ERC20InvalidSpender),
    InvalidApprover(erc20::ERC20InvalidApprover),
    EnforcedPause(pausable::EnforcedPause),
    ExpectedPause(pausable::ExpectedPause),
}

impl From<erc20::Error> for Error {
    fn from(value: erc20::Error) -> Self {
        match value {
            erc20::Error::InsufficientBalance(e) => {
                Error::InsufficientBalance(e)
            }
            erc20::Error::InvalidSender(e) => Error::InvalidSender(e),
            erc20::Error::InvalidReceiver(e) => Error::InvalidReceiver(e),
            erc20::Error::InsufficientAllowance(e) => {
                Error::InsufficientAllowance(e)
            }
            erc20::Error::InvalidSpender(e) => Error::InvalidSpender(e),
            erc20::Error::InvalidApprover(e) => Error::InvalidApprover(e),
        }
    }
}

impl From<pausable::Error> for Error {
    fn from(value: pausable::Error) -> Self {
        match value {
            pausable::Error::EnforcedPause(e) => Error::EnforcedPause(e),
            pausable::Error::ExpectedPause(e) => Error::ExpectedPause(e),
        }
    }
}

#[storage]
#[entrypoint]
struct Erc20Example {
    erc20: Erc20,
    pausable: Pausable,
}

#[public]
#[inherit(IErc20, IPausable)]
impl Erc20Example {}

#[public]
impl IPausable for Erc20Example {
    fn paused(&self) -> bool {
        self.pausable.paused()
    }
}

#[public]
impl IErc20 for Erc20Example {
    fn total_supply(&self) -> U256 {
        self.erc20.total_supply()
    }

    fn balance_of(&self, account: Address) -> U256 {
        self.erc20.balance_of(account)
    }

    fn transfer(
        &mut self,
        to: Address,
        value: U256,
    ) -> Result<bool, erc20::Error> {
        Ok(self.erc20.transfer(to, value)?)
    }

    fn allowance(&self, owner: Address, spender: Address) -> U256 {
        self.erc20.allowance(owner, spender)
    }

    fn approve(
        &mut self,
        spender: Address,
        value: U256,
    ) -> Result<bool, erc20::Error> {  // we cannot return the local `Error`, as `IErc20` trait requires `erc20::Error`
        self.erc20.approve(spender, value)
    }

    fn transfer_from(
        &mut self,
        from: Address,
        to: Address,
        value: U256,
    ) -> Result<bool, erc20::Error> { // we cannot return the local `Error`, as `IErc20` trait requires `erc20::Error`
        self.erc20.transfer_from(from, to, value)
    }
}
```
Here's the problem - `transfer` and `transfer_from` should actually be returning our locally defined `enum Error`, i.e. they must be able to return `pausable::Error` too. But without associated error types, this is not possible to do.

**We must have associated error types.**

The only alternative I see is to omit error enums completely and just return `Vec<u8>` everywhere, which will be extremely dev-unfriendly. But even this is not without its issues - I actually just tried implementing this, and the same errors popped up that [we have now](https://github.com/OpenZeppelin/rust-contracts-stylus/actions/runs/14832140078/job/41635428689?pr=647):
```
no associated item named `__stylus_assert_overrides` found for struct `Erc20` in the current scope
associated item not found in `Erc20`
```

There seem to be still some issues with the SDK's inheritance model.

qalisander:
@0xNeshi that was actually my point when I've introduced associated error types, considering the new inheritance model. I agree that the only alternative is to use `Vec<u8>` and it looks quite a downgrade from the currently existing errors.

0xNeshi:
> @0xNeshi that was actually my point when I've introduced associated error types, considering the new inheritance model. I agree that the only alternative is to use `Vec<u8>` and it looks quite a downgrade from the currently existing errors.

I wasn't aware you've foreseen this problem, kudos! 
Associated error types is the most dev-friendly way to go.

0xNeshi:
@bidzyyys can we close this?

PR #662, opened 2025-05-20
build: remove std feature

qalisander:
#### PR Checklist

- [x] Tests
- [ ] Documentation
- [ ] Changelog

### DIFF
--- a/contracts/Cargo.toml
+++ b/contracts/Cargo.toml
@@ -23,12 +23,8 @@ alloy-primitives = { workspace = true, features = ["arbitrary"] }
 motsu.workspace = true
 
 [features]
-# Enables using the standard library. This is not included in the default
-# features, because this crate is meant to be used in a `no_std` environment.
-# Currently, the std feature is only used for testing purposes.
-std = []
 reentrant = ["stylus-sdk/reentrant"]
-export-abi = ["stylus-sdk/export-abi", "std"]
+export-abi = ["stylus-sdk/export-abi"]
 
 [lib]
 crate-type = ["lib"]
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,7 +11,7 @@ permissions:
   contents: read
 on:
   push:
-    branches: [main, v*]
+    branches: [ main, v* ]
     paths-ignore:
       - "**.md"
       - "**.adoc"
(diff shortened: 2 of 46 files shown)


contracts/Cargo.toml
```diff
@@ -23,12 +23,8 @@ alloy-primitives = { workspace = true, features = ["arbitrary"] }
 motsu.workspace = true
 
 [features]
-# Enables using the standard library. This is not included in the default
-# features, because this crate is meant to be used in a `no_std` environment.
-# Currently, the std feature is only used for testing purposes.
-std = []
```
0xNeshi, line 29: Do we need the [panic_handler section](https://github.com/OpenZeppelin/rust-contracts-stylus/blob/20ffabe9cf3876989aa5ea1e002eaab6cc7485a6/README.md?plain=1#L50) in _README.md_?

0xNeshi, line 29: Do we need [nostd.yml](https://github.com/OpenZeppelin/rust-contracts-stylus/blob/20ffabe9cf3876989aa5ea1e002eaab6cc7485a6/.github/workflows/nostd.yml)?

0xNeshi, line 29: Remove `--skip std` from [check.yml](https://github.com/OpenZeppelin/rust-contracts-stylus/blob/20ffabe9cf3876989aa5ea1e002eaab6cc7485a6/.github/workflows/check.yml#L121)

0xNeshi (approved):
Awesome! Change requests are small, approving in advance

bidzyyys (approved):

contracts/Cargo.toml
bidzyyys: IMHO it is not needed.

qalisander: @bidzyyys what do you think @bidzyyys ? I feel that we don't need anymore

qalisander: It actually assures, that library can compile for WASM (non std target). Seems useful, but there are other workflows that implicitly check that

bidzyyys: Looks like it exists only in README so not needed.

bidzyyys: So maybe it should be part of `check.yml`? Wdyt?

PR #665, opened 2025-05-21
test: create `constructor!` macro to aid type-safety for constructors

0xNeshi:
There may be an even better way by somehow getting exact types from `ContractExample::constructor`, would be awesome if we managed to implement that, but not high priority imo, as this is just for testing.


examples/erc20/tests/erc20.rs
```diff
@@ -15,14 +15,7 @@ const TOKEN_SYMBOL: &str = "TTK";
...
 fn ctr(cap: U256) -> Constructor {
-    Constructor {
-        signature: "constructor(string,string,uint256)".to_string(),
-        args: vec![
-            TOKEN_NAME.to_string(),
-            TOKEN_SYMBOL.to_string(),
-            cap.to_string(),
-        ],
-    }
+    constructor!(TOKEN_NAME.to_string(), TOKEN_SYMBOL.to_string(), cap)
```
qalisander, line 18:
```suggestion
    constructor!(TOKEN_NAME, TOKEN_SYMBOL, cap)
```

qalisander: I think the same is applicable to the other place. We don't need to call `to_string` more than once

qalisander (approved):
Good idea @0xNeshi !

contracts/src/token/erc721/extensions/consecutive.rs
```diff
@@ -315,19 +315,7 @@ impl IErc721 for Erc721Consecutive {
...
-impl Erc721Consecutive {
-    // TODO: remove once function overriding is possible, so `max_batch_size`
-    // can be set that way.
-    /// Constructor.
-    ///
-    /// # Arguments
-    ///
-    /// * `&mut self` - Write access to the contract's state.
-    #[constructor]
-    pub fn constructor(&mut self) {
```
bidzyyys, line 327: Why did you remove it?

bidzyyys (approved):
Good job, approved in advance, make sure it works for benches too!

contracts/src/token/erc721/extensions/consecutive.rs
0xNeshi: Useless. Devs can set `max_batch_size` manually, see [example](https://github.com/OpenZeppelin/rust-contracts-stylus/blob/b83948d3b04c4f5e4f1078667cfb71dd8ad72934/examples/erc721-consecutive/src/lib.rs#L33)

examples/erc20/tests/erc20.rs
bidzyyys: I agree with @qalisander

0xNeshi: We have to, because `AbiType` is not implemented for `&str`

PR #669, opened 2025-05-22
chore: merge v0.2.0-rc.0 

bidzyyys:

PR #670, opened 2025-05-26
chore: Merge v0.2 (with conflicts resolved) into main

0xNeshi:
Original PR https://github.com/OpenZeppelin/rust-contracts-stylus/pull/669

- resolved conflicts with origin/main
- removed _check-wasm.yml_ <- if a contract is not deployable, corresponding e2e tests would fail; we also never check and/or compare changes in actual wasm size, we only care if it's within limits.


0xNeshi:
Closed in favor of https://github.com/OpenZeppelin/rust-contracts-stylus/pull/671

PR #673, opened 2025-05-27
fix(audit): L-04: Absence of Overflow Check in from_str_hex Function

qalisander:
Fixes L-04: Absence of Overflow Check in from_str_hex Function


lib/crypto/src/arithmetic/uint.rs
```diff
@@ -979,11 +985,14 @@ impl<const N: usize> WideUint<N> {
 
 #[cfg(test)]
 mod test {
+    use num_traits::ConstZero;
```
0xNeshi, line 988: CI warnings https://github.com/OpenZeppelin/rust-contracts-stylus/actions/runs/15267242864/job/42935024899?pr=673#step:7:668

PR #674, opened 2025-05-27
fix(audit): L-01: Verification Functions for Custom Hashing in Merkle Trees Assume Commutative Hashing

qalisander:
fix L-01: Verification Functions for Custom Hashing in Merkle Trees Assume Commutative Hashing

### DIFF
--- a/scripts/check-wasm.sh
+++ b/scripts/check-wasm.sh
@@ -26,4 +26,5 @@ cargo build --release --target wasm32-unknown-unknown -Z build-std=std,panic_abo
 
 for CRATE_NAME in $(get_example_crate_names); do
   check_wasm "$CRATE_NAME"
+  sleep 2
 done


lib/crypto/src/merkle.rs
```diff
@@ -150,6 +150,9 @@ where
     /// `proof` and a custom hashing algorithm defined by `builder`. See
     /// [`BuildHasher`] for more information on how to construct a builder.
     ///
+    /// Merkle tree hashing process must be constructed commutatively when using
+    /// custom hashing algorithms.
```
0xNeshi, line 154: Since this is an important point to make, maybe we should wrap this in a [warning](https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html#adding-a-warning-block)?
```rust
/// documentation
///
/// <div class="warning">A big warning!</div>
///
/// more documentation
```

0xNeshi (approved):

PR #681, opened 2025-05-28
fix(audit): L-03: Encoding Ambiguity in ct_num_bits Function

qalisander:
Fixes L-03: Encoding Ambiguity in ct_num_bits Function


lib/crypto/src/arithmetic/uint.rs
```diff
@@ -188,15 +188,21 @@ impl<const N: usize> Uint<N> {
...
 
     /// Return the minimum number of bits needed to encode this number.
+    ///
+    /// One bit is necessary to encode zero.
     #[doc(hidden)]
     #[must_use]
     pub const fn ct_num_bits(&self) -> usize {
+        // One bit is necessary to encode zero.
+        if self.ct_is_zero() {
+            return 1;
```
0xNeshi, line 198: Update the [`#[test] fn num_bits`](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/681/files#diff-463d4484afcd15f2bd182d700a3a87aa0f71df2dc27652948308d0e99f384ec6R1046) to include cases for `0` and `1`.

0xNeshi (approved):
Looks good, approving in advance

lib/crypto/src/arithmetic/uint.rs
bidzyyys: Yeah, you should include these cases @qalisander.

bidzyyys (approved):

lib/crypto/src/arithmetic/uint.rs
qalisander: It is actually already covered by doctest [here](https://github.com/OpenZeppelin/rust-contracts-stylus/blob/affc2ee0e4da35bea16de20d0e8de5a1e9e640a8/lib/crypto/src/arithmetic/mod.rs#L120)

PR #689, opened 2025-06-05
feat: add Precompiles trait

0xNeshi:
Proposal for unifying all precompiles in a single extension trait that can be more ergonomically invoked on contracts.

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add `EnumerableAddressSet`. #622
 - Add Twisted-Edwards Curves. #633
 - Add Elliptic Curve Configurations: secp256k1, Baby Jubjub, Bandersnatch, Curve25519, Jubjub. #738
+- Add `Precompiles` trait for ergonomic precompile invocation. #689
 
 ### Changed
 


contracts/src/utils/precompiles.rs
```diff
@@ -0,0 +1,111 @@
...
+    ///
+    /// Re-exports selected ECDSA types and constants specifically relevant
+    /// to the ecrecover operation.
+    pub mod ecrecover {
+        pub use crate::utils::cryptography::ecdsa::{
+            ECDSAInvalidSignature, ECDSAInvalidSignatureS, EcRecoverData,
+            Error, ECRECOVER_ADDR, SIGNATURE_S_UPPER_BOUND,
+        };
+    }
+}
```
0xNeshi: Didn't remove `utils::cryptography::ecdsa` for two reasons:
1. The original crate can be expanded to include additional ECDSA-related operations, which have nothing to do with ecrecover precompile; the crate would be useful to devs as a standalone.
2. Backward compatibility.

The `mod primitives` module is used to make it clear to devs where to find relevant precompile-related types, constants, enums etc.
In this specific case, the reason for this reexport is that it should be easier for the average Stylus developer to understand that the relevant ecrecover primitives are in the same module, i.e. `precompiles::primitives::ecrecover`, than that they should actually import them from a completely different module `utils::cryptography::ecdsa`.

bidzyyys (commented):
I like the idea, but my main concern is that we are not able to build common interface (`trait`) for all the precompiles. IMHO each precompile is specific, and I do not see a way how we can handle it.

0xNeshi:
> I like the idea, but my main concern is that we are not able to build common interface (`trait`) for all the precompiles. IMHO each precompile is specific, and I do not see a way how we can handle it.

The idea is that each new precompile has its own function within `trait Precompiles`; the dev then imports `utils::precompiles::Precompiles` and they have access to all the wrapper functions.

Can you think of any precompile that would not work with this approach?

contracts/src/utils/precompiles.rs
```diff
@@ -0,0 +1,111 @@
+//! `ArbOS` precompile wrapper enabling easier precompile invocation.
```
bidzyyys:
```suggestion
//! `ArbOS` precompiles wrapper enabling easier invocation.
```

0xNeshi:
> I think I may like the solution but please provide a refactor to our contracts too with the new precompile (Erc20Permit).

That's already done:
- [permit](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/689/files#diff-a9763b4322d00399b611debb4a9b0a588db404648293647d1e0cbc3d78098075R251)
- [ecdsa example](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/689/files#diff-4291a7075d528531a5eaa322e6ad591e4190f94e698340f1f41b6030d7c94569R25) (updated now)

examples/ecdsa/src/lib.rs
```diff
@@ -20,6 +22,6 @@ impl ECDSAExample {
         r: B256,
         s: B256,
     ) -> Result<Address, ecdsa::Error> {
-        ecdsa::recover(self, hash, v, r, s)
+        self.ecrecover(hash, v, r, s)
```
bidzyyys, line 25: I like it 💯

contracts/src/utils/precompiles.rs
```diff
@@ -0,0 +1,111 @@
+//! `ArbOS` precompiles wrapper enabling easier invocation.
+use alloy_primitives::{Address, B256};
+use primitives::ecrecover::Error;
+use stylus_sdk::prelude::*;
+
+use crate::utils::cryptography::ecdsa::recover;
+
+/// Precompile primitives.
+pub mod primitives {
+    /// The ecrecover precompile primitives.
```
bidzyyys:
```suggestion
    /// The `ecrecover` precompile primitives.
```

qalisander (approved):
Looks quite handy! I think it is a good candidate to be included into stylus-sdk

PR #690, opened 2025-06-05
build: Fix failing cargo-stylus@0.6.0 install -> bump nightly channel to 2025-06-03 

0xNeshi:
This nightly channel is post-Rust-1.87.0, which is [default](https://github.com/OffchainLabs/stylus-hello-world/blob/5b3fa18f70ff58b8964b089be688f9616de54499/rust-toolchain.toml) for new Stylus projects.


0xNeshi:
It seems all nightly versions after v1.85.0 increase the WASM size by about 1KB.
If we want to fix the `edition2024` errors, we'll have to live with the size increase...

qalisander (approved):

PR #696, opened 2025-06-10
fix(audit): L-06

### LINKED ISSUE #695: [Fix(Audit)]: L-06: State Mutability Can Be Restricted in IErc4626

The  IErc4626  trait currently marks several methods with  &mut self  even though they only
need to read contract storage, e.g.,  total_assets ,  convert_to_shares , 
convert_to_assets , etc. By requiring  &mut self  on methods that are conceptually read‐
only, the trait allows implementations to introduce unintended or malicious state changes,
since a developer could override a “read-only” function and write to storage. To preserve
expected invariants and prevent misuse, considering changing the signature of any  IErc4626
method that does not modify state to take  &self  rather than  &mut self 

### Contribution Guidelines

0xNeshi:
Resolves #695 

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed (Breaking)
 
+- Read-only functions in `IErc4626` now accept `&self`. #696
+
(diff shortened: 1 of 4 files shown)


0xNeshi:
Closing in favor of #698

PR #699, opened 2025-06-11
fix(L-05): Erc20Wrapper::decimals matches Solidity implementation

### LINKED ISSUE #638: [Feature]: `ERC20Wrapper::decimals` should match Solidity version

Solidity first checks if underlying decimals can be returned, and if not, only then does it return the wrapper's decimals:
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/56c07e59c4ddc95f77e71ddb54976f13b6863f90/contracts/token/ERC20/extensions/ERC20Wrapper.sol#L37

### Contribution Guidelines

bidzyyys:
Resolves audit issue: L-05: Underlying Decimals In  `Erc20Wrapper` Can Be Set Incorrectly.
Resolves #638


contracts/src/token/erc20/extensions/wrapper.rs
```diff
@@ -224,7 +226,11 @@ impl Erc20Wrapper {
     /// See [`IErc20Wrapper::decimals`].
     #[must_use]
     pub fn decimals(&self) -> U8 {
-        self.underlying_decimals.get()
+        U8::from(
+            IErc20MetadataInterface::new(self.underlying())
+                .decimals(self)
+                .unwrap_or(DEFAULT_DECIMALS),
```
0xNeshi, line 232: I think the correct behavior here would be to return whatever is the base contract's decimals. 

Ahh, it's an even larger problem than just the default decimals (**note: see end of comment for recommended next steps)** :
- `IErc20Wrapper` should technically have `IErc20Metadata` as its supertrait, and `fn decimals` should be implemented as part of `impl IErc20Metadata for Erc20Wrapper`.
- For `Erc20Wrapper::decimals` to have _exactly the same behavior_ as in Solidity, it should have this impl:
```rust
fn decimals(&self, erc20: &Erc20) -> U8 {
    U8::from(
        IErc20MetadataInterface::new(self.underlying())
            .decimals(self)
            .unwrap_or(erc20.decimals()),
}
```

An even larger missing feature: our `Erc20` should implement `IErc20Metadata`, i.e. it should be merged with current `struct Erc20Metadata`, as [it is in Solidity](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/5c79432e4091c20fa5ea057261f0bf528ea70a7c/contracts/token/ERC20/ERC20.sol#L29). The same is the case for [Erc721](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/5c79432e4091c20fa5ea057261f0bf528ea70a7c/contracts/token/ERC721/ERC721.sol#L19) and [Erc1155](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/5c79432e4091c20fa5ea057261f0bf528ea70a7c/contracts/token/ERC1155/ERC1155.sol#L19).

Reasoning: our traits (interfaces) represent and conform to actual EIPs, and we treat our contracts (structs) as simple straightforward implementations of just those traits. 
But the contract implementations themselves should be our recommendations for how these EIPs should be implemented **and _combined_**. At least this is the approach taken in Solidity, and I tend to agree with this approach.

If we were to all agree that this change was necessary, it would be a much bigger change than warranted by this audit report, as it affects multiple contracts, tests, e2e, docs etc. It could be a perfect candidate for M3 though. **For now we can leave the implementation as in this PR**.

0xNeshi: Created issue https://github.com/OpenZeppelin/rust-contracts-stylus/issues/715

PR #717, opened 2025-06-18
build(ci): update nitro-testnode

bidzyyys:
We need to update the nitro-testnode as the CI jobs started failing, like in #716.

### DIFF
--- a/lib/e2e/src/deploy.rs
+++ b/lib/e2e/src/deploy.rs
@@ -16,7 +16,7 @@ use stylus_sdk::{abi::Bytes, alloy_primitives, function_selector};
 use crate::{project::Crate, system::DEPLOYER_ADDRESS, Constructor, Receipt};
 
 const CONTRACT_INITIALIZATION_ERROR_SELECTOR: [u8; 4] =
-    function_selector!("ContractInitializationError", Address);
+    function_selector!("ContractInitializationError", Address, Bytes);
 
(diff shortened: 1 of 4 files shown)


lib/e2e/src/deploy.rs
```diff
@@ -16,7 +16,7 @@ use stylus_sdk::{abi::Bytes, alloy_primitives, function_selector};
 use crate::{project::Crate, system::DEPLOYER_ADDRESS, Constructor, Receipt};
 
 const CONTRACT_INITIALIZATION_ERROR_SELECTOR: [u8; 4] =
-    function_selector!("ContractInitializationError", Address);
+    function_selector!("ContractInitializationError", Address, Bytes);
```
0xNeshi, line 19: Created issue to improve error assertion using this field
https://github.com/OpenZeppelin/rust-contracts-stylus/issues/718

0xNeshi: Short explanation: the Stylus team applied my suggestion to add a [bytes data](https://github.com/OffchainLabs/nitro-contracts/blob/0b8c04e8f5f66fe6678a4f53aa15f23da417260e/src/stylus/StylusDeployer.sol#L16) field to this error emitted by the `StylusDeployer`, which contains the constructor's revert reason, causing the error selector to change, which broke our error handling. 
With this field added in our e2e, the selector is correct again and this now works.

PR #733, opened 2025-07-11
EnumerableSet Trait and Implementations

### LINKED ISSUE #687: [Feature]: Make EnumerableSet generic

Make [EnumerableSet](https://github.com/OpenZeppelin/rust-contracts-stylus/blob/main/contracts/src/utils/structs/enumerable_set.rs) generic so that it will be able to store not only `Address`, but sth else -> `EnumerableSet<T>`.

### Contribution Guidelines

davebryson:
This PR adds a trait definition for `EnumerableSet` along with a macro to generate common sets.  It includes implementations for: `Address, FixedBytes<32>, U8, U16, U32, U64, U128, U256` along with extensive testing.

Resolves #687

#### PR Checklist

- [x] Tests
- [x] Documentation
- [ ] Changelog

### DIFF
--- a/contracts/src/utils/structs/enumerable_set.rs
+++ b/contracts/src/utils/structs/enumerable_set.rs
@@ -9,12 +9,15 @@
 /// * Set can be cleared (all elements removed) in O(n).
 use alloc::{vec, vec::Vec};
 
-use alloy_primitives::{uint, Address, FixedBytes, U256};
+use alloy_primitives::{
+    uint, Address, FixedBytes, U128, U16, U256, U32, U64, U8,
+};
 use stylus_sdk::{
     prelude::*,
     storage::{
         StorageAddress, StorageFixedBytes, StorageKey, StorageMap, StorageType,
-        StorageU256, StorageVec,
+        StorageU128, StorageU16, StorageU256, StorageU32, StorageU64,
+        StorageU8, StorageVec,
     },
 };
 
@@ -86,6 +89,13 @@ where
 macro_rules! impl_set {
     ($($name:ident $key:ident $value:ident)+) => {
         $(
+            ///
+            #[storage]
+            pub struct $name {
+                values: StorageVec<$value>,
+                positions: StorageMap<$key, StorageU256>,
+            }
+
             impl EnumerableSet<$key, $value> for $name {
                 fn add(&mut self, value: $key) -> bool {
                     if self.contains(value) {
@@ -173,55 +183,20 @@ macro_rules! impl_set {
     };
 }
 
-/// State of an [`EnumerableAddressSet`] contract.
-#[storage]
-pub struct EnumerableAddressSet {
-    /// Values in the set.
-    values: StorageVec<StorageAddress>,
-    /// Value -> Index of the value in the `values` array.
-    positions: StorageMap<Address, StorageU256>,
-}
-impl_set!(EnumerableAddressSet Address StorageAddress);
-
-/// Sets to add: Uint
-
-#[storage]
-pub struct EmumerableBytes32Set {
-    /// Values in the set.
-    values: StorageVec<StorageFixedBytes<32>>,
-    /// Value -> Index of the value in the `values` array.
-    positions: StorageMap<FixedBytes<32>, StorageU256>,
-}
 // use alias to avoid macro issues with brackets
 type Bytes32 = FixedBytes<32>;
 type StorageBytes32 = StorageFixedBytes<32>;
-impl_set!(EmumerableBytes32Set Bytes32 StorageBytes32);
 
-/*
-#[storage]
-pub struct EmumerableStringSet {
-    /// Values in the set.
-    values: StorageVec<StorageString>,
-    /// Value -> Index of the value in the `values` array.
-    positions: StorageMap<String, StorageU256>,
-}
-impl_set!(EmumerableStringSet String StorageString);
-*/
-
-macro_rules! impl_uint_sets {
-    ($($uint:ident $int:ident)+) => {
-        $(
-            #[storage]
-            pub struct EmumerableBytes32Set {
-                /// Values in the set.
-                values: StorageVec<StorageFixedBytes<32>>,
-                /// Value -> Index of the value in the `values` array.
-                positions: StorageMap<FixedBytes<32>, StorageU256>,
-            }
-
-        )+
-    };
-}
+impl_set!(
+    EnumerableAddressSet Address StorageAddress
+    EmumerableBytes32Set Bytes32 StorageBytes32
+    EnumerableU8Set U8 StorageU8
+    EnumerableU16Set U16 StorageU16
+    EnumerableU32Set U32 StorageU32
+    EnumerableU64Set U64 StorageU64
+    EnumerableU128Set U128 StorageU128
+    EnumerableU256Set U256 StorageU256
+);
 
 #[cfg(test)]
 mod tests {
--- a/contracts/src/utils/structs/enumerable_set.rs
+++ b/contracts/src/utils/structs/enumerable_set.rs
@@ -9,13 +9,169 @@
 /// * Set can be cleared (all elements removed) in O(n).
 use alloc::{vec, vec::Vec};
 
-use alloy_primitives::{uint, Address, U256};
+use alloy_primitives::{uint, Address, FixedBytes, U256};
 use stylus_sdk::{
     prelude::*,
     storage::{
-        StorageAddress, StorageMap, StorageType, StorageU256, StorageVec,
+        StorageAddress, StorageFixedBytes, StorageKey, StorageMap, StorageType,
+        StorageU256, StorageVec,
     },
 };
+
+/// EnumarableSet trait for defining new sets.
+///
+/// This trait is automatically implemented if using the `impl_set` macro.
+pub trait EnumerableSet<K, V>
+where
+    K: StorageKey,
+    V: StorageType,
+{
+    /// Adds a value to a set.
+    ///
+    /// Returns true if the `value` was added to the set, that is if it was not
+    /// already present.
+    ///
+    /// # Arguments
+    ///
+    /// * `&mut self` - Write access to the set's state.
+    /// * `value` - The value to add to the set.
+    fn add(&mut self, value: K) -> bool;
+
+    /// Removes a `value` from a set.
+    ///
+    /// Returns true if the `value` was removed from the set, that is if it was
+    /// present.
+    ///
+    /// # Arguments
+    ///
+    /// * `&mut self` - Write access to the set's state.
+    /// * `value` - The value to remove from the set.
+    fn remove(&mut self, value: K) -> bool;
+
+    /// Returns true if the `value` is in the set.
+    ///
+    /// # Arguments
+    ///
+    /// * `&self` - Read access to the set's state.
+    /// * `value` - The value to check for in the set.
+    fn contains(&self, value: K) -> bool;
+
+    /// Returns the number of values in the set.
+    ///
+    /// # Arguments
+    ///
+    /// * `&self` - Read access to the set's state.
+    fn length(&self) -> U256;
+
+    /// Returns the value stored at position `index` in the set.
+    ///
+    /// Note that there are no guarantees on the ordering of values inside the
+    /// array, and it may change when more values are added or removed.
+    ///
+    /// # Arguments
+    ///
+    /// * `&self` - Read access to the set's state.
+    /// * `index` - The index of the value to return.
+    fn at(&self, index: U256) -> Option<K>;
+
+    /// Returns the entire set in an array.
+    ///
+    /// # Arguments
+    ///
+    /// * `&self` - Read access to the set's state.
+    fn values(&self) -> Vec<K>;
+}
+
+///
+macro_rules! impl_set {
+    ($($name:ident $key:ident $value:ident)+) => {
+        $(
+            impl EnumerableSet<$key, $value> for $name {
+                fn add(&mut self, value: $key) -> bool {
+                    if self.contains(value) {
+                        false
+                    } else {
+                        self.values.push(value);
+                        // The value is stored at length-1, but we add 1 to all indexes
+                        // and use [`U256::ZERO`] as a sentinel value.
+                        self.positions.setter(value).set(U256::from(self.values.len()));
+                        true
+                    }
+                }
+
+                fn remove(&mut self, value: $key) -> bool {
+                    // We cache the value's position to prevent multiple reads from the same
+                    // storage slot.
+                    let position = self.positions.get(value);
+
+                    if position.is_zero() {
+                        false
+                    } else {
+                        // To delete an element from the `self.values` array in O(1),
+                        // we swap the element to delete with the last one in the array,
+                        // and then remove the last element (sometimes called as 'swap and
+                        // pop'). This modifies the order of the array, as noted
+                        // in [`Self::at`].
+                        let one = uint!(1_U256);
+                        let value_index = position - one;
+                        let last_index = self.length() - one;
+
+                        if value_index != last_index {
+                            let last_value = self
+                                .values
+                                .get(last_index)
+                                .expect("element at `last_index` must exist");
+
+                            // Move the `last_value` to the index where the value to delete
+                            // is.
+                            self.values
+                                .setter(value_index)
+                                .expect(
+                                    "element at `value_index` must exist - is being removed",
+                                )
+                                .set(last_value);
+                            // Update the tracked position of the lastValue (that was just
+                            // moved).
+                            self.positions.setter(last_value).set(position);
+                        }
+
+                        // Delete the slot where the moved value was stored.
+                        self.values.pop();
+
+                        // Delete the tracked position for the deleted slot.
+                        self.positions.delete(value);
+
+                        true
+                    }
+                }
+
+                fn contains(&self, value: $key) -> bool {
+                    !self.positions.get(value).is_zero()
+                }
+
+                fn length(&self) -> U256 {
+                    U256::from(self.values.len())
+                }
+
+                fn at(&self, index: U256) -> Option<$key> {
+                    self.values.get(index)
+                }
+
+                fn values(&self) -> Vec<$key> {
+                    let mut values = Vec::new();
+                    for idx in 0..self.values.len() {
+                        values.push(
+                            self.values
+                                .get(idx)
+                                .expect("element at index: {idx} must exist"),
+                        );
+                    }
+                    values
+                }
+            }
+        )+
+    };
+}
 
 /// State of an [`EnumerableAddressSet`] contract.
 #[storage]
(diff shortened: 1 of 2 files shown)


contracts/src/utils/structs/enumerable_set.rs
```diff
@@ -131,175 +74,985 @@ impl EnumerableAddressSet {
...
-                    .get(idx)
-                    .expect("element at index: {idx} must exist"),
-            );
-        }
-        values
-    }
+    fn values(&self) -> Vec<T>;
+}
+
+// Implements the EnumerableSet trait for the given types
```
bidzyyys:
```suggestion
// Implements the [`EnumerableSet`] trait for the given types.
```

contracts/src/utils/structs/enumerable_set.rs
```diff
@@ -131,175 +74,985 @@ impl EnumerableAddressSet {
...
+macro_rules! impl_set {
+    ($($name:ident $skey:ident $svalue:ident)+) => {
+        $(
+            /// Storage for [`$name`]
```
bidzyyys:
```suggestion
            /// Storage for [`$name`].
```

contracts/src/utils/structs/enumerable_set.rs
```diff
@@ -131,175 +74,985 @@ impl EnumerableAddressSet {
...
+            }
+        )+
+    };
 }
 
+// use alias to avoid macro issues with brackets
+type Bytes32 = FixedBytes<32>;
+type StorageBytes32 = StorageFixedBytes<32>;
+
+impl_set!(
```
bidzyyys: You have a bunch of code duplication in your tests. Eg `U8`, .., `U256` have the same code. Maybe there is a way to generate the tests somehow? If not, how about using some property testing?

bidzyyys (commented):
I like the way you implemented `EnumerableSet`, but I would avoid code duplication in tests for all the data types. Please remeber also to update CHANGELOG.md properly to your changes.

contracts/src/utils/structs/enumerable_set.rs
davebryson: Agree.  I didn't like it either, but wanted to be thorough.  I was originally going to use a macro to generate the `uint` tests.  I'll revisit that or the property test approach.

davebryson: Consolidated all tests via macro that generates property tests for every enumerable_set type.  Much cleaner.

CHANGELOG.md
```diff
@@ -12,13 +12,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add Solidity interface bindings for beacon-related contracts. #729
 - Add internal utilities for interacting with beacon proxies and validating beacon implementations. #729
 - Add `AccessControlEnumerable` extension that supports role member enumeration. #622
-- Add `EnumerableAddressSet`. #622
+- Add EnumerableAddressSet. #622
+- Added `EnumerableSet` trait and macro to generate implementations for different types. #687
```
bidzyyys:
```suggestion
- Add `EnumerableSet` trait and macro to generate implementations for different types. #687
```

CHANGELOG.md
```diff
@@ -12,13 +12,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add Solidity interface bindings for beacon-related contracts. #729
 - Add internal utilities for interacting with beacon proxies and validating beacon implementations. #729
 - Add `AccessControlEnumerable` extension that supports role member enumeration. #622
-- Add `EnumerableAddressSet`. #622
+- Add EnumerableAddressSet. #622
```
bidzyyys:
```suggestion
```

contracts/Cargo.toml
```diff
@@ -21,6 +21,7 @@ openzeppelin-stylus-proc.workspace = true
 [dev-dependencies]
 alloy-primitives = { workspace = true, features = ["arbitrary"] }
 motsu.workspace = true
+proptest.workspace = true
```
0xNeshi:
```suggestion
```
You should be able to access `proptest` crate through `alloy_primitives::private::proptest`

contracts/src/utils/structs/enumerable_set.rs
```diff
@@ -131,175 +75,319 @@ impl EnumerableAddressSet {
...
+                    }
+                    values
+                }
+            }
+        )+
+    };
+}
+
+// use aliases to avoid macro issues with brackets.
+type Bytes32 = FixedBytes<32>;
```
0xNeshi: Isn't this type a duplicate of `alloy_primitives::aliases::B32`?

contracts/src/utils/structs/enumerable_set.rs
```diff
@@ -131,175 +75,319 @@ impl EnumerableAddressSet {
...
+type StorageBytes32 = StorageFixedBytes<32>;
```
0xNeshi: Isn't this type a duplicate of `stylus_sdk::storage::StorageB32`?

contracts/src/utils/structs/enumerable_set.rs
```diff
@@ -131,175 +75,319 @@ impl EnumerableAddressSet {
...
+/// Implements the [`EnumerableSet`] trait for the given types.
+macro_rules! impl_set {
+    ($($name:ident, $skey:ident, $svalue:ident);+ $(;)?) => {
+        $(
+            /// State of an [`$name`] contract.
+            #[storage]
+            pub struct $name {
+                /// Values in the set.
+                values: StorageVec<$svalue>,
+                /// Value -> Index of the value in the `values` array.
```
0xNeshi:
```suggestion
                /// Position is the index of the value in the `values` array plus 1.
                /// Position 0 is used to mean a value is not in the set.
```
I think this is clearer

contracts/src/utils/structs/enumerable_set.rs
```diff
@@ -131,175 +75,319 @@ impl EnumerableAddressSet {
...
+
+            impl EnumerableSet<$skey> for $name {
+                fn add(&mut self, value: $skey) -> bool {
+                    if self.contains(value) {
+                        false
+                    } else {
+                        self.values.push(value);
+                        // The value is stored at length-1, but we add 1 to all indexes
+                        // and use [`U256::ZERO`] as a sentinel value.
+                        self.positions.setter(value).set(U256::from(self.values.len()));
```
0xNeshi:
```suggestion
                        self.positions.setter(value).set(self.length());
```

contracts/src/utils/structs/enumerable_set.rs
```diff
@@ -131,175 +75,319 @@ impl EnumerableAddressSet {
...
+                        // Delete the slot where the moved value was stored.
+                        self.values.pop();
+
+                        // Delete the tracked position for the deleted slot.
+                        self.positions.delete(value);
+
+                        true
+                    }
+                }
+
```
0xNeshi: [Function clear](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/32e7a6ffbc5af9ab0e6dfdbc58508511d0f0b4a2/contracts/utils/structs/EnumerableSet.sol#L235) was missing even before this PR, but please implement it.

0xNeshi (commented):
Overall good job @davebryson !

contracts/src/utils/structs/enumerable_set.rs
bidzyyys: Great catch @0xNeshi: https://docs.rs/stylus-sdk/latest/stylus_sdk/storage/type.StorageB32.html

contracts/src/utils/structs/enumerable_set.rs
bidzyyys: Great catch @0xNeshi: https://docs.rs/stylus-sdk/latest/stylus_sdk/storage/type.StorageB32.html

bidzyyys: It should be: https://docs.rs/stylus-sdk/latest/stylus_sdk/storage/type.StorageB256.html

contracts/src/utils/structs/enumerable_set.rs
bidzyyys: It should be `B256`:  https://docs.rs/alloy-primitives/0.8.20/alloy_primitives/aliases/type.B256.html

contracts/src/utils/structs/enumerable_set.rs
davebryson: @0xNeshi  Didn't catch that!  Thanks!

contracts/src/utils/structs/enumerable_set.rs
bidzyyys: @davebryson leaving the comment for you!

davebryson (commented):
some code and comment clean-up

contracts/src/utils/structs/enumerable_set.rs
davebryson: clear function added

contracts/src/utils/structs/enumerable_set.rs
```diff
@@ -9,25 +9,25 @@
...
-    values: StorageVec<StorageAddress>,
-    /// Value -> Index of the value in the `values` array.
-    positions: StorageMap<Address, StorageU256>,
-}
-
-impl EnumerableAddressSet {
-    /// Adds a value to a set. O(1).
+/// [`EnumerableSet`] trait for defining sets of primitive types.
+///
+/// See the `impl_set` macro for the implementation of this trait for many
```
0xNeshi: Can you make this `impl_set` an actual link to the macro definition?

contracts/src/utils/structs/enumerable_set.rs
```diff
@@ -131,175 +80,358 @@ impl EnumerableAddressSet {
...
-        assert_eq!(contract.sender(alice).at(uint!(0_U256)), Some(alice));
-        assert_eq!(contract.sender(alice).at(uint!(1_U256)), Some(bob));
-        assert_eq!(contract.sender(alice).at(uint!(2_U256)), None);
-    }
+    /// Property tests for [`EnumerableSet<T>`] trait.
+    ///
+    /// The property tests will automatically generate random test cases and
+    /// verify that the [`EnumerableSet<T>`] implementations maintain their
+    /// properties across all types.
+    macro_rules! impl_set_property_tests {
```
0xNeshi: The macro should be renamed like:
`impl_set_property_tests` -> `impl_set_invariant_tests`

These are more like invariant tests than they're true property tests. Property-based tests should verify _mathematical properties_ (e.g. from Set theory) that hold for all inputs, such as:
```rust
// Example of a true property test:
// "Adding an element and then removing it should result in the original set"
proptest!(|(set: Vec<T>, element: T)| {
    let mut test_set = create_set_from_vec(set);
    let original_length = test_set.length();
    
    test_set.add(element);
    test_set.remove(element);
    
    prop_assert_eq!(test_set.length(), original_length);
});
```

Also, remove the `prop_` prefix from all test names.

---
Relatedly, if you'd like to add prop tests, you can start with these:
```rust
// Property: Set operations are idempotent
#[test]
fn prop_add_idempotent() {
    proptest!(|(value: T)| {
        let mut set = create_empty_set();
        let first_add = set.add(value);
        let second_add = set.add(value);
        
        prop_assert!(first_add);
        prop_assert!(!second_add);
        prop_assert_eq!(set.length(), U256::from(1));
    });
}

// Property: Add then remove returns to original state
#[test]
fn prop_add_remove_identity() {
    proptest!(|(values: Vec<T>, new_value: T)| {
        let mut set = create_set_from_vec(values);
        let original_length = set.length();
        
        set.add(new_value);
        set.remove(new_value);
        
        prop_assert_eq!(set.length(), original_length);
    });
}

// Property: Set union properties
#[test]
fn prop_set_union_properties() {
    proptest!(|(values1: Vec<T>, values2: Vec<T>)| {
        let mut set1 = create_set_from_vec(values1);
        let set2 = create_set_from_vec(values2);
        
        // Add all elements from set2 to set1
        for value in set2.values() {
            set1.add(value);
        }
        
        // Property: final set contains all unique elements from both
        let all_values: BTreeSet<_> = values1.iter().chain(values2.iter()).collect();
        let final_values: BTreeSet<_> = set1.values().iter().collect();
        
        prop_assert_eq!(final_values, all_values);
    });
}

// Property: Commutativity of add operations
#[test]
fn prop_add_commutative() {
    proptest!(|(value1: T, value2: T)| {
        let mut set1 = create_empty_set();
        let mut set2 = create_empty_set();
        
        set1.add(value1);
        set1.add(value2);
        
        set2.add(value2);
        set2.add(value1);
        
        prop_assert_eq!(set1.values().sort(), set2.values().sort());
    });
}
```
These are just some prop tests from the Set theory, feel free to add more

contracts/src/utils/structs/enumerable_set.rs
```diff
@@ -131,175 +80,358 @@ impl EnumerableAddressSet {
...
+    fn values(&self) -> Vec<T>;
 }
 
+/// Implements the [`EnumerableSet`] trait for the given types.
+macro_rules! impl_set {
+    ($($name:ident, $skey:ident, $svalue:ident);+ $(;)?) => {
+        $(
+            /// State of an [`$name`] contract.
+            #[storage]
+            pub struct $name {
```
0xNeshi: This makes the docs appear like:
image

This should show the actual `$name` _value_ in the docs

contracts/src/utils/structs/enumerable_set.rs
```diff
@@ -131,175 +80,358 @@ impl EnumerableAddressSet {
...
+                    self.values.erase();
+                }
+
+                fn contains(&self, value: $skey) -> bool {
+                    !self.positions.get(value).is_zero()
+                }
+
+                fn length(&self) -> U256 {
+                    U256::from(self.values.len())
+                }
```
0xNeshi:
```suggestion

                /// See [`EnumerableSet::length`] for general usage.
                ///
                /// This implementation's maximum length will technically be `usize` because
                /// that is the type returned by the underlying [`StorageVec::len`].
                fn length(&self) -> U256 {
                    U256::from(self.values.len())
                }
```
We should make this technical limitation clear to users.

0xNeshi: @davebryson if you feel you're up for an interesting challenge after this PR is merged, try to refactor `EnumerableSet` so that it avoids this limitation, i.e. so that it really does have a maximum length size of `U256`.

This would probably entail having multiple underlying `StorageVec`, the exact number depending on the target platform (as `usize` has variable size). This would also require more comprehensive tests than currently present.

EDIT: There's some talk that the Stylus team will change `StorageVec::len` return type to `U256` at some point. Maybe no need to implement this with multiple `StorageVec`s for now.

contracts/src/utils/structs/enumerable_set.rs
```diff
@@ -131,175 +80,358 @@ impl EnumerableAddressSet {
...
+                        unsafe impl TopLevelStorage for $set_type {}
+
+                        #[public]
+                        impl $set_type {}
+
+                        // verifies that adding values returns correct boolean results and
+                        // all added values are contained.
+                        #[test]
+                        fn prop_add_contains_consistency() {
+                            proptest!(|(values: Vec<$value_type>, alice: Address)| {
```
0xNeshi:
```suggestion
                        #[motsu::test]
                        fn prop_add_contains_consistency(alice: Address) {
                            proptest!(|(values: Vec<$value_type>)| {
```
No need to regenerate `alice` each test run, as it might slow tests down (applies to other tests too)

contracts/src/utils/structs/enumerable_set.rs
```diff
@@ -131,175 +80,358 @@ impl EnumerableAddressSet {
...
+                                let contract = Contract::<$set_type>::default();
+                                let mut expected_values = Vec::new();
```
0xNeshi: Just curious, why sometimes use `Vec` and sometimes `BTreeSet`?

contracts/src/utils/structs/enumerable_set.rs
```diff
@@ -131,175 +80,358 @@ impl EnumerableAddressSet {
...
+                                    }
+                                }
+
+                                prop_assert_eq!(contract.sender(alice).at(length), None);
+                            });
+                        }
+
+                        // confirms that [`EnumerableSet::<T>::values()`] returns exactly the set of added elements.
+                        #[test]
+                        fn prop_values_completeness() {
```
0xNeshi: Isn't this a duplicate of [prop_add_contains_consistency](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/733/files#diff-81232eb6f5bdbeea4b947b640898f51378a9aa66228b352f7c74a25b7081b029R248)?

contracts/src/utils/structs/enumerable_set.rs
```diff
@@ -131,175 +80,358 @@ impl EnumerableAddressSet {
...
+                            proptest!(|(alice: Address, value: $value_type)| {
+                                let contract = Contract::<$set_type>::default();
+
+                                prop_assert_eq!(contract.sender(alice).length(), U256::ZERO);
+                                prop_assert!(!contract.sender(alice).contains(value));
+                                prop_assert_eq!(contract.sender(alice).at(U256::ZERO), None);
+                                prop_assert_eq!(contract.sender(alice).values(), Vec::<$value_type>::new());
+                                prop_assert!(!contract.sender(alice).remove(value));
+                            });
+                        }
```
0xNeshi: This can be just a regular unit test (not proptest) - it makes no difference what the value of `value` is

contracts/src/utils/structs/enumerable_set.rs
```diff
@@ -131,175 +80,358 @@ impl EnumerableAddressSet {
...
+                                    for expected_value in expected_set.iter() {
+                                        prop_assert!(contract.sender(alice).contains(*expected_value));
+                                    }
+                                }
+                            });
+                        }
+
+                        // verifies correct behavior on empty sets.
+                        #[test]
+                        fn prop_empty_set_invariants() {
```
0xNeshi: Missing edge case:
- single-element set tests (remove, access at(0) and at(1), clear)

0xNeshi: Missing `EnumerableSet::clear` invocation on empty set, just to verify it at least does not panic

contracts/src/utils/structs/enumerable_set.rs
```diff
@@ -131,175 +80,358 @@ impl EnumerableAddressSet {
...
+
+                                prop_assert_eq!(contract.sender(alice).at(U256::ZERO), None);
+                            });
+                        }
+                    }
+                )+
+            };
+        }
+
+    impl_set_property_tests!(
```
qalisander: I think we also need string set like in [solidity](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/32e7a6ffbc5af9ab0e6dfdbc58508511d0f0b4a2/contracts/utils/structs/EnumerableSet.sol#L509)

contracts/src/utils/structs/enumerable_set.rs
davebryson: Making it a link: [`impl_set`] displays correctly in vscode, but rustdoc results in a intra-doc link error: 

*macro_rules named impl_set exists in this crate, but it is not in scope at this link's location*.  

I've tried adjusting the link to bring it in scope...but no luck.  This seems to be a know issue with `macros` and links

davebryson: Thinking a little more, does the macro actually belong in the docs?  It's an implementation details not for API related use

contracts/src/utils/structs/enumerable_set.rs
davebryson: On the `prop_add_remove_sequence_invariants` test the `btreeset` insert/remove returns a bool for an additional check.  However, `btreeset` is not needed in the `prop_values_completeness' test

contracts/src/utils/structs/enumerable_set.rs
davebryson: @0xNeshi  Good catch.   Would it more consistent then to have the trait use `fn length(&self) -> usize` vs. returning `U256`?  This is more directly inline with the underlying `StorageVec` which this API has no control over.  Adding a work around for the limitation would seem to add unnecessary complexity.

davebryson: @0xNeshi  Thinking about this more, the max size of a `usize` on a 64 bit arch is 2^64.  While a `StorageVec` could potentially have a length larger than that, the storage cost would probably prevent anyone from actually doing so.
  
Rather than add the additional complexity to handle this, I think it'd be better to return the `usize`, to be inline with the `StorageVec` API and add the technical limitations to the comments as you mention above. 

If there's a strong need to overcome the `usize` limitation, that should probably happen in StorageVec vs. in this library.

Thoughts?

contracts/src/utils/structs/enumerable_set.rs
davebryson: @qalisander I'll add that next after cleaning up the testing recommendations above

contracts/src/utils/structs/enumerable_set.rs
davebryson: @0xNeshi  I have a fix for this in my next push

contracts/src/utils/structs/enumerable_set.rs
0xNeshi: The only reason to keep this return type `U256` is for consistent ABI with Solidity's version. This would allow existing users/clients to call the Stylus contract as they would a Solidity one.

There's some talk that the Stylus team will change `StorageVec::len` return type to `U256` at some point. Maybe no need to implement this with multiple `StorageVec`s for now after all :+1:

contracts/src/utils/structs/enumerable_set.rs
0xNeshi: ~~That makes sense, and even if it were public, users of the trait don't need to know about `impl_set`.~~
~~You can remove the reference~~

EDIT: On second thought, it _does_ make sense for users of this trait to be aware of `impl_set` - they become aware that they don't need to manually implement the trait, when they can use this macro.
You should leave the (proper) link to the macro and find a way to fix this intra-doc link error.
_Hint: the macro itself should be publicly exposed, so that users can implement the trait for their own storage types - this might help solve the link error._

contracts/src/utils/structs/enumerable_set.rs
davebryson: @neshi Interestingly if you run these two test together, one will panic with `storage locked` from `motsu`.   If you run one in isolation, there are no problems.   Any idea what may be going on?
```rust
                      // tests commutativity: order of adding elements doesn't affect final set
                      #[motsu::test]
                       fn commutativity_add(
                            contract1: Contract::<$set_type>,
                            contract2: Contract::<$set_type>,
                            alice: Address,
                            bob: Address
                        ) {
                            proptest!(|(mut values: Vec<$value_type>)| {
                                if values.len() < 2 { return Ok(()); }

                                // Add elements in original order to contract1
                                for value in values.iter() {
                                    contract1.sender(alice).add(*value);
                                }

                                // Add elements in reverse order to contract2
                                values.reverse();
                                for value in values.iter() {
                                    contract2.sender(bob).add(*value);
                                }

                                // Both sets should be identical
                                prop_assert_eq!(contract1.sender(alice).length(), contract2.sender(bob).length());

                                let values1 = contract1.sender(alice).values();
                                let values2 = contract2.sender(bob).values();
                                let set1: BTreeSet<_> = values1.into_iter().collect();
                                let set2: BTreeSet<_> = values2.into_iter().collect();
                                prop_assert_eq!(set1, set2);
                            });
                        }

                        // tests associativity: grouping of operations doesn't matter
                        #[motsu::test]
                        fn associativity_operations(
                            contract1: Contract::<$set_type>,
                            contract2: Contract::<$set_type>,
                            alice: Address,
                            bob: Address
                        ) {
                            proptest!(|(values: Vec<$value_type>)| {
                                if values.len() < 3 { return Ok(()); }

                                let mid = values.len() / 2;

                                // Strategy 1: Add first half, then second half
                                for value in values[..mid].iter() {
                                    contract1.sender(alice).add(*value);
                                }
                                for value in values[mid..].iter() {
                                    contract1.sender(alice).add(*value);
                                }

                                // Strategy 2: Add all at once
                                for value in values.iter() {
                                    contract2.sender(bob).add(*value);
                                }

                                // Both strategies should result in identical sets
                                prop_assert_eq!(contract1.sender(alice).length(), contract2.sender(bob).length());

                                let values1 = contract1.sender(alice).values();
                                let values2 = contract2.sender(bob).values();
                                let set1: BTreeSet<_> = values1.into_iter().collect();
                                let set2: BTreeSet<_> = values2.into_iter().collect();
                                prop_assert_eq!(set1, set2);
                            });
                        }
```

davebryson: @0xNeshi  I'm running into an issue coming from `motsu`.  Various tests will pass and then **randomly** fail as a result of a panic: `storage locked`, from `storage_access` A test may pass 10 times in a row, then fail. 

```
motsu-0.10.0/src/storage_access.rs:56:25:
storage is locked
...
```   

Here's an example of one of the tests.  But other tests have the same issue.    Most often this is related to the b256 set. 

Have you seen this problem before when running multiple tests? 

```rust
#[motsu::test]
 fn inverse_add_remove(contract: Contract::<$set_type>, alice: Address) {
      proptest!(|(values: Vec<$value_type>)| {
          // Add all unique values
          let mut unique_values = Vec::new();
          for value in values.iter() {
              if contract.sender(alice).add(*value) {
                  unique_values.push(*value);
               }
           }

           // Remove all added values
           for value in unique_values.iter() {
               let was_removed = contract.sender(alice).remove(*value);
               prop_assert!(was_removed);
            }

            // Should be back to empty state
            prop_assert_eq!(contract.sender(alice).length(), U256::ZERO);
            prop_assert_eq!(contract.sender(alice).values(), Vec::<$value_type>::new());

            // None of the values should be contained anymore
            for value in unique_values.iter() {
                prop_assert!(!contract.sender(alice).contains(*value));
            }
      });
}
```

davebryson:
@qalisander @bidzyyys @0xNeshi   It looks like PR #751 is evolving in a (nice) different direction. Should I close this PR?

contracts/src/utils/structs/enumerable_set.rs
0xNeshi: It's probably because all of the `proptest!` runs target the same `contract` instance, which is initiated only once in the motsu test signature. This is problematic not only because of the race condition you experienced, but because each test run should have its own _clean_ contract storage, uncorrupted by previous test runs.

Try instantiating the contract directly inside the `proptest!` macro with `Contract::<$set_type>::new()`, see if that fixes the issue.

contracts/src/utils/structs/enumerable_set.rs
davebryson: Delay this waiting on desired implementation approach

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add `EnumerableSet` generic type. #687
```
bidzyyys:
```suggestion
- Add `EnumerableSet` generic type. #733
```

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Add `EnumerableSet` implementation for: `Address`, `B256`, `U8`, `U16`, `U32`, `U64`, `U128`, `U256`, `Bytes`. #687
```
bidzyyys:
```suggestion
- Add `EnumerableSet` implementation for: `Address`, `B256`, `U8`, `U16`, `U32`, `U64`, `U128`, `U256`, `Bytes`. #733
```

CHANGELOG.md
```diff
@@ -17,6 +19,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The `Precompiles::ec_recover` is now callable on `&self`. #754
 - The `ecdsa::recover` function now accepts `impl StaticCallContext` instead of `&mut impl TopLevelStorage`. #754
 
+### Changed
+
+- Remove initial `EnumerableAddressSet` implementation. #687
```
bidzyyys:
```suggestion
- Remove initial `EnumerableAddressSet` implementation. #733
```

bidzyyys (approved):

CHANGELOG.md
```diff
@@ -17,6 +19,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The `Precompiles::ec_recover` is now callable on `&self`. #754
 - The `ecdsa::recover` function now accepts `impl StaticCallContext` instead of `&mut impl TopLevelStorage`. #754
 
+### Changed
+
+- Remove initial `EnumerableAddressSet` implementation. #733
```
0xNeshi: Move this to breaking changes

CHANGELOG.md
```diff
@@ -17,6 +19,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
 ### Changed (Breaking)
 
 - Rename `IERC721Receiver` Solidity Interface to `IErc721ReceiverInterface`. #743
```
0xNeshi, line 25:
```suggestion
- Remove initial `EnumerableAddressSet` implementation. #733
- Rename `IERC721Receiver` Solidity Interface to `IErc721ReceiverInterface`. #743
```

CHANGELOG.md
```diff
@@ -17,6 +19,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The `Precompiles::ec_recover` is now callable on `&self`. #754
 - The `ecdsa::recover` function now accepts `impl StaticCallContext` instead of `&mut impl TopLevelStorage`. #754
 
+### Changed
+
+- Remove initial `EnumerableAddressSet` implementation. #733
+
```
0xNeshi:
```suggestion
```

PR #738, opened 2025-07-17
build: add more curves for Pedersen Hash

bidzyyys:
After evaluating other Pedersen Hash variants (e.g., bitwise Pedersen, Jubjub-style from Zcash/Sapling), I've decided not to integrate additional variants due to WASM contract size limitations and missing dependencies.

- Zcash-style Pedersen requires dynamic generator derivation via Blake2s, hash-to-curve mapping, and windowed scalar multiplication. We currently do not have our own Blake2s implementation, and adding one would significantly increase code size and complexity.
- Bitwise-input variants (e.g., zkSync, Arkworks-style) require 256–512 generator points or non-trivial logic for dynamic generation. Both approaches introduce substantial storage or runtime overhead, making them unsuitable for our on-chain constraints.
- Our current field-splitting approach (StarkNet-style) uses only 5 fixed generators, provides predictable performance, and fits well within WASM limits. It fully satisfies our foreseeable needs.

I’ve also added support for multiple elliptic curve instances that can be used to parametrize the existing Pedersen hash implementation. This makes the system flexible for different curve backends without increasing contract size.


qalisander (approved):
Good job! This approach looks good to me!

PR #751, opened 2025-07-24
feat: generic EnumerableSet

### LINKED ISSUE #687: [Feature]: Make EnumerableSet generic

qalisander:
Implementation of EnumerableSet with generic types

Resolves #687

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog


qalisander:
fix clippy lints

contracts/src/utils/structs/enumerable_set/generic_size.rs
```diff
@@ -0,0 +1,61 @@
...
+};
+
+/// Trait that associate set element with storage type.
+pub trait Element: StorageKey + Copy {
+    /// Set element type in storage.
+    type StorageElement: for<'a> StorageType<Wraps<'a> = Self>
+        + Accessor<Wraps = Self>
+        + for<'a> SimpleStorageType<'a>
+        + Erase;
+}
```
0xNeshi, line 19:
```suggestion
/// Trait that associate set element with storage type.
pub trait Element: StorageKey + Copy {
    /// Set element type in storage.
    type StorageElement: for<'a> StorageType<Wraps<'a> = Self>
        + Accessor<Wraps = Self>
        + Erase;
}
```
Deleting the `SimpleStorageType` supertrait would allow `Element` to be implemented for `StorageString` and `StorageBytes`. The `Element` trait could then be separately implemented for all `T: SimpleStorageType`.

This should auto-implement `EnumerableSet<T>` for the latter two missing storage types as well.

contracts/src/utils/structs/enumerable_set/generic_size.rs
```diff
@@ -0,0 +1,61 @@
+//! Contains generic size utilities for enumerable set storage contract.
```
0xNeshi, line 1: File name `generic_size.rs` is misleading, as there are no "sizes" defined.

Also, if devs wanted to create `EnumerableSet` on their custom storage types, they would have to implement the below two traits. As this might not be immediately clear to dev, we should provide a clear example either here in the mod-level docs or in Antora for how this should be done.

contracts/src/utils/structs/enumerable_set/generic_size.rs
```diff
@@ -0,0 +1,61 @@
...
+use stylus_sdk::{
+    prelude::{Erase, SimpleStorageType, StorageType},
+    storage::{
+        StorageAddress, StorageB256, StorageKey, StorageU128, StorageU16,
+        StorageU256, StorageU32, StorageU64, StorageU8,
+    },
+};
+
+/// Trait that associate set element with storage type.
+pub trait Element: StorageKey + Copy {
```
0xNeshi, line 13: This is like an inverted `StorageType` :D Wouldn't using a definition like `Element: StorageType + Accessor` work (omitted lifetimes for simplicity)?

bidzyyys:
Closing as the PR has been merged to #687.

PR #752, opened 2025-07-24
feat: UUPS Proxy

### LINKED ISSUE #749: [Feature]: Proxy-UUPS

In #14 we added all the primitives (like ERC1967) and Beacon Proxy, but forgot to add UUPS contract.

### Contribution Guidelines

bidzyyys:
Resolves #749 

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog


contracts/src/proxy/utils/erc1822.rs
```diff
@@ -0,0 +1,46 @@
...
+    /// Returns the storage slot that the proxiable contract assumes is being
+    /// used to store the implementation address.
+    ///
+    /// # Arguments
+    ///
+    /// * `&self` - Read access to the contract's state.
+    ///
+    /// # Errors
+    ///
+    /// TODO!
```
bidzyyys: Missing error section

contracts/src/proxy/erc1967/utils.rs
```diff
@@ -99,28 +99,34 @@ impl MethodError for Error {
...
         .finalize();
-    U256::from_be_bytes(HASH).wrapping_sub(uint!(1_U256))
+    B256::new(
+        U256::from_be_bytes(HASH).wrapping_sub(uint!(1_U256)).to_be_bytes(),
+    )
 };
 
 /// Storage slot with the admin of the contract.
-const ADMIN_SLOT: U256 = {
+const ADMIN_SLOT: B256 = {
```
0xNeshi: Make all constants `pub`

contracts/src/utils/storage_slot.rs
```diff
@@ -32,12 +32,12 @@ const SLOT_BYTE_SPACE: u8 = 32;
 /// #[public]
 /// impl Erc1967 {
 ///     fn get_implementation(&self) -> Address {
-///         return StorageSlot::get_slot::<StorageAddress>(IMPLEMENTATION_SLOT.into()).get();
+///         return StorageSlot::get_slot::<StorageAddress>(IMPLEMENTATION_SLOT).get();
```
0xNeshi, line 35: Nice catch, it was never necessary to do `.into()` in these docs

examples/ownable/src/lib.rs
```diff
@@ -64,6 +64,12 @@ impl OwnableExample {
     fn constructor(&mut self, initial_owner: Address) -> Result<(), Error> {
         Ok(self.ownable.constructor(initial_owner)?)
     }
+
+    // Dummy function for some other E2E tests.
```
0xNeshi, line 68: Reference the exact e2e tests

contracts/src/proxy/utils/uups_upgradeable.rs
```diff
@@ -0,0 +1,866 @@
...
+    ///   fails.
+    /// * [`Error::FailedCallWithReason`] - If the delegate call fails with a
+    ///   specific reason.
+    ///
+    /// # Events
+    ///
+    /// * [`crate::proxy::erc1967::Upgraded`]: Emitted when the implementation
+    ///   is upgraded.
+    #[selector(name = "upgradeToAndCall")]
+    fn upgrade_to_and_call(
```
0xNeshi: Isn't the selector redundant?

0xNeshi: Doesn't need to be addressed now, but it makes little sense to list errors for trait function defs that return `Vec<u8>` errors. The most we can say here is that "the function returns any errors that might happen by the implementing contract", and let the implementing contract's function docs be the ones that lists the exact errors.

examples/uups-proxy/src/lib.rs
```diff
@@ -0,0 +1,148 @@
...
+#[implements(IErc20<Error = erc20::Error>, IUUPSUpgradeable, IErc1822Proxiable, IOwnable)]
+impl UUPSProxyErc20Example {
+    #[constructor]
+    fn constructor(
+        &mut self,
+        initial_owner: Address,
+    ) -> Result<(), ownable::Error> {
+        self.uups.constructor();
+        self.ownable.constructor(initial_owner)
+    }
```
0xNeshi, line 38: You might not even need the constructor, as it only initializes the the logic contract's state, which is completely irrelevant to the proxy

examples/uups-proxy/src/lib.rs
```diff
@@ -0,0 +1,148 @@
...
+
+    fn mint(&mut self, to: Address, value: U256) -> Result<(), erc20::Error> {
+        self.erc20._mint(to, value)
+    }
+
+    /// Initializes the contract.
+    ///
+    /// NOTE: Make sure to provide a proper initialization in your logic
+    /// contract, [`Self::initialize`] should be invoked at most once.
+    fn initialize(
```
0xNeshi: This is a good workaround to initialize the `self_address` properly.

The downside is that it relies on the developer to properly implement the requirement that ```[`Self::initialize`] should be invoked at most once```.

examples/uups-proxy/src/lib.rs
```diff
@@ -0,0 +1,148 @@
...
+        self_address: Address,
+        owner: Address,
+    ) -> Result<(), ownable::Error> {
+        // Ugly hack with setting the `self_address` storage value.
+        // Stylus SDK doesn't support setting the immutable storage values as
+        // in Solidity:
+        //
+        // ```solidity
+        // address private immutable __self = address(this);
+        self.uups.self_address.set(self_address);
```
0xNeshi: Explanation for the reason `self_address` is a proper storage field instead of `immutable` like in Solidity should be part of the docs

contracts/src/proxy/utils/uups_upgradeable.rs
```diff
@@ -0,0 +1,866 @@
...
+    #[payable]
+    fn upgrade_to_and_call(
+        &mut self,
+        new_implementation: Address,
+        data: Bytes,
+    ) -> Result<(), Vec<u8>> {
+        self.only_proxy()?;
+        self._upgrade_to_and_call_uups(new_implementation, &data)?;
+        Ok(())
+    }
```
0xNeshi: 1. Isn't the selector redundant?
2. The virtual [_authorizeUpgrade](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/3790c59623e99cb0272ddf84e6a17a5979d06b35/contracts/proxy/utils/UUPSUpgradeable.sol#L87) function is missing; how did you imagine this be achieved in Stylus?

contracts/src/proxy/utils/uups_upgradeable.rs
```diff
@@ -0,0 +1,866 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `&self` - Read access to the contract's state.
+    ///
+    /// # Errors
+    ///
+    /// * [`Error::UnauthorizedCallContext`] - If the execution is performed via
+    ///   delegate call.
+    fn _check_not_delegated(&self) -> Result<(), Error> {
```
0xNeshi: This and `not_delegated` can be merged into one function.

The probable reason why they are separated in Solidity is that functions can't be used as modifiers directly and also that this function should be available to inheriting contracts.

contracts/src/proxy/utils/uups_upgradeable.rs
```diff
@@ -0,0 +1,866 @@
...
+    /// # Arguments
+    ///
+    /// * `&self` - Read access to the contract's state.
+    ///
+    /// # Errors
+    ///
+    /// * [`Error::UnauthorizedCallContext`] - If the execution is not performed
+    ///   via delegate call or the execution context is not of a proxy with an
+    ///   ERC-1967 compliant implementation pointing to self.
+    fn _check_proxy(&self) -> Result<(), Error> {
```
0xNeshi: This and `only_proxy` can be merged, see previous comment for explanation

contracts/src/proxy/utils/uups_upgradeable.rs
```diff
@@ -0,0 +1,866 @@
...
+        #[implements(IErc20<Error = erc20::Error>, IUUPSUpgradeable, IErc1822Proxiable, IOwnable)]
+        impl UUPSErc20Example {
+            #[constructor]
+            pub(super) fn constructor(
+                &mut self,
+                initial_owner: Address,
+            ) -> Result<(), ownable::Error> {
+                self.uups.constructor();
+                self.ownable.constructor(initial_owner)
+            }
```
0xNeshi: redundant, logic contract storage is ignored

contracts/src/proxy/utils/uups_upgradeable.rs
```diff
@@ -0,0 +1,866 @@
...
+    /// Initializes the contract by storing its own address for later context
+    /// validation.
+    ///
+    /// # Arguments
+    ///
+    /// * `&mut self` - Write access to the contract's state.
+    #[constructor]
+    pub fn constructor(&mut self) {
+        self.self_address.set(contract::address());
+    }
```
0xNeshi:
```suggestion
    /// Initializes the contract by storing the self address for later context
    /// validation.
    ///
    /// # Arguments
    ///
    /// * `&mut self` - Write access to the contract's state.
    /// * `self_address` - Own address passed after deployment.
    pub fn initialize(&mut self, self_address: Address) {
        self.self_address.set(self_address);
    }
```
Constructor is redundant, as the contract is only ever used as a logic contract behind a proxy.

`fn initialize(&mut self, self_address: Address)` on the other hand would be much more handy to be called when [initializing the contract through proxies](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/752/files#diff-d776663c727654f643a5f1589334c1ca2a2f3cf29168c10e4db4980d9a14fc04R443).
Technically you could leave this constructor and use it in initializers, but I'd argue that having a constructor might cause confusion, as devs would expect that it has to be called when logic contract is being deployed (which it does not).

NOTE: It would be awesome to implement `Initializable` at some point to properly set the logic up...

contracts/src/proxy/utils/uups_upgradeable.rs
```diff
@@ -0,0 +1,866 @@
...
+                sender: alice,
+                balance: U256::ZERO,
+                needed: amount,
+            }
+            .abi_encode()
+        );
+    }
+
+    #[motsu::test]
+    fn upgrade_by_non_owner_fails(
```
0xNeshi: Isn't this testing `Ownable`? How does testing `Ownable` increase reliability of UUPS?

contracts/src/proxy/utils/uups_upgradeable.rs
```diff
@@ -0,0 +1,866 @@
...
+    use crate::{access::ownable, token::erc20};
+
+    #[cfg_attr(coverage_nightly, coverage(off))]
+    mod test_contracts {
+        use alloy_sol_macro::sol;
+        use stylus_sdk::{alloy_primitives::Address, prelude::*, ArbResult};
+
+        use super::*;
+        use crate::{
+            access::ownable::{IOwnable, Ownable},
```
0xNeshi: nit: `Ownable` makes more sense in examples than in unit tests; the latter verifies specific contract's behavior, while the former is our recommendation for how the contract _should_ be implemented.

examples/uups-proxy/tests/uups-proxy.rs
```diff
@@ -0,0 +1,800 @@
...
+            balance: U256::ZERO,
+            needed: U256::from(1000),
+        }
+    ));
+
+    Ok(())
+}
+
+#[e2e::test]
+async fn upgrade_by_non_owner_fails(
```
0xNeshi, line 232: Possibly redundant test, it verifies `Ownable` behavior

examples/uups-proxy/tests/uups-proxy.rs
```diff
@@ -0,0 +1,800 @@
...
+
+    // check that the balance can be fetched through the upgraded proxy.
+    let balance = proxy_contract.balanceOf(alice_addr).call().await?.balance;
+    assert_eq!(balance, U256::ZERO);
+
+    Ok(())
+}
+
+#[e2e::test]
+async fn upgrades_with_ownership_transfer(
```
0xNeshi: Isn't this, again, testing `Ownable` behavior?

examples/uups-proxy/tests/uups-proxy.rs
```diff
@@ -0,0 +1,800 @@
...
+
+    // confirm implementation didn't change.
+    let new_impl = proxy_contract.implementation().call().await?.implementation;
+    assert_eq!(new_impl, logic_addr);
+
+    Ok(())
+}
+
+#[e2e::test]
+async fn upgrade_after_renounce_ownership_reverts(
```
0xNeshi: `Ownable` test

examples/uups-proxy/tests/uups-proxy.rs
```diff
@@ -0,0 +1,800 @@
...
+        UUPSProxyErc20Example::UUPSUnsupportedProxiableUUID {
+            slot: B256::ZERO
+        }
+    ));
+
+    Ok(())
+}
+
+#[e2e::test]
+async fn upgrades_preserve_storage(
```
0xNeshi, line 531: I like that tests covered these contract invariants, good job :fire:

examples/uups-proxy/tests/uups-proxy.rs
bidzyyys: But this test shows what may happen with proxy. This one I am going to keep.

contracts/src/proxy/utils/uups_upgradeable.rs
bidzyyys: 1. Yes, removed.
2. Added comment where `_authorizeUpgrade` should take place.

examples/uups-proxy/src/lib.rs
```diff
@@ -0,0 +1,154 @@
...
+    ownable: Ownable,
+    uups: UUPSUpgradeable,
+}
+
+#[public]
+#[implements(IErc20<Error = erc20::Error>, IUUPSUpgradeable, IErc1822Proxiable, IOwnable)]
+impl UUPSProxyErc20Example {
+    #[constructor]
+    #[allow(deprecated)]
+    fn constructor(
```
bidzyyys, line 32: TODO: Try to remove constructor.

examples/uups-proxy/tests/uups-proxy.rs
0xNeshi: It's failing on the ownable-related check in [upgrade_to_and_call](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/752/files/5b6b7bf9244c5e849560144b68127463714567d6#diff-41872fca53e93a90c7c631cf2c319533ee7715b2ce4263bdc612298523b6e529R118):
```rust
        self.ownable.only_owner()?;
```
- Having this check as part of example makes sense :heavy_check_mark: 
- E2E however should be testing **UUPS Proxy behavior**, not ownable behavior.

I wouldn't mind if you left it in, but I don't think that it increases UUPS reliability, it just increases test execution time. If you think we should keep the test, feel free to resolve the conversation

examples/uups-proxy/src/lib.rs
0xNeshi: Related to https://github.com/OpenZeppelin/rust-contracts-stylus/pull/752#discussion_r2247218569 and https://github.com/OpenZeppelin/rust-contracts-stylus/pull/752#discussion_r2247283755

contracts/src/proxy/utils/uups_upgradeable.rs
0xNeshi: See https://github.com/OpenZeppelin/rust-contracts-stylus/pull/752#discussion_r2247555494

contracts/src/proxy/utils/uups_upgradeable.rs
0xNeshi: See https://github.com/OpenZeppelin/rust-contracts-stylus/pull/752#discussion_r2247555494

examples/uups-proxy/src/lib.rs
0xNeshi: @bidzyyys we actually _NEED_ the constructors for `proxiableUUID` to work properly, so a UUPS contract in Stylus must be deployed first by invoking the constructor as you did in this implementation, and they also by invoking the `fn initialize` from the proxy...

The reason is that internal UUPS storage is actually used in one case - when invoking `proxiableUUID`. This function can only ever be invoked directly on the logic contract itself.
The reason we need to initiate the `self_address` field both in the UUPS constructor _and_ in its `fn initialize` is that we need to bake in the same `self_address` value in both the proxy and the logic contract separately. This simulates Solidity's `immutable`, which bakes in the `__self` value during deployment in the logic contract.

NOTE: wdyt about just leaving only `constructor` **without** `fn initialize`, since they will perform the same logic?

0xNeshi (commented):
Responded to some comments

examples/uups-proxy/src/lib.rs
bidzyyys: I would leave `initialize` function to show people the proper attitude -- this is the use case for initializers.

contracts/src/proxy/utils/uups_upgradeable.rs
```diff
@@ -0,0 +1,717 @@
...
+        self._upgrade_to_and_call_uups(new_implementation, &data)?;
+        Ok(())
+    }
+}
+
+impl UUPSUpgradeable {
+    /// Check that the execution is being performed through a
+    /// [`stylus_sdk::call::delegate_call`] call and that the execution
+    /// context is a proxy contract with an implementation (as defined in
+    /// ERC-1967) pointing to `self`. This should only be the case for
```
0xNeshi:
```suggestion
    /// Check that the execution is being performed through a
    /// [`stylus_sdk::call::delegate_call`] call and that the execution
    /// context is a proxy contract with an implementation (as defined in
    /// ERC-1967) pointing to `self`.
    ///
    /// This should only be the case for
```
First sentence should be short, see https://rust-lang.github.io/rust-clippy/master/index.html#too_long_first_doc_paragraph

Also, link to the ERC in this doc comment too, would be make it easier to navigate to the ERC from the function.

docs/modules/ROOT/pages/uups-proxy.adoc
```diff
@@ -0,0 +1,245 @@
...
+* DAO frameworks.
+* NFT marketplaces.
+* Access control registries.
+* Cross-chain bridges.
+
+== Related
+
+* xref:erc1967.adoc[ERC-1967 Proxy]
+* xref:beacon-proxy.adoc[Beacon Proxy]
+* xref:proxy.adoc[Basic Proxy]
```
0xNeshi, line 257: Check if other proxy adocs link to uups-proxy too

contracts/src/proxy/utils/uups_upgradeable.rs
```diff
@@ -0,0 +1,717 @@
...
+
+#[public]
+impl IErc1822Proxiable for UUPSUpgradeable {
+    fn proxiable_uuid(&self) -> Result<B256, Vec<u8>> {
+        self.not_delegated()?;
+        Ok(IMPLEMENTATION_SLOT)
+    }
+}
+
+impl UUPSUpgradeable {
```
0xNeshi, line 279: `UUPSUpgradeable` should have a constructor that initializes `self_address`, to make it clear to devs that they are supposed to invoke this when deploying an upgradeable contract. You can then use it in [e2e](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/752/files#diff-41872fca53e93a90c7c631cf2c319533ee7715b2ce4263bdc612298523b6e529R62)

examples/uups-proxy/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+        value: U256,
+    ) -> Result<bool, Self::Error> {
+        self.erc20.approve(spender, value)
+    }
+}
+
+#[public]
+impl IUUPSUpgradeable for UUPSProxyErc20Example {
+    #[selector(name = "UPGRADE_INTERFACE_VERSION")]
+    fn upgrade_interface_version(&self) -> String {
```
0xNeshi, line 108: Missing e2e test for this function.

Since the underlying [self.uups.upgrade_interface_version](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/752/files#diff-d776663c727654f643a5f1589334c1ca2a2f3cf29168c10e4db4980d9a14fc04R184) logic depends on default trait impl, and taking into consideration possible Stylus SDK limitations related to this, we should ensure that this function:
1. Is callable on the deployed contract
2. Returns the expected value

docs/modules/ROOT/pages/uups-proxy.adoc
bidzyyys: Done.

examples/uups-proxy/src/lib.rs
bidzyyys: It's a const value in solidity and cannot have `UPGRADE_INTERFACE_VERSION` name. Work in E2E tests with `upgradeInterfaceVersion`.

0xNeshi: Public constant values in Solidity cause the compiler to automatically generate getters for them. This means that `UPGRADE_INTERFACE_VERSION` _does_ in fact exist in Solidity.
Consider updating e2e tests to use `function UPGRADE_INTERFACE_VERSION() public pure returns (string)`.

examples/uups-proxy/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+        spender: Address,
+        value: U256,
+    ) -> Result<bool, Self::Error> {
+        self.erc20.approve(spender, value)
+    }
+}
+
+#[public]
+impl IUUPSUpgradeable for UUPSProxyErc20Example {
+    fn upgrade_interface_version(&self) -> String {
```
0xNeshi, line 108: Add back the all-capital-letters selector, see https://github.com/OpenZeppelin/rust-contracts-stylus/pull/752#discussion_r2253253272

docs/modules/ROOT/pages/uups-proxy.adoc
```diff
@@ -0,0 +1,245 @@
...
+- **Standardized** — Conforms to ERC-1822 and ERC-1967.
+- **Flexible** — Upgrade logic can include custom access control and validation.
+- **Safe by Design** — Uses dedicated ERC-1967 slots to prevent storage collisions.
+
+== How It Works
+
+. Deploy `Erc1967Proxy` with an initial implementation and encoded `initialize` data.
+. Proxy delegates all calls to the implementation contract via `delegate_call`.
+. Implementation exposes `upgradeToAndCall`, guarded by access control (e.g. `Ownable`).
+. Upgrades validate the new implementation using `proxiableUUID()`.
```
0xNeshi: should use snake_case?

docs/modules/ROOT/pages/uups-proxy.adoc
```diff
@@ -0,0 +1,245 @@
...
+
+== Implementing a UUPS Contract (Rust / Stylus)
```
0xNeshi: These are `Rust / Stylus` docs, is mentioning this in the subtitle necessary?

docs/modules/ROOT/pages/uups-proxy.adoc
```diff
@@ -0,0 +1,245 @@
...
+    ownable: Ownable,
+    uups: UUPSUpgradeable,
+}
+
+#[public]
+#[implements(IErc20<Error = erc20::Error>, IUUPSUpgradeable, IErc1822Proxiable, IOwnable)]
+impl MyUUPSContract {
+    /// Initializes the contract.
+    ///
+    /// NOTE: Make sure to provide a proper initialization in your logic
```
0xNeshi: Add constructor to the adoc example

contracts/src/proxy/utils/uups_upgradeable.rs
```diff
@@ -0,0 +1,743 @@
...
+#[interface_id]
+pub trait IUUPSUpgradeable: IErc1822Proxiable {
+    /// Returns the version of the upgrade interface of the contract.
+    ///
+    /// # Arguments
+    ///
+    /// * `&self` - Read access to the contract's state.
+    fn upgrade_interface_version(&self) -> String {
+        UPGRADE_INTERFACE_VERSION.into()
+    }
```
0xNeshi: Maybe include a note reminding devs to set the all-cap selector

contracts/src/proxy/utils/uups_upgradeable.rs
```diff
@@ -0,0 +1,743 @@
...
+    /// validation.
+    ///
+    /// Ugly hack with setting the `self_address` storage value.
+    ///
+    /// Stylus SDK doesn't support setting the immutable storage values as
+    /// in Solidity:
+    ///
+    /// ```solidity
+    /// address private immutable __self = address(this);
+    /// ```
```
0xNeshi:
```suggestion
    /// Unlike Solidity's immutable variables, Stylus requires storing the
    /// contract address in a storage field. This additional storage slot
    /// enables the same upgrade safety checks as the Solidity implementation
    /// without affecting the contract's upgrade behavior.
```
Stating that our implementation uses an "ugly hack" won't inspire confidence in devs. The updated description is just as clear and sounds more professional.

examples/uups-proxy/tests/abi/mod.rs
```diff
@@ -0,0 +1,66 @@
...
+        error ERC20InvalidSpender(address spender);
+
+        error OwnableUnauthorizedAccount(address account);
+
+        error UUPSUnauthorizedCallContext();
+        error UUPSUnsupportedProxiableUUID(bytes32 slot);
+        error ERC1967InvalidImplementation(address implementation);
+
+        function upgradeToAndCall(address newImplementation, bytes calldata data) external payable;
+        function upgradeInterfaceVersion() external view returns (string);
```
0xNeshi: Add back the all-capital-letters, see https://github.com/OpenZeppelin/rust-contracts-stylus/pull/752#discussion_r2253253272

examples/uups-proxy/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+    erc20: Erc20,
+    ownable: Ownable,
+    uups: UUPSUpgradeable,
+}
+
+#[public]
+#[implements(IErc20<Error = erc20::Error>, IUUPSUpgradeable, IErc1822Proxiable, IOwnable)]
+impl UUPSProxyErc20Example {
+    #[constructor]
+    #[allow(deprecated)]
```
0xNeshi:
```suggestion
```
redundant?

contracts/src/proxy/utils/uups_upgradeable.rs
```diff
@@ -124,6 +124,7 @@ pub trait IUUPSUpgradeable: IErc1822Proxiable {
     /// # Arguments
     ///
     /// * `&self` - Read access to the contract's state.
+    #[selector(name = "UPGRADE_INTERFACE_VERSION")]
     fn upgrade_interface_version(&self) -> String {
         UPGRADE_INTERFACE_VERSION.into()
     }
```
0xNeshi, line 131: The selector serves no purpose here (no selector inheritance :sadpepe: ), but a [note](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/752#discussion_r2253264371) might come in handy

0xNeshi (approved):
approving in advance

contracts/src/proxy/utils/uups_upgradeable.rs
0xNeshi: Unresolving the comment, as it was neither applied nor addressed

contracts/src/proxy/utils/uups_upgradeable.rs
0xNeshi: I don't think even adding `#[interface_id]` makes sense on this trait. The trait exists only to accommodate Stylus SDKs inheritance requirements, the interface does not exist in Solidity, so there's no need to expose the interface ID

bidzyyys: Removed ✅

PR #753, opened 2025-07-24
fix: export-abi bug for reentrant feature

### LINKED ISSUE #742: [Bug]: `cargo stylus export-abi` not working
### What happened?

This stopped working on https://github.com/OpenZeppelin/rust-contracts-stylus/pull/729

### platform

- [x] linux
- [x] windows
- [x] macos

### Expected behavior

This should work.

Bumping to Stylus SDK v0.10.0 will probably enable more straightforward `StorageSlot` logic that will make it work with `cargo stylus export-abi` (note that contract deployment, which depends on the `stylus-sdk/export-abi` feature, still does work). Before the audited version of the SDK is released however, we may need to find an alternative way to handle this if we want to make it work.

### Contribution Guidelines

bidzyyys:
The problem was for [`reentrant` feature](https://github.com/OffchainLabs/stylus-sdk-rs/blob/v0.9.0/stylus-sdk/Cargo.toml#L53).
Resolves #742 

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog

### DIFF
--- a/.github/workflows/check-abi.yml
+++ b/.github/workflows/check-abi.yml
@@ -15,7 +15,6 @@ jobs:
   check-abi:
     name: Check ABI
     runs-on: ubuntu-latest
-    if: false  # disables this job; set to true or remove to re-enable
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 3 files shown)


0xNeshi (approved):
Great job! :fire:

qalisander (approved):
Good job!

PR #757, opened 2025-07-29
feat: Ed25519 signature (EDDSA)

### LINKED ISSUE #260: [Feature]: Implement Ed25519 signature

Reference implementation: https://github.com/RustCrypto/signatures/tree/master/ed25519

### Contribution Guidelines

qalisander:
Resolves #260

#### PR Checklist

- [ ] Tests
- [x] Documentation
- [ ] Changelog


qalisander:
Address clippy lints

Cargo.toml
```diff
@@ -128,6 +128,7 @@ zeroize = { version = "1.8.1", features = ["derive"] }
 proptest = "1"
 educe = "0.6.0"
 hex-literal = "0.4.1"
+sha2 = "0.10.9"
```
bidzyyys, line 141: I am not a fan of adding additional crypto dependencies.

bidzyyys: How about abstracting it to e.g. use the EVM precompile in a final contract?
https://www.evm.codes/precompiled?fork=cancun#0x02

bidzyyys (commented):
Missing:
- [ ] CHANGELOG.md
- [ ] Ed25519 example
- [ ] Many TODOs
- [ ] clippy warnings

Cargo.toml
qalisander: Agree that abstraction should be better.. We also should change tiny-keccak in our examples to use precompile

contracts/src/token/erc721/mod.rs
```diff
@@ -517,8 +517,8 @@ impl Erc721 {
     /// not tracked by the core [`Erc721`] logic MUST be matched with the use
     /// of [`Self::_increase_balance`] to keep balances consistent with
     /// ownership. The invariant to preserve is that for any address `a` the
-    /// value returned by [`Self::balance_of(a)`] must be equal to the number of
-    /// tokens such that [`Self::_owner_of(token_id)`] is `a`.
+    /// value returned by [`IErc721::balance_of`] must be equal to the number of
```
qalisander: I see we had a problem with cargo docs in main

contracts/src/token/erc721/mod.rs
```diff
@@ -517,8 +517,8 @@ impl Erc721 {
...
+    /// tokens such that [`Self::_owner_of`] is `a`.
```
bidzyyys:
```suggestion
    /// tokens such that `Self::_owner_of(token_id)` is `a`.
```

contracts/src/token/erc721/mod.rs
bidzyyys:
```suggestion
    /// value returned by `IErc721::balance_of(a)` must be equal to the number of
```

bidzyyys: You lost context with the change you applied.

contracts/src/token/erc721/mod.rs
bidzyyys: You lost context with the change you applied.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The `Precompiles::ec_recover` is now callable on `&self`. #754
 - The `ecdsa::recover` function now accepts `impl StaticCallContext` instead of `&mut impl TopLevelStorage`. #754
 - Add bidirectional conversions between `ruint::Uint` and crypto library `Uint` types behind `ruint` feature toggle. #758
+- Add EDDSA signature based on `openzeppelin-crypto`. #757
```
bidzyyys:
```suggestion
- Add EDDSA (Ed25519) signature scheme. #757
```

lib/crypto/src/eddsa.rs
```diff
@@ -0,0 +1,375 @@
...
+
+/// Ed25519 secret key as defined in [RFC8032 § 5.1.5]:
+///
+/// The private key is 32 octets (256 bits, corresponding to b) of
+/// cryptographically secure random data.
+///
+/// [RFC8032 § 5.1.5]: https://www.rfc-editor.org/rfc/rfc8032#section-5.1.5
+pub type SecretKey = [u8; SECRET_KEY_LENGTH];
+
+/// The length of an ed25519 `SecretKey` in bytes.
```
bidzyyys:
```suggestion
/// The length of an ed25519 [`SecretKey`] in bytes.
```

lib/crypto/src/eddsa.rs
```diff
@@ -0,0 +1,375 @@
...
+}
+
+impl SigningKey {
+    /// Get the [`VerifyingKey`] for this [`SigningKey`].
+    #[must_use]
+    pub fn verifying_key(&self) -> VerifyingKey {
+        self.verifying_key
+    }
+
+    /// Construct a [`SigningKey`] from a [`SecretKey`]
```
bidzyyys:
```suggestion
    /// Construct a [`SigningKey`] from a [`SecretKey`].
```

lib/crypto/src/eddsa.rs
```diff
@@ -0,0 +1,375 @@
...
+
+impl ZeroizeOnDrop for ExpandedSecretKey {}
+
+impl ExpandedSecretKey {
+    /// Construct secret key [`Self`] from a byte string of any length.
+    ///
+    /// Secret key will be derived from hashed `bytes`.
+    pub(crate) fn from_bytes(bytes: &[u8]) -> ExpandedSecretKey {
+        let hash = Sha512::default().chain_update(bytes).finalize();
+        let bytes = &*hash;
```
0xNeshi:
```suggestion
        let bytes = hash.as_slice();
```
nit

lib/crypto/src/eddsa.rs
```diff
@@ -0,0 +1,375 @@
...
+}
+
+/// Ed25519 signing key which can be used to produce signatures.
+/// Invariant: `verifying_key` is always the public key of
+/// `secret_key`.
+/// This prevents the signing function [oracle attack].
+///
+/// [oracle attack]: https://github.com/MystenLabs/ed25519-unsafe-libs
+#[derive(Copy, Clone, PartialEq)]
+pub struct SigningKey {
```
0xNeshi:
```suggestion
/// Ed25519 signing key which can be used to produce signatures.
///
/// Invariant: `verifying_key` is always the public key of
/// `secret_key`.
/// This prevents the signing function [oracle attack].
///
/// [oracle attack]: https://github.com/MystenLabs/ed25519-unsafe-libs
#[derive(Copy, Clone, PartialEq)]
pub struct SigningKey {
```
avoids https://rust-lang.github.io/rust-clippy/master/index.html#too_long_first_doc_paragraph

lib/crypto/src/eddsa.rs
```diff
@@ -0,0 +1,375 @@
...
+///
+/// Signature verification libraries are expected to reject invalid
+/// field elements at the time a signature is verified.
+#[derive(Copy, Clone, Eq, PartialEq)]
+pub struct Signature {
+    /// `R` is an `EdwardsPoint`, formed by using an hash function with
+    /// `512-bits` output to produce the digest of:
+    ///
+    /// - the nonce half of the `ExpandedSecretKey`, and
+    /// - the message to be signed.
```
0xNeshi:
```suggestion
    /// * the nonce half of the `ExpandedSecretKey`, and
    /// * the message to be signed.
```
nit: current convention

lib/crypto/src/eddsa.rs
```diff
@@ -0,0 +1,375 @@
...
+    /// element in ℤ/lℤ.  The scalar is then multiplied by the distinguished
+    /// basepoint to produce `R`, and `EdwardsPoint`.
+    pub(crate) R: ProjectivePoint,
+
+    /// `s` is a `Scalar`, formed by using a hash function with `512-bits`
+    /// output to produce the digest of:
+    ///
+    /// - the `r` portion of this `Signature`,
+    /// - the `PublicKey` which should be used to verify this `Signature`, and
+    /// - the message to be signed.
```
0xNeshi:
```suggestion
    /// * the `r` portion of this `Signature`,
    /// * the `PublicKey` which should be used to verify this `Signature`, and
    /// * the message to be signed.
```
nit: update any other dash-list occurrences

lib/crypto/src/eddsa.rs
```diff
@@ -0,0 +1,375 @@
...
+        &self,
+        message: &[u8],
+        signature: &Signature,
+    ) -> bool {
+        let expected_r = self.compute_R(signature, message);
+        expected_r == signature.R
+    }
+
+    /// Helper for verification. Computes the expected R component of the
+    /// signature. The caller compares this to the real R component.
```
0xNeshi:
```suggestion
    /// Helper for verification. Computes the expected `R` component of the
    /// signature. The caller compares this to the real `R` component.
```

lib/crypto/src/eddsa.rs
```diff
@@ -0,0 +1,375 @@
...
+        assert!(signing_key.is_valid_signature(message, &signature));
+
+        // Verify with a different message
+        let invalid_message = b"I'm not signed!";
+        assert!(!signing_key.is_valid_signature(invalid_message, &signature));
+    }
+
+    #[test]
+    fn sign_and_verify() {
+        proptest!(|(secret_key: SecretKey, message: String)| {
```
0xNeshi: are there any low-hanging-fruit properties that we can proptest?

lib/crypto/src/field/fp.rs
```diff
@@ -474,6 +475,26 @@ impl<P: FpParams<N>, const N: usize> Fp<P, N> {
         }
         Uint::new(limbs)
     }
+
+    /// Construct `Self` from the other [`Fp`] element of different size.
+    ///
+    /// # Panics
+    ///
+    /// * If `value` is bigger than `Self` maximum size.
+    pub(crate) fn from_fp<P2: FpParams<N2>, const N2: usize>(
```
0xNeshi, line 484: is there any sense in adding `try_from_fp`?

qalisander: I think we should address it within this [issue](https://github.com/OpenZeppelin/rust-contracts-stylus/issues/769). All of it

lib/crypto/src/eddsa.rs
qalisander: Idk what else.. We don't have a rich api for signature

qalisander:
@bidzyyys 
* "Ed25519 Example Contract" the plan was to add it later altogether with benchmarks. But we can included within current changes ofcourse.
* "Unit Test with Known Vectors" for this thing would be better to implement proper serialization for all crypto primitives. That will take a bit more time

lib/crypto/src/eddsa.rs
0xNeshi: we can leave this for when we work on code coverage

0xNeshi:
About unit test vectors, see: https://datatracker.ietf.org/doc/html/rfc8032#section-7

0xNeshi (approved):
Excellent work

examples/eddsa/src/lib.rs
```diff
@@ -0,0 +1,27 @@
...
+
+#[entrypoint]
+#[storage]
+struct EddsaExample;
+
+#[public]
+impl EddsaExample {
+    fn sign(
+        &mut self,
+        secret_key: alloy_primitives::U256,
```
qalisander: This example is a bit out of the world: We usually don't need to send any secrets on-chain. But at the moment we will test it like that, `VerificationKey` conversion from bytes is not implemented

qalisander: Added issues #775 #776

0xNeshi: > This example is a bit out of the world: We usually don't need to send any secrets on-chain.

Yeah, this is kind of weird. If there's no better example, what would even be another use case for oz-crypto::eddsa?

0xNeshi (approved):
Good unit tests. Approving in advance, but left a question

examples/eddsa/src/lib.rs
qalisander: The best and the most used is verification. But we need a bit more "ground" to implement it smoothly. At least it checks that WASM size is fine enough for contract to be deployed

0xNeshi: Maybe would be a good idea to add a comment above the example contract stating that we DO NOT recommend devs use the library like this, and that this is just an example contract we created to:
1. Show how it should be configured to use EDDSA
2. Check that it's deployable

Since this is an example for devs, the contract should contain a "verify signature" function, to show what it should look like; no need to create a test for it though.

lib/crypto/src/field/fp.rs
bidzyyys: Yeah, the issue should be solved in potential M4 💯

lib/crypto/src/eddsa.rs
```diff
@@ -0,0 +1,548 @@
...
+                secret_key: &hex!($secret_key),
+                expected_public_key: &hex!($public_key),
+                message: &hex!($message),
+                expected_signature: &hex!($signature),
+            }
+        };
+    }
+
+    #[test]
+    fn rfc8032_known_signatures() {
```
bidzyyys, line 456: I love it 🚀

examples/eddsa/src/lib.rs
bidzyyys: In my opinion, the example that demonstrates signing and sending a secret key on-chain should not be included in our examples - it is against security best practises. I'd suggest we provide an example showing how to build an EdDSA smart contract with a verify function, but omit the E2E tests for this particular case. The unit tests you've added already demonstrate that your solution is consistent with the specifications.

0xNeshi: Summarizing internal discussion based on @bidzyyys's [suggestion](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/757#discussion_r2259589013):

We should add a single e2e test, that sings a message and sends it for verification to the example contract.
Remove the existing signing function from the contract and the related test.

PR #758, opened 2025-07-29
feat: implement conversions ruint::Uint <-> openzeppelin_crypto::Uint

### LINKED ISSUE #513: [Feature]: Implement RUint conversions from and to Uint

Add `ruint` feature for conversions to [ruint](https://crates.io/crates/ruint)

### Contribution Guidelines

bidzyyys:
Resolves #513 

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -114,6 +114,7 @@ alloy-sol-types = { version = "=0.8.20", default-features = false }
 alloy-sol-macro = { version = "=0.8.20", default-features = false }
 alloy-sol-macro-expander = { version = "=0.8.20", default-features = false }
 alloy-sol-macro-input = { version = "=0.8.20", default-features = false }
+ruint = { version = "=1.12.3", default-features = false }
 
 eyre = "0.6.8"
 keccak-const = "0.2.0"
--- a/lib/crypto/src/arithmetic/uint.rs
+++ b/lib/crypto/src/arithmetic/uint.rs
@@ -523,6 +523,75 @@ impl_from_primitive!(u64, from_u64);
 impl_from_primitive!(usize, from_usize);
 impl_from_primitive!(u128, from_u128);
 
+/// This macro implements bidirectional conversions between [`ruint::Uint`] and
+/// [`Uint`] types.
+///
+/// For each bit size, it creates:
+/// - `From<ruint::Uint<BITS, LIMBS>>` for the corresponding `Uint` type
+/// - `From<Uint>` for the corresponding `ruint::Uint<BITS, LIMBS>` type
+///
+/// The number of limbs is automatically calculated using `usize::div_ceil(bits,
+/// Limb::BITS)`.
+macro_rules! impl_from_ruint {
+    ($num:ident, $bits:expr) => {
+        impl
+            From<
+                ruint::Uint<
+                    $bits,
+                    {
+                        usize::div_ceil(
+                            $bits,
+                            $crate::arithmetic::Limb::BITS as usize,
+                        )
+                    },
+                >,
+            > for $num
+        {
+            fn from(
+                val: ruint::Uint<
+                    $bits,
+                    {
+                        usize::div_ceil(
+                            $bits,
+                            $crate::arithmetic::Limb::BITS as usize,
+                        )
+                    },
+                >,
+            ) -> Self {
+                $num::from_bytes_le(&val.to_le_bytes_vec())
+            }
+        }
+
+        impl From<$num>
+            for ruint::Uint<
+                $bits,
+                {
+                    usize::div_ceil(
+                        $bits,
+                        $crate::arithmetic::Limb::BITS as usize,
+                    )
+                },
+            >
+        {
+            fn from(val: $num) -> Self {
+                ruint::Uint::<
+                    $bits,
+                    {
+                        usize::div_ceil(
+                            $bits,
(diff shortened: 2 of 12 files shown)


lib/crypto/src/arithmetic/uint.rs
```diff
@@ -523,6 +523,75 @@ impl_from_primitive!(u64, from_u64);
...
+                        usize::div_ceil(
+                            $bits,
+                            $crate::arithmetic::Limb::BITS as usize,
+                        )
+                    },
+                >,
+            ) -> Self {
+                $num::from_bytes_le(&val.to_le_bytes_vec())
+            }
+        }
```
0xNeshi: Given how these conversions were previously [ad-hoc implemented](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/758/files#diff-671c0df3badd9ed03acaf67fd8a6c0b153146a54e02eb9cb041d2dbe8db65265L18-L20):
```rust
fn to_alloy_u256(value: &U256) -> alloy_primitives::U256 {
    alloy_primitives::U256::from_le_slice(&value.into_bytes_le())
}
```
couldn't we have referenced `alloy_primitives::Uint` (which is a re-export of `ruint::Uint`) instead of explicitly adding the `ruint` dependency?

0xNeshi (commented):
Amazing :fire:

lib/crypto/src/arithmetic/uint.rs
bidzyyys: IMHO we do not want to have `alloy_primitives` as a dependency for `openzeppelin-crypto` -- we want to keep the size as minimum as possible. But curious to hear @qalisander opinion.

0xNeshi: Ah I see now that `alloy_primitives` is just a dev. dependency. In that case, you can ignore my comment :+1:

Cargo.toml
```diff
@@ -115,6 +115,8 @@ alloy-sol-macro = { version = "=0.8.20", default-features = false }
 alloy-sol-macro-expander = { version = "=0.8.20", default-features = false }
 alloy-sol-macro-input = { version = "=0.8.20", default-features = false }
 
+ruint = "=1.14.0"
```
0xNeshi: Why the version bump? 

Since `alloy-primitives@0.8.20` uses `ruint@1.12.3` (see [source](https://github.com/alloy-rs/core/blob/v0.8.20/Cargo.toml#L100)), it definitely should support this updated version, but aren't we then forcing lib users to depend on this updated version unnecessarily? Maybe they want 1.12.3..

It might be better to set this version to `ruint = "1.12.3"` (note there's no `=`), which would make this the _minimal supported version_, allowing devs to use whichever one they wish and reducing chances of crate version conflicts.

bidzyyys: @0xNeshi when you check the current [Cargo.lock](https://github.com/OpenZeppelin/rust-contracts-stylus/blob/main/Cargo.lock#L3550-L3575) we already use the “bumped” version. Secondly, when you check CI jobs, before "bump" there was an increase in WASM sizes that was causing CI job check fails for `erc721-metadata-example` as you can see [here](https://github.com/OpenZeppelin/rust-contracts-stylus/actions/runs/16593553479/job/46935945845) ; this is how WASM sizes get bigger: [link](https://github.com/OpenZeppelin/rust-contracts-stylus/actions/runs/16593553478/job/46934774143).

0xNeshi: Both the [failing](https://github.com/OpenZeppelin/rust-contracts-stylus/actions/runs/16593553479/job/46935945845#step:6:46) and the [succeeding](https://github.com/OpenZeppelin/rust-contracts-stylus/actions/runs/16593553478/job/46934774143#step:5:45) builds used the same `ruint@1.12.3` version :eyes:

---

Even though we specified a lower minimal version, cargo did use the bumped version because cargo prefers the highest version available.
But using minimal versions allows devs to use our library with a wider range of dependency versions on their side. Currently, all devs are forced to use ruint@1.14.0 exactly, otherwise their builds fail; to make matters worse, they can't even use ruint@1.14.1 . If their other deps depend on other ruint versions, they may have problems. In the end, we can never know whether some dev has legitimate reasons why they do not wish to use the exact dep. version we specified and they'd like more flexibility.

Given we're creating an open source crate for other devs to use, it would make sense to always set minimal supporter versions for dependencies and ensuring that our lib compiles for those minimal versions, which can be achieved with the -Z direct-minimal-versions flag - we should maybe look into this after the audit :point_left:

0xNeshi (approved):
We can table the dep. version discussion for another time, just left [one comment](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/758#discussion_r2242371554) to address

Cargo.toml
0xNeshi: @bidzyyys I would just remove `=` from the dependency declaration, as leaving it there would break developers' builds if they tried using e.g. `1.14.1`

0xNeshi: Tracking issue https://github.com/OpenZeppelin/rust-contracts-stylus/issues/760

qalisander (approved):
LGTM!

Added `ruint` feature toggle and generic `From` conversions and updated changelog

PR #764, opened 2025-07-31
feat: conversion between Uint and primitive integers

### LINKED ISSUE #517: [Feature]: Add conversion openzeppelin_crypto::Uint -> u8..u64

Make sense to write macro to simplify boilerplate

### Contribution Guidelines

qalisander:
Resolves #517

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog


bidzyyys (commented):
Missing:
- [x] unit tests
- [ ] CHANGELOG

lib/crypto/src/arithmetic/uint.rs
```diff
@@ -523,7 +523,70 @@ impl_from_primitive!(u64, from_u64);
...
+            #[allow(clippy::cast_possible_truncation)]
+            pub const fn $func_name(self) -> $int {
+                assert!(N >= 1, "number of limbs must be greater than zero");
+                ct_for!((i in 1..N) {
+                    assert!(self.limbs[i] == 0, "Uint type is to large to fit");
+                });
+                assert!(
+                    self.limbs[0] <= <$int>::MAX as Limb,
+                    "Uint type is to large to fit"
+                );
```
0xNeshi, line 551: Isn't this guaranteed to never fail? The macro is not public, and you've used it with unsigned ints up to `u64`.

lib/crypto/src/arithmetic/uint.rs
```diff
@@ -523,7 +523,70 @@ impl_from_primitive!(u64, from_u64);
...
+macro_rules! impl_ct_into_primitive {
+    ($int:ty, $func_name:ident) => {
+        impl<const N: usize> Uint<N> {
+            #[doc = "Create a"]
+            #[doc = stringify!($int)]
+            #[doc = "integer from [`Uint`] (constant)."]
+            #[must_use]
+            #[allow(clippy::cast_possible_truncation)]
+            pub const fn $func_name(self) -> $int {
+                assert!(N >= 1, "number of limbs must be greater than zero");
```
0xNeshi, line 541: Isn't this guaranteed to never fail? Can there ever be custom `Uint<0>`?

lib/crypto/src/arithmetic/uint.rs
```diff
@@ -523,7 +523,70 @@ impl_from_primitive!(u64, from_u64);
 impl_from_primitive!(usize, from_usize);
 impl_from_primitive!(u128, from_u128);
 
-// ----------- Traits Impls -----------
+/// Constant conversions into primitive types.
+macro_rules! impl_ct_into_primitive {
```
0xNeshi, line 530: Isn't this macro redundant? Its output functions are only used in the `impl_from_uint` macro and nowhere else in the code (so the #[doc] output is unnecessary), it could be merged with  the `impl_from_uint` macro directly.

Here's a shortened implementation:

```rust
// ----------- From traits implementation for [`Uint`] into primitive types -----------

macro_rules! impl_from_uint {
    ($int:ty) => {
        impl<const N: usize> From<Uint<N>> for $int {
            #[inline]
            fn from(val: Uint<N>) -> $int {
                assert!(N >= 1, "number of limbs must be greater than zero");
                ct_for!((i in 1..N) {
                    assert!(val.limbs[i] == 0, "Uint type is too large to fit");
                });
                assert!(
                    val.limbs[0] <= <$int>::MAX as Limb,
                    "Uint type is to large to fit"
                );

                self.limbs[0] as $int
            }
        }
    };
}

impl_from_uint!(u8);
impl_from_uint!(u16);
impl_from_uint!(u32);
impl_from_uint!(u64);
impl_from_uint!(usize);

impl<const N: usize> From<Uint<N>> for u128 {
    #[inline]
    fn from(val: Uint<N>) -> u128 {
        assert!(N >= 1, "number of limbs must be greater than zero");
        ct_for!((i in 2..N) {
            assert!(val.limbs[i] == 0, "Uint type is too large to fit");
        });

        let res0 = self.limbs[0] as u128;
        let res1 = (self.limbs[1] as u128) << 64;
        res0 | res1
    }
}
```

lib/crypto/src/arithmetic/uint.rs
```diff
@@ -1291,4 +1358,23 @@ mod test {
...
+                    assert_eq!(expected_primitive_num, primitive_num);
+                });
+            };
+        }
+
+        test_uint_conversion!(u8);
+        test_uint_conversion!(u16);
+        test_uint_conversion!(u32);
+        test_uint_conversion!(u64);
+        test_uint_conversion!(u128);
```
0xNeshi:
```suggestion
        test_uint_conversion!(usize);
        test_uint_conversion!(u128);
```

lib/crypto/src/const_helpers.rs
```diff
@@ -10,6 +10,13 @@
 //! [loop unroll]: https://en.wikipedia.org/wiki/Loop_unrolling
 
 /// Allows writing `for` cycle in constant context.
+///
+/// ```rust,ignore
+/// // This loop will start from 0 till 10 exclusive.
+/// ct_for!((i in 0..10) {
+///
+/// });
+/// ```
```
0xNeshi:
```suggestion
/// Allows writing `for` loops in constant context.
///
/// ```rust,ignore
/// // This loop runs from 0 to 10 (not including 10).
/// ct_for!((i in 0..10) {
///     // loop body here
/// });
/// ```
```
nit: clearer wording

CHANGELOG.md
```diff
@@ -29,7 +30,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Change `Erc1155Receiver` constants `SINGLE_TRANSFER_FN_SELECTOR` and `BATCH_TRANSFER_FN_SELECTOR` to type `B32`. #747
 - Change `Erc721Receiver` constant `RECEIVER_FN_SELECTOR` to type `B32`. #747
 - Rename `Precompiles::ecrecover` wrapper function to `Precompiles::ec_recover`. #754
-- Simplify Pedersen hash API to accept any type that implements `Into<P::BaseField>`. #758
```
0xNeshi, line 34: This def. isn't a breaking change, but shouldn't it still be noted in `### Changed`?

qalisander: Thnx! I think it is due to bad merge conflict resolution

lib/crypto/src/arithmetic/uint.rs
qalisander: We're casting `Uint` type integer. Think of casting `Uint<1>` ([u64;1]) to `u16`. So there is a risk that overflow will occur.

lib/crypto/src/arithmetic/uint.rs
qalisander: But `Uint::<_>::from(1_u8)` will never be constant. Since trait methods cannot be constant methods

lib/crypto/src/arithmetic/uint.rs
qalisander: Good point. We don't have any application for `Uint<0>` in our library. May be better idea would be prohibit compilation of `Uint<0>` with constant assertions.
But at the moment I'm trying to reject this type early in most functions. Wdt?

contracts/src/token/erc721/mod.rs
```diff
@@ -517,8 +517,8 @@ impl Erc721 {
     /// not tracked by the core [`Erc721`] logic MUST be matched with the use
     /// of [`Self::_increase_balance`] to keep balances consistent with
     /// ownership. The invariant to preserve is that for any address `a` the
-    /// value returned by [`Self::balance_of(a)`] must be equal to the number of
-    /// tokens such that [`Self::_owner_of(token_id)`] is `a`.
+    /// value returned by [`IErc721::balance_of`] must be equal to the number of
```
bidzyyys:
```suggestion
    /// value returned by `Self::balance_of(a)` must be equal to the number of
```

contracts/src/token/erc721/mod.rs
```diff
@@ -517,8 +517,8 @@ impl Erc721 {
...
+    /// tokens such that [`Self::_owner_of`] is `a`.
```
bidzyyys:
```suggestion
    /// tokens such that `Self::_owner_of(token_id)` is `a`.
```

lib/crypto/src/arithmetic/uint.rs
bidzyyys: > But at the moment I'm trying to reject this type early in most functions. Wdt?

Sounds ok for me.

lib/crypto/src/arithmetic/uint.rs
0xNeshi: Hmm, then in addition to `From` impls, wouldn't it make sense to also provide `TryFrom`? Maybe some devs want to be able to handle the error instead of reverting the tx.
Shouldn't be too big of a change.

lib/crypto/src/arithmetic/uint.rs
```diff
@@ -1291,4 +1367,24 @@ mod test {
     test_ruint_conversion!(test_ruint_conversion_u64, U64, 64);
     test_ruint_conversion!(test_ruint_conversion_u128, U128, 128);
     test_ruint_conversion!(test_ruint_conversion_u256, U256, 256);
+
+    #[test]
+    fn uint_and_primitives_conversion() {
+        macro_rules! test_uint_conversion {
+            ($type:ty) => {
+                proptest!(|(expected_primitive_num: $type)| {
```
0xNeshi: Each uint type should have its own separate proptest case (see `test_ruint_conversion`)

lib/crypto/src/arithmetic/uint.rs
```diff
@@ -1291,4 +1367,24 @@ mod test {
     test_ruint_conversion!(test_ruint_conversion_u64, U64, 64);
     test_ruint_conversion!(test_ruint_conversion_u128, U128, 128);
```
0xNeshi: Unrelated to this PR, but still important to address:
1. **These tests are never run.** 
**Solution:** Our CI needs to be updated to run `cargo nextest` with `--features ruint` enabled.
2. **These two tests actually do not compile when `--features ruint` is enabled.** That's because `U64` and `U128` are never imported in tests.
**Solution:** Update the macro definition to locally import the types.
[This](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/764/files#diff-463d4484afcd15f2bd182d700a3a87aa0f71df2dc27652948308d0e99f384ec6R1359):
```rust
                    let uint_from_ruint: $uint_type = value.into();
```
Becomes this:
```rust
                    let uint_from_ruint: alloy_primitives::$uint_type = value.into();
```

lib/crypto/src/arithmetic/uint.rs
```diff
@@ -1291,4 +1367,24 @@ mod test {
...
+        }
+
+        test_uint_conversion!(u8);
+        test_uint_conversion!(u16);
+        test_uint_conversion!(u32);
+        test_uint_conversion!(u64);
+        test_uint_conversion!(u128);
+        test_uint_conversion!(usize);
+    }
 }
```
0xNeshi, line 1404: Can you fix this CI warning?
https://github.com/OpenZeppelin/rust-contracts-stylus/actions/runs/16719357718/job/47319998815?pr=764#step:6:19

lib/crypto/src/arithmetic/uint.rs
0xNeshi: I see that we have constant conversion for `impl_ct_from_primitive`.

Why do you think we need constant conversions at all?

qalisander: We actually don't need many things in the pr right now. We're not using conversions to specific specific types (ike u8). It is usability feature, to be able to convert integer in `const` functions. Otherwise it won't work.

lib/crypto/src/arithmetic/uint.rs
qalisander: We actually should prefer `TryFrom` only for fallible conversions. Also some of the conversions are assumed to never fail: like `u8 -> Uint<1>`. And there we should probably implement also `From`. But it will complicate design, since we should add errors and error handling. It will also complicate design in `const` functions, but is a good way to go. But IMHO not a priority at the moment, since we're using those conversions in `const` functions and tests mostly.

lib/crypto/src/arithmetic/uint.rs
0xNeshi: Ok then, leave it as-is

lib/crypto/src/arithmetic/uint.rs
0xNeshi: Can you create an issue to track the `TryFrom` impls and we can resolve this?

lib/crypto/src/arithmetic/uint.rs
bidzyyys: It is fixed in #749

lib/crypto/src/arithmetic/uint.rs
qalisander: Why so?

lib/crypto/src/arithmetic/uint.rs
qalisander: Good point! Thought we have `--all-features` enabled for tests

lib/crypto/src/arithmetic/uint.rs
qalisander: Consider we don't need to have a test case name as a parameter with my approach. And some other test [suites](https://github.com/OpenZeppelin/rust-contracts-stylus/blob/7d5c72d6ca663ef96ef33811b5e6711073cb3000/lib/crypto/src/bits.rs#L68) at crypto library following the same pattern. It is also more convenient to run test cases in ide's, since `#[test]` attribute outside of macro.

lib/crypto/src/arithmetic/uint.rs
qalisander: [Here](https://github.com/OpenZeppelin/rust-contracts-stylus/issues/769)

lib/crypto/src/arithmetic/uint.rs
0xNeshi: It's just a general convention, and since we use it in the project, I immediately noticed this being inconsistent.
Grouping proptests like this has many pitfalls:
- If any particular proptest fails, it will be impossible to rerun just that proptest (you have to rerun all of them).
- If any particular proptest fails, _all of them fail_. Instead of it being apparent that e.g. `u128` is failing, we won't be sure that other uint type tests are passing until we run them separately.
- Instead of running in parallel, these proptests will run sequentially.

qalisander: It will be still a bit convoluted to me rerunning a specific test behind macro:)
I feel we can spend more time on arguments, than reap actual benefits..
Nevertheless @bidzyyys do you want also change it everywhere? (to hide the whole test case behind macro)

0xNeshi (commented):
https://github.com/OpenZeppelin/rust-contracts-stylus/pull/764#discussion_r2256371403

lib/crypto/src/arithmetic/uint.rs
qalisander: > If any particular proptest fails, all of them fail. Instead of it being apparent that e.g. u128 is failing, we won't be sure that other uint type tests are passing until we run them separately.

Agree here, but not sure it will help to save much time on debugging

> Instead of running in parallel, these proptests will run sequentially

And agree here, but parallel execution won't help when you already have many test cases. e.g. that exact fat test case runs for 40ms compare to the whole test suite running 2 sec

0xNeshi: > It will be still a bit convoluted to me rerunning a specific test behind macro:)

Why? The [test_ruint_conversion](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/764/files#diff-463d4484afcd15f2bd182d700a3a87aa0f71df2dc27652948308d0e99f384ec6R1363) macro definition is pretty straightforward, you can take it as inspiration.

0xNeshi: > but not sure it will help to save much time on debugging

How will it _not_ save time? No need to comment out other tests, rerun them separately to confirm the others are actually working - literally what the whole discussion is about - separate the tests preemptively :)

> And agree here, but parallel execution won't help when you already have many test cases. e.g. that exact fat test case runs for 40ms compare to the whole test suite running 2 sec

Although the parallel execution is undeniably still a reason _for_ separating tests, by itself it wouldn't be that important. Still, it's just one small point besides 2 more important ones.

0xNeshi (commented):
https://github.com/OpenZeppelin/rust-contracts-stylus/pull/764#discussion_r2256760688

lib/crypto/src/arithmetic/uint.rs
qalisander: > Why? The [test_ruint_conversion](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/764/files#diff-463d4484afcd15f2bd182d700a3a87aa0f71df2dc27652948308d0e99f384ec6R1363) macro definition is pretty straightforward, you can take it as inspiration.

Only using cli instead of IDE's shortcut. Test cases behind macro are not recognized by IDEs as an individual tests.

qalisander: So it is actually quite simple workflow for me. Comment working test cases (few clicks). Setup failing case with debugger to see (ide shortcut: debug <<test case>>) which conditions are wrong.
Otherwise debugging with `dbg!(..)` macro only.

bidzyyys:
> [#764 (comment)](https://github.com/OpenZeppelin/rust-contracts-stylus/pull/764#discussion_r2256760688)

Agree with @0xNeshi.

lib/crypto/src/arithmetic/uint.rs
qalisander: I would be up to this approach:
image

Or include that kind of test cases into separate module, like I did with `ruint` in this pr.

Wdt @0xNeshi ?

0xNeshi: > I would be up to this approach:

Sure, that might be a good compromise solution.

As a side note, these tests only verify `uint <-> U256` conversion, but it would be useful to have other `Uxxx` conversions as well to verify behavior. Maybe do this as part of a separate PR, I wouldn't hold this PR for it.

0xNeshi: @bidzyyys wdyt?

bidzyyys: > @bidzyyys wdyt?

Looks good to me 👍

0xNeshi (approved):

PR #796, opened 2025-08-14
fix: storage slot compilation with stylus-sdk/stylus-test feature

### LINKED ISSUE #793: [Bug]: Ensure Tests Compile when `stylus-sdk/stylus-test` Feature is Enabled
### What happened?

When this feature is enabled on `stylus-sdk`, the tests cannot be built due to:
```bash
error[E0423]: expected function, tuple struct or tuple variant, found struct VM
   --> /Users/ayushboss/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/openzeppelin-stylus-0.3.0-rc.1/src/utils/storage_slot.rs:67:20
    |
67  |         let host = VM(stylus_sdk::host::WasmVM {});
    |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
   ::: /Users/ayushboss/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/stylus-sdk-0.9.0/src/host/mod.rs:329:9
    |
329 |         pub struct VM {
    |         ------------- VM defined here
    |
help: use struct literal syntax instead of calling
    |
67  -         let host = VM(stylus_sdk::host::WasmVM {});
67  +         let host = VM{host: stylus_sdk::host::WasmVM {}};
    |

For more information about this error, try rustc --explain E0423.
error: could not compile openzeppelin-stylus (lib) due to 1 previous error
```

Possibly related https://github.com/OpenZeppelin/rust-contracts-stylus/issues/730

### platform

- [x] linux
- [x] windows
- [x] macos

### Expected behavior

Unit tests should build with the `stylus-sdk/stylus-test` feature enabled.

### Contribution Guidelines

sumitvekariya:
## Description
Fixes #793 

This PR resolves a compilation error that occurs when the `stylus-sdk/stylus-test` feature is enabled. The issue was that the `VM` struct changes from a tuple struct to a regular struct when this feature is active, but the conditional compilation logic didn't account for this case.

## Changes Made
- Updated conditional compilation in `contracts/src/utils/storage_slot.rs`
- Added logic to detect when struct literal syntax should be used vs tuple struct syntax
- Enhanced comments to explain the different compilation paths

## Testing
- ✅ Tests pass with `stylus-sdk/stylus-test` feature enabled
- ✅ Tests pass without the feature (normal case)
- ✅ Specific `test_storage_slot` test works in both configurations

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,7 +34,7 @@ jobs:
         # version of the rust toolchain.
         toolchain: [stable, beta]
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
         with:
           submodules: true
 
@@ -73,7 +73,7 @@ jobs:
         # Windows fails because of `stylus-proc`.
         # os: [macos-latest, windows-latest]
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
         with:
           submodules: true
 
@@ -120,7 +120,7 @@ jobs:
     runs-on: ubuntu-latest
     name: ubuntu / nightly / coverage
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
         with:
           submodules: true
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -64,7 +64,7 @@ jobs:
 
         # https://github.com/rust-lang/cargo/issues/6669
       - name: Run doc tests
-        run: cargo test --locked --doc -p openzeppelin-stylus --features stylus-test
+        run: cargo test --locked --doc -p openzeppelin-stylus -p openzeppelin-stylus-proc -p openzeppelin-crypto --features stylus-test
   os-check:
     # Run cargo test on MacOS and Windows.
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -57,14 +57,14 @@ jobs:
 
         # https://twitter.com/jonhoo/status/1571290371124260865
       - name: Run unit tests
-        run: cargo nextest run --locked --all-targets -p openzeppelin-stylus -p openzeppelin-stylus-proc -p openzeppelin-crypto --features ruint
+        run: cargo nextest run --locked --all-targets -p openzeppelin-stylus -p openzeppelin-stylus-proc -p openzeppelin-crypto --features ruint,stylus-test
 
       - name: Run repro tests (repros/*)
         run: ./scripts/test-repros.sh
 
         # https://github.com/rust-lang/cargo/issues/6669
       - name: Run doc tests
-        run: cargo test --locked --doc
+        run: cargo test --locked --doc -p openzeppelin-stylus --features stylus-test
   os-check:
     # Run cargo test on MacOS and Windows.
     runs-on: ${{ matrix.os }}
@@ -96,7 +96,7 @@ jobs:
         run: cargo generate-lockfile
 
       - name: Run unit tests
-        run: cargo nextest run --locked --all-targets -p openzeppelin-stylus -p openzeppelin-stylus-proc -p openzeppelin-crypto --features ruint
+        run: cargo nextest run --locked --all-targets -p openzeppelin-stylus -p openzeppelin-stylus-proc -p openzeppelin-crypto --features ruint,stylus-test
 
   coverage:
     # Use llvm-cov to build and collect coverage and outputs in a format that
@@ -156,7 +156,7 @@ jobs:
         run: echo "RUST=$(rustc --version)" >> "$GITHUB_ENV"
 
       - name: Prepare report for Codecov
-        run: cargo nextest run --profile ci --locked --all-targets -p openzeppelin-stylus -p openzeppelin-stylus-proc -p openzeppelin-crypto --features ruint
+        run: cargo nextest run --profile ci --locked --all-targets -p openzeppelin-stylus -p openzeppelin-stylus-proc -p openzeppelin-crypto --features ruint,stylus-test
 
       - name: Upload test results to Codecov
         if: ${{ !cancelled() }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -59,6 +59,9 @@ jobs:
       - name: Run unit tests
         run: cargo nextest run --locked --all-targets -p openzeppelin-stylus -p openzeppelin-stylus-proc -p openzeppelin-crypto --features ruint
 
+      - name: Run repro tests (repros/*)
+        run: ./scripts/test-repros.sh
+
         # https://github.com/rust-lang/cargo/issues/6669
       - name: Run doc tests
         run: cargo test --locked --doc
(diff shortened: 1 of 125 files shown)


contracts/src/utils/storage_slot.rs
```diff
@@ -64,9 +64,19 @@ impl StorageSlot {
...
+            // the correct syntax based on compilation context.
+            #[cfg(any(test, debug_assertions))]
+            {
+                VM { host: alloc::boxed::Box::new(stylus_sdk::host::WasmVM {}) }
+            }
+            #[cfg(not(any(test, debug_assertions)))]
+            {
+                VM(stylus_sdk::host::WasmVM {})
+            }
+        };        // SAFETY: Truncation is safe here because ST::SLOT_BYTES is never
```
0xNeshi:
```suggestion
        };

        // SAFETY: Truncation is safe here because ST::SLOT_BYTES is never
```

contracts/src/utils/storage_slot.rs
```diff
@@ -64,9 +64,19 @@ impl StorageSlot {
             target_arch = "wasm32",
             all(not(test), not(feature = "reentrant"))
         ))]
-        let host = VM(stylus_sdk::host::WasmVM {});
-
-        // SAFETY: Truncation is safe here because ST::SLOT_BYTES is never
+        let host = {
```
0xNeshi: Can we simplify the whole `cfg` logic? Seems to have too many branches.

```rust
        #[cfg(all(
            not(target_arch = "wasm32"),
            any(test, feature = "reentrant", debug_assertions)
        ))]
        let host =
            VM { host: alloc::boxed::Box::new(stylus_sdk::host::WasmVM {}) };
        #[cfg(any(
            target_arch = "wasm32",
            not(any(test, feature = "reentrant", debug_assertions))
        ))]
        let host = VM(stylus_sdk::host::WasmVM {});
```

EDIT: updated the suggestion so that it fixes the [failing `cargo package` CI](https://github.com/OpenZeppelin/rust-contracts-stylus/actions/runs/16955836792/job/48068046318?pr=796).

contracts/src/utils/storage_slot.rs
```diff
@@ -64,9 +64,19 @@ impl StorageSlot {
...
+            // When stylus-sdk/stylus-test feature is enabled, VM is a regular struct.
+            // We can't directly detect this feature, but we can try to determine
+            // the correct syntax based on compilation context.
```
0xNeshi: Move this comment above the `let host` declaration, so that it reads:
```rust
        // When stylus-sdk/stylus-test feature is enabled, VM is a regular struct.
        // We can't directly detect this feature, but we can try to determine
        // the correct syntax based on compilation context.
        //
        // TODO: Remove this once we have a proper way to inject the host for
        // custom storage slot access.
        // This has been implemented on Stylus SDK 0.10.0.
```

0xNeshi (commented):
Awesome work, seems to fix the issue :fire:

To ensure this type of bug doesn't happen again, let's create a CI job verifying the fix still holds.

I created a branch against your PR, please approve and merge: https://github.com/sumitvekariya/rust-contracts-stylus/pull/1

contracts/src/utils/storage_slot.rs
0xNeshi: Unfortunately, the solution still causes a failure when [generating ABI](https://github.com/OpenZeppelin/rust-contracts-stylus/actions/runs/16955836781/job/48068046260?pr=796).

Could you try the following approach: declare a feature in `./contracts/Cargo.toml`:
```toml
[features]
reentrant = ["stylus-sdk/reentrant"]
export-abi = ["stylus-sdk/export-abi"]
stylus-test = ["stylus-sdk/stylus-test"]
```

and try to set the `host` instantiation based on whether this feature is enabled? 

This would also require:
- updating all examples to enable the `openzeppelin-stylus/stylus-test` feature in their `Cargo.toml > [dev-dependencies]`.
- updating all CI scripts that run tests to enable the `openzeppelin-stylus/stylus-test` feature.

.github/workflows/test.yml
```diff
@@ -57,11 +57,14 @@ jobs:
...
-        run: cargo nextest run --locked --all-targets -p openzeppelin-stylus -p openzeppelin-stylus-proc -p openzeppelin-crypto --features ruint
+        run: cargo nextest run --locked --all-targets -p openzeppelin-stylus -p openzeppelin-stylus-proc -p openzeppelin-crypto --features ruint,stylus-test
+
+      - name: Run repro tests (repros/*)
+        run: ./scripts/test-repros.sh
 
         # https://github.com/rust-lang/cargo/issues/6669
       - name: Run doc tests
-        run: cargo test --locked --doc
+        run: cargo test --locked --doc -p openzeppelin-stylus --features stylus-test
```
0xNeshi: Why focus only on `-p openzeppelin-stylus`? If it's because of the feature, ensure that the same doc test command is run against other crates in the project

.github/workflows/test.yml
```diff
@@ -57,11 +57,14 @@ jobs:
 
         # https://twitter.com/jonhoo/status/1571290371124260865
       - name: Run unit tests
-        run: cargo nextest run --locked --all-targets -p openzeppelin-stylus -p openzeppelin-stylus-proc -p openzeppelin-crypto --features ruint
+        run: cargo nextest run --locked --all-targets -p openzeppelin-stylus -p openzeppelin-stylus-proc -p openzeppelin-crypto --features ruint,stylus-test
```
0xNeshi, line 60: Let's see if this works, running the CI job :crossed_fingers:

contracts/src/utils/storage_slot.rs
```diff
@@ -51,18 +51,22 @@ impl StorageSlot {
...
+        //
         // TODO: Remove this once we have a proper way to inject the host for
         // custom storage slot access.
         // This has been implemented on Stylus SDK 0.10.0.
         #[cfg(all(
             not(target_arch = "wasm32"),
-            any(test, feature = "reentrant")
+            any(test, feature = "reentrant", feature = "stylus-test")
         ))]
         let host =
```
0xNeshi, line 65: Does this work now just with `cfg(feature = "stylus-test")`?
If not, could we simplify this `cfg` further?

sumitvekariya: tried that approach but it doesn't work in all contexts. The [debug_assertions](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) approach is more reliable and works consistently across debug/release modes. please share the feedback on this

sumitvekariya:
> * cargo hac

addresed both the suggestions

sumitvekariya:
> Address failing CI jobs:
> 
> * format storage_slot.rs
> * exclude `stylus-test` from `cargo hack` (the command targets `wasm32-unknown-unknown`, which by default excludes `stylus-test` feature)
> 
> I think we got the solution.

addresed both the suggestions

bidzyyys:
Replaced by #823 @sumitvekariya @0xNeshi. Cherry-picked all the commits!

PR #806, opened 2025-08-19
feat: add eddsa benchmark

qalisander:
Benchmark for eddsa signature verification.

### DIFF
--- a/benches/src/eddsa.rs
+++ b/benches/src/eddsa.rs
@@ -1,0 +1,92 @@
+use alloy::{
+    network::{AnyNetwork, EthereumWallet},
+    primitives::Address,
+    providers::ProviderBuilder,
+    sol,
+    sol_types::SolCall,
+};
+use alloy_primitives::hex;
+use e2e::{receipt, Account};
+use openzeppelin_crypto::{
+    curve::CurveGroup,
+    eddsa::{Signature, SigningKey, VerifyingKey},
+    field::prime::PrimeField,
+};
+
+use crate::{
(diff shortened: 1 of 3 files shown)


benches/src/eddsa.rs
```diff
@@ -0,0 +1,92 @@
...
+    );
+    let signing_key = SigningKey::from_bytes(&secret_key);
+
+    // Verify with signed message.
+    let message = b"Sign me!";
+    let signature = signing_key.sign(message);
+
+    #[rustfmt::skip]
+    let receipts = vec![
+        (EddsaExample::verifyCall::SIGNATURE, receipt!(contract.verify(
```
0xNeshi, line 56: Shouldn't we be testing and benchmarking signature creation too?

qalisander: Since we're testing it in e2e.. That is exact same test case

0xNeshi: But we're not benchmarking it, shouldn't we do that too?
@bidzyyys @qalisander

0xNeshi: If we should be benchmarking signature creation, but there are more pressing issues atm, could you create an issue to track this?

0xNeshi: @bidzyyys made a great point:
>IMHO we should not benchmark signature creation as it should not happen onchain. What we have in benchmarks are onchain operations, and signature verification is one of them. Curious to hear your thoughts!

No action is necessary then, resolving the conversation.

PR #815, opened 2025-08-20
fix(M-02): Fix edge case with U64 -> u128 conversion

qalisander:


bidzyyys (approved):
@qalisander missing changelog

Issue #735, opened 2025-07-14
[Feature]: Return `Result` instead of `panic` Wherever Possible


0xNeshi:
### What is the feature you would like to see?

We should use Rust's type system to its fullest potential. Panics bring little value to us, as they cannot be "handled" in the way that errors can, introducing gotchas to lib users.

This would entail introducing new errors into existing `enum Error`s, which do not exist in Solidity. This should be fine, as Stylus devs should expect Rust smart contracts to have wider and better error handling than Solidity.

@bidzyyys @qalisander 

Related https://github.com/OpenZeppelin/stylus-test-helpers/issues/38

### Contribution Guidelines

- [x] I agree to follow this project's Contribution Guidelines

0xNeshi:
Hi @temaniarpit27 , thanks for taking an interest in contributing!

Because this is a big issue, could you start with creating a PR with updates to just `VestingWallet` and any related code? 
That way we can more easily review and coordinate what needs to be done to other contracts.

Once you create the PR, please link it to this issue.

Thanks!

0xNeshi:
@temaniarpit27 just saw that you showed interest in another issue, it would be more helpful to us if you could implement that:
https://github.com/OpenZeppelin/rust-contracts-stylus/issues/745#issuecomment-3125846507

PR #816, opened 2025-08-20
fix(N-01): add description for Projective -> Affine edwards point conversion

qalisander:


lib/crypto/src/curve/te/affine.rs
```diff
@@ -238,8 +238,11 @@ impl<P: TECurveConfig> From<Projective<P>> for Affine<P> {
             // If Z is one, the point is already normalized.
             Affine::new_unchecked(p.x, p.y)
         } else {
-            // Z is nonzero, so it must have inverse in a field.
-            let z_inv = p.z.inverse().unwrap();
+            // Projective Z coordinate should not be zero after computations,
+            // unless it wasn't instantiated invalid.
+            let z_inv =
+                p.z.inverse()
+                    .expect("projective Z coordinate should not be zero");
```
0xNeshi, line 245: I'm more and more in favor of replacing all panics with Errors (see: https://github.com/OpenZeppelin/rust-contracts-stylus/issues/735).
It would allow devs to handle the errors gracefully instead of panicking.

NOTE: there may be specific cases where panics make sense, like here where it's not expected that a type is invalidly instantiated. I'm more referring to the general case of replacing panics with errors.

bidzyyys: Just to summarize my attitude around errors vs panics in Rust:
- **Errors**: Part of normal control flow. They occur when external conditions (like user input, files, or the environment) don’t meet the program’s requirements, but the program itself is still in a valid state.
- **Panics**: Indicate bugs or impossible states in the program logic. They’re used to catch developer mistakes or situations that should never occur if the code is correct.

qalisander: I see it should part of the bigger effort, creating error type system for crypto library

qalisander: Created [issue](https://github.com/OpenZeppelin/rust-contracts-stylus/issues/820)

PR #824, opened 2025-08-26
build(ci): optimize CI jobs

bidzyyys:


.github/workflows/gas-bench.yml
```diff
@@ -12,27 +12,54 @@ concurrency:
...
     runs-on: ubuntu-latest
+    timeout-minutes: 45
     steps:
-      - uses: actions/checkout@v4
+      - name: Checkout code
+        uses: actions/checkout@v4
 
-      - name: Install rust
+      - name: Install Rust toolchain
         uses: actions-rust-lang/setup-rust-toolchain@v1
```
qalisander, line 27: I do remember the main reason I've added this gh action: it has caching features.
Wasn't it working before?
If it doesn't work may be just drop it?

.github/workflows/check-abi.yml
```diff
@@ -11,20 +11,36 @@ concurrency:
...
     name: Check ABI
     runs-on: ubuntu-latest
+    timeout-minutes: 30
     steps:
-      - uses: actions/checkout@v4
+      - name: Checkout code
+        uses: actions/checkout@v4
 
-      - name: Install rust
+      - name: Install Rust toolchain
```
0xNeshi, line 25: The name changes are useful, but the additional caching steps don't seem to be reducing the run time :thinking: 
Maybe I'm missing something?

qalisander: Yeah. We need to research that later and address this pr towards main

PR #827, opened 2025-08-29
fix(N-02): add EnumerableSet::values_slice to prevent out-of-gas scenarios when querying large sets

bidzyyys:


pepebndc (approved):

pepebndc (approved):

contracts/src/utils/structs/enumerable_set/mod.rs
```diff
@@ -165,6 +167,68 @@ impl<T: Element> EnumerableSet<T> {
...
+            let end_idx: usize =
+                end.try_into().expect("`end` index must fit in `usize`");
+
+            (start_idx..end_idx)
+                .map(|idx| {
+                    self.at(U256::from(idx))
+                        .expect("element at index: {idx} must exist")
+                })
+                .collect()
+        }
```
qalisander:
```suggestion
        if let (Ok(start_idx), Ok(end_idx)) =
            (usize::try_from(start), usize::try_from(end))
        {
            // fast path: use native [`usize`] iteration with iterator
            // chains.
            (start_idx..end_idx)
                .map(|idx| {
                    self.at(U256::from(idx))
                        .expect("element at index: {idx} must exist")
                })
                .collect()
        } else {
            // slow path: pure [`U256`] arithmetic for extremely large
            // indices.
            let mut result = Vec::new();
            let mut current = start;

            while current < end {
                let value = self
                    .at(current)
                    .expect("element at index: {current} must exist");
                result.push(value);

                current += ONE;
            }

            result
        }
```

qalisander: Will have a bit less conversions. Wdt?

contracts/src/utils/structs/enumerable_set/mod.rs
```diff
@@ -18,6 +18,8 @@ use stylus_sdk::{
     storage::{StorageMap, StorageType, StorageU256, StorageVec},
 };
 
+const ONE: U256 = uint!(1_U256);
```
qalisander: I think we can use `U256::ONE`

qalisander (commented):
Good job!
Left some comments:

contracts/src/utils/structs/enumerable_set/mod.rs
bidzyyys: Great catch!

pepebndc (approved):

PR #832, opened 2025-09-01
fix(N-05): add warnings about gas consumption

bidzyyys:


pepebndc (approved):

qalisander:
Should we add a changelog note since we're changing part of public docs?

qalisander (approved):

